PR #21, opened 2026-05-15
fix(bip): bind to INADDR_ANY so broadcasts reach the socket

chappo:
## Summary

`BipTransport::start()` binds the UDP socket to the user-supplied `interface` IP.  On Linux this means the kernel will not deliver packets whose destination IP is the subnet broadcast (e.g. `10.10.0.255`) or the limited broadcast (`255.255.255.255`) to that socket — the destination IP does not match the bind IP.  `SO_BROADCAST` (already set on line 306) only permits **sending** broadcasts; it has no effect on reception.

The effect is symmetric and silent: a `bacnet-server` bound to a specific interface IP never sees Who-Is broadcasts, and a client bound to a specific interface IP never sees the broadcast I-Am replies.  Devices on the same LAN can only talk via unicast.

This was reproducible end-to-end with two unmodified v0.9.0 processes on the same /24 — `discover_devices` returns `No devices discovered` against a server that is reachable and answers unicast Who-Is correctly.

## Repro

```bash
# Server (10.10.0.220) — runs in a macvlan'd container
bacnet-device --transport=bip --interface=10.10.0.220 --port=47808 \
              --broadcast=10.10.0.255 --device-instance=1001 --objects=8

# Client (10.10.0.106) — via bacnet-mcp v0.9.6 with transports.bip.interface=10.10.0.106
# MCP call: discover_devices low=1000 high=2000
```

Expected: server is in the I-Am reply set.  Actual: `No devices discovered.`

## Wire-level proof

Two simultaneous tcpdumps — inside the server net-namespace and on the client's eth0 — during a single `discover_devices` call.

**Before this patch (both sides bound to specific interface IP):**
```
# Inside server (bound 10.10.0.220:47808):
06:29:22.593798  IP 10.10.0.106.47808 > 10.10.0.255.47808: UDP, length 18  (Who-Is)
                 # ← packet visible on the wire; bound socket never receives it.
```

**After patching only the server to bind 0.0.0.0:**
```
# Inside server (bound 0.0.0.0:47808):
06:29:22.593798  IP 10.10.0.106.47808 > 10.10.0.255.47808: UDP, length 18  (Who-Is)
06:29:22.594085  IP 10.10.0.220.47808 > 10.10.0.255.47808: UDP, length 20  (I-Am, 287µs later)

# On client eth0 (still bound 10.10.0.106:47808):
06:29:22.593927  IP 10.10.0.220.47808 > 10.10.0.255.47808: UDP, length 20  (I-Am)
                 # ← I-Am visible on the NIC; client socket never receives it.
                 # discover_devices still returns 0.
```

After patching both sides, broadcast discovery works through the full stack: 3 sims found, vendor/firmware/protocol fields all populated by the downstream enrichment step.

## Fix

`crates/bacnet-transport/src/bip/mod.rs:309` — bind to `Ipv4Addr::UNSPECIFIED` instead of `self.interface`.  The `self.interface` value continues to be used on line 318 to compute the announced `local_mac`, so I-Am responses still carry the correct source IP.

Single-line change plus a comment that explains *why* (the bug recurs easily without one).

## Test

`bip::tests::bind_address_is_inaddr_any` asserts that after `BipTransport::new(Ipv4Addr::LOCALHOST, 0, _).start()`, the socket's `local_addr()` is `0.0.0.0`, and the announced `local_mac` still reflects `127.0.0.1`.  No external network required, runs in <10ms.

187 tests pass (`cargo test -p bacnet-transport`), `cargo fmt --check` clean.

## Relation to prior work

Follows PR #8 (chappo, v0.8.1) which fixed the port-binding side of the same code path — that PR made the BIP **port** match the configured value so reply traffic on UDP/47808 reached the socket.  This PR fixes the **address** side: even with the right port, broadcast packets miss the socket when it's bound to a specific interface IP.  Same file, same lines, the two fixes complete the binding correctness story.

## Related issue

A separate issue will reference this PR for tracking, but the diagnosis and fix are self-contained here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,6 +135,11 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           key: ${{ matrix.os }}
+          # Do not cache ~/.cargo/bin; a stale cache from an older runner
+          # image can clobber rustup's shims after dtolnay installs them,
+          # causing `cargo test` to invoke a broken `rustup-init`. This
+          # bit us on macos-15-arm64 image 20260511.0048 — see PR #21.
+          cache-bin: false
       - run: cargo test --workspace --exclude rusty-bacnet --exclude bacnet-wasm --locked ${{ matrix.features }}
 
   deny:
--- a/crates/bacnet-transport/src/bip/mod.rs
+++ b/crates/bacnet-transport/src/bip/mod.rs
@@ -306,6 +306,18 @@ impl TransportPort for BipTransport {
         socket2.set_broadcast(true).map_err(Error::Transport)?;
         socket2.set_nonblocking(true).map_err(Error::Transport)?;
 
+        // Validate self.interface is a real local IP before binding the real
+        // socket to 0.0.0.0 below. Previously the kernel enforced this when
+        // we bound directly to self.interface (EADDRNOTAVAIL on typos); now
+        // we probe with a throwaway bind on an ephemeral port so a
+        // misconfigured interface fails fast at startup rather than silently
+        // advertising an unowned IP in I-Am replies via local_mac. See
+        // tests::start_fails_on_nonlocal_interface.
+        if !self.interface.is_unspecified() {
+            std::net::UdpSocket::bind(SocketAddrV4::new(self.interface, 0))
+                .map_err(Error::Transport)?;
+        }
+
         // Always bind to INADDR_ANY so subnet- and limited-broadcast packets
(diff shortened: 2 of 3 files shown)


jscott3201:
@codex

jscott3201:
Merge this soon. MacOS CI has been flaky lately with rust builds.

jscott3201:
@chappo If you get a chance to test that those patches I made didn't break that otherwise I can this weekend get a quick setup to verify. LGTM to me otherwise

PR #23, opened 2026-06-27
feat(samples): add discovery and point enumeration binaries

bbartling:


examples/rust/samples/mini-device-revisited/src/main.rs
```diff
@@ -0,0 +1,632 @@
...
+    Ok(frame.to_vec())
+}
+
+fn send_iam_broadcast(
+    net: &NetworkConfig,
+    port: u16,
+    instance: u32,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let frame = build_iam_broadcast_frame(instance)?;
+    let socket = StdUdpSocket::bind(SocketAddrV4::new(Ipv4Addr::UNSPECIFIED, 0))?;
```
jscott3201: This manual I-Am is sent from a fresh UDP socket bound to port `0`, so the packet leaves with an ephemeral source port instead of the BACnet/IP service port. In this stack, the BACnet/IP source MAC is derived from the UDP sender IP and port for `ORIGINAL_BROADCAST_NPDU`, and the client stores that source MAC directly in the discovery table. The table is keyed by device instance, so a later periodic I-Am from this path can overwrite a correct Who-Is response with an unreachable ephemeral port.

Impact: passive scanners, or any client that receives this periodic announcement after a normal response, can cache this device at `ip:ephemeral_port`. Follow-up reads/writes will then time out or target the wrong endpoint.

Please route unsolicited I-Am announcements through the same bound BACnet transport/server socket that handles Who-Is responses, or expose a server/transport API for broadcasting APDUs from the bound socket. Avoid constructing BVLL frames on a separate socket. Add a regression/integration test if possible: after receiving one of these announcements, the discovered MAC should contain the configured BACnet port, not an ephemeral source port.

examples/rust/samples/mini-device-revisited/src/main.rs
```diff
@@ -0,0 +1,632 @@
...
+    }
+
+    let net = resolve_network_config(&args);
+
+    info!(
+        "Starting mini-device-revisited: name={} instance={} device_ip={} bind={}:{} broadcast={}",
+        args.name, args.instance, net.device_ip, net.socket_bind, args.port, net.broadcast
+    );
+
+    kill_prior_listeners(args.port);
```
jscott3201: Starting the sample unconditionally calls `kill_prior_listeners`, which can terminate unrelated processes. That helper first kills any process whose `/proc/<pid>/exe` matches this binary and then runs `fuser -k {port}/udp`, which kills anything holding the UDP port. The wrapper script also runs `fuser -k ${BACNET_PORT:-47808}/udp`.

Impact: running an example can terminate another local BACnet server, BBMD, scanner, test harness, or a production process using the same UDP port. Examples should be safe to run by default and fail clearly when the port is busy.

Please remove the automatic process killing from the default startup path. Prefer a normal bind failure with a clear error message. If replacement behavior is useful for local demos, make it explicit, for example `--replace-existing`, and scope it to a PID file created by this sample rather than killing every holder of the BACnet UDP port.

examples/rust/samples/mini-device-revisited/src/main.rs
```diff
@@ -0,0 +1,632 @@
...
+    verify_udp_bind(net.socket_bind, args.port);
+
+    let db = build_database(&args)?;
+    info!(
+        "Object database: {} objects (device + 4 points in object-list)",
+        db.len()
+    );
+
+    let mut server = match BACnetServer::bip_builder()
```
jscott3201, line 506: The Device object and manual I-Am frame use `VENDOR_ID = 999`, but the server builder does not set `ServerConfig.vendor_id`, so built-in Who-Is responses use the default vendor ID `0`. That means scanners can see different vendor IDs for the same device depending on whether they receive the manual periodic I-Am or the server's Who-Is response.

Please set the server config vendor ID when building the server, e.g. add `.vendor_id(VENDOR_ID)` to this builder chain. It would also be worth testing that a normal Who-Is response and the Device object's `vendor-identifier` agree.

examples/rust/samples/mini-device-revisited/src/main.rs
```diff
@@ -0,0 +1,632 @@
...
+        }
+    }
+}
+
+fn build_database(args: &Args) -> Result<ObjectDatabase, Box<dyn std::error::Error>> {
+    let mut db = ObjectDatabase::new();
+    let device_oid = ObjectIdentifier::new(ObjectType::DEVICE, args.instance)?;
+
+    // --- Read-only analogValue:1 ---
+    let mut read_only_av = AnalogValueObject::new(1, "read-only-av", UNITS_DEGF)?;
```
jscott3201: These points are labelled and documented as read-only, but `AnalogValueObject` and `BinaryValueObject` are commandable object types in this crate: they expose priority arrays and accept `Present_Value` writes. Discovery tools will therefore report priority-array support for the supposed read-only AV:1/BV:1 points, and clients can command them.

Please either use object types/implementations that actually reject writes for the read-only demo points, or update the sample names, comments, README, and discovery expectations so they do not claim these points are read-only. If the goal is to demonstrate both read-only and commandable points, this should be enforced in behavior, not only in labels.

examples/rust/samples/bacnet-write/src/main.rs
```diff
@@ -0,0 +1,546 @@
...
+    }
+    println!("OK: write taken at P{} (present-value + priority-array match)", args.priority);
+
+    if args.no_revert {
+        let _ = client.stop().await;
+        println!("\nDone (--no-revert: priority slot left active).");
+        return;
+    }
+
+    if let Err(e) = write_present_value(
```
jscott3201, line 573: The default auto-revert path can delete an existing command at the selected priority. The code reads `baseline.priority_slot`, writes the new value into that same slot, then always reverts by writing `Null` at the selected priority. If the target priority already contained a real command before the sample ran, this clears it instead of restoring it.

Impact: this sample can permanently relinquish an operator/controller command on a real device while printing that it reverted. That is especially risky because `--priority` is configurable and the default behavior is to revert automatically.

Please preserve and restore the original slot value: if `baseline.priority_slot` is `Some(v)`, write `v` back at the same priority; if it was `None`, write `Null`. Alternatively, refuse to run unless the selected priority slot is initially null, with an explicit `--force-relinquish`/`--no-revert` style override. The final verification should compare both present-value and the selected priority-array slot back to the baseline, not only check that the slot is null.

examples/rust/samples/bacnet-write/src/main.rs
```diff
@@ -0,0 +1,546 @@
...
+            eprintln!("ERROR: baseline read failed: {e}");
+            let _ = client.stop().await;
+            process::exit(1);
+        }
+    };
+
+    println!("\n=== Baseline ===");
+    print_snapshot("Before", &point_oid, &baseline, args.priority);
+
+    let write_val = PropertyValue::Real(args.value);
```
jscott3201, line 521: `parse_point` accepts binary and multi-state object types, but the write value is always encoded as `PropertyValue::Real`. That is only appropriate for analog present-values. Binary present-values should be encoded as `Enumerated`, and multi-state present-values as `Unsigned`.

Impact: users can pass a supported-looking point such as `binary-output:1` or `multi-state-output:1`, but the sample sends the wrong BACnet application type. Depending on the device, this will fail with a property/type error or produce misleading troubleshooting output.

Please either restrict the CLI to analog object types until the other types are implemented, or branch on `point_oid.object_type()` and parse/encode values with the correct BACnet application type for analog, binary, and multi-state objects. The README should document the accepted value format for each supported object family.

examples/rust/samples/point-discover/src/main.rs
```diff
@@ -0,0 +1,729 @@
...
+            }
+            if !oids.is_empty() {
+                return Ok(oids);
+            }
+        }
+    }
+
+    // Fallback: single ReadProperty for the full list.
+    let ack = read_prop(client, device_instance, device_oid, None).await?;
+    if let Some(PropertyValue::List(items)) = decode_prop(&ack.property_value) {
```
jscott3201: The full `object-list` fallback appears unable to decode a normal multi-element BACnet property value. `decode_prop` calls `decode_application_value` once and discards the returned offset, while `PropertyValue::List` is encoded by writing consecutive application-tagged values, not by wrapping them in a distinct list tag. As a result, a device that supports a whole-object-list read but not indexed object-list reads can still be reported as empty/unsupported.

Please decode repeated application values until the buffer offset reaches the end, collecting each `PropertyValue::ObjectIdentifier`. A small unit test using an encoded full object-list with multiple object identifiers would cover this path and prevent regressions.

examples/rust/samples/point-discover/src/main.rs
```diff
@@ -0,0 +1,729 @@
...
+    let device_oid = match ObjectIdentifier::new(ObjectType::DEVICE, args.device) {
+        Ok(o) => o,
+        Err(e) => {
+            eprintln!("ERROR: bad device instance {}: {e}", args.device);
+            process::exit(1);
+        }
+    };
+
+    if let Some(device_ip) = args.address {
+        let mac = encode_bip_mac(device_ip.octets(), DEFAULT_BACNET_PORT).to_vec();
```
jscott3201: The fixed-address path hardcodes the target device port to `DEFAULT_BACNET_PORT`. The same pattern exists in `rpm-read` and `bacnet-write`. `--port` controls the local client bind port, so there is currently no way to use `--address` with a sample server or real device listening on a non-default BACnet/IP port.

Impact: `mini-device-revisited --port <non-default>` can be run, but the direct-address clients cannot target it directly; they silently register `address:47808` and then time out.

Please add an explicit target-port option or accept `--address ip[:port]`, and use that target port in `encode_bip_mac`. Apply the same fix to `point-discover`, `rpm-read`, and `bacnet-write`, and update the run scripts/READMEs to clarify the difference between local bind port and remote device port.

Cargo.toml
```diff
@@ -15,6 +15,13 @@ members = [
     "crates/bacnet-cli",
     "benchmarks",
 ]
+exclude = [
```
jscott3201, line 18: Because the new sample crates are excluded from the workspace, the existing workspace CI jobs will not compile, lint, test, or audit them. That leaves the main contents of this PR dependent on someone manually running each manifest.

Please add explicit CI coverage for these excluded sample crates, for example a matrix over the five `examples/rust/samples/*/Cargo.toml` files that runs `cargo check --locked --manifest-path ...` and `cargo clippy --locked --manifest-path ... --all-targets`. Since each sample has its own committed lockfile, consider also auditing those lockfiles or making the crates workspace members if that is compatible with the repo structure.

jscott3201 (changes requested):
Thanks for adding these network samples. I did a deep pass over the PR with a few focused sub-reviews. The sample crates compile locally, but there are several correctness and safety issues that should be addressed before merging, especially around BACnet/IP discovery identity and writes to real devices.

Validation run locally on PR head `be66942412868df1cbedeb49c397caac04a8f274`:
- `cargo check --manifest-path` passed for all five new sample crates.
- `cargo clippy --no-deps --manifest-path ... -- -D warnings` passed for all five new sample crates.
- `cargo check --workspace` passed.
- `git diff --check origin/dev...HEAD` reports `examples/rust/samples/README.md:59: new blank line at EOF`.

Please address the inline comments below.

bbartling:
Pushed review fixes in commit `be2c24c` on `examples/network-samples`. Summary by review item:

**I-Am / discovery**
- Sample devices call `BacnetServer::broadcast_i_am()` instead of hand-rolling NPDU/unconfirmed I-Am on the transport.
- Startup and optional periodic I-Am use the server API; Who-Is handling stays on the server path.

**vendor_id**
- `BacnetServerBuilder::vendor_id` and server lifecycle propagate vendor ID into I-Am and device identity consistently (mini-device-revisited uses the builder).

**AI/BI read-only**
- mini-device-revisited: analog/binary *input* objects are read-only; commandable writes remain on AV/BV only; simulation updates read-only points only.

**bacnet-write**
- Restored intended write/demo behavior: `--force` for out-of-service writes, analog-only targeting where appropriate, and verification aligned with priority-array / relinquished demo semantics from prior fix.

**object-list decode**
- point-discover decodes ReadProperty object-list payloads instead of treating raw bytes as opaque.

**ip:port address**
- Samples bind/advertise using explicit `ip:port` (socket address) rather than IP-only assumptions.

**CI matrix**
- `.github/workflows/ci.yml` runs `cargo check --locked` across the Rust network sample crates in CI.

**replace-existing**
- Added/clarified replace-existing handling where the review asked for explicit behavior (object creation / idempotency in samples).

**mini-device-revisited compile**
- Fixed `local_mac()` lifetime (E0597): copy 6-byte MAC out of the mutex guard before use.

All five sample crates under `examples/rust/samples/` (`bacnet-write`, `mini-device-revisited`, `point-discover`, `rpm-read`, `whois-scan`) pass `cargo check --locked` locally.

Cargo.toml
jscott3201: Ignored, no new CI on PRs. Dev is fast lane.

crates/bacnet-server/src/server/lifecycle.rs
```diff
@@ -469,6 +469,38 @@ impl<T: TransportPort + 'static> BACnetServer<T> {
...
+
+        let pdu = Apdu::UnconfirmedRequest(UnconfirmedRequestPdu {
+            service_choice: UnconfirmedServiceChoice::I_AM,
+            service_request: service_buf.freeze(),
+        });
+
+        let mut buf = BytesMut::new();
+        encode_apdu(&mut buf, &pdu)?;
+
+        self.network
```
jscott3201: `broadcast_i_am` still holds the object database read guard while awaiting the network send. The `db` guard acquired above is not scoped or explicitly dropped before this `broadcast_apdu(...).await`, so any writer that needs the database, such as simulation updates or future server-side mutations, can be blocked for the duration of the socket send. It also makes this helper harder to reuse safely because a future broadcast path that indirectly needs database access could deadlock.

Please limit the read lock to just the device-object lookup before doing any encoding or network I/O, for example:

```rust
let device_oid = {
    let db = self.db.read().await;
    db.list_objects()
        .into_iter()
        .find(|oid| oid.object_type() == ObjectType::DEVICE)
        .ok_or_else(|| Error::Encoding("no Device object in database".into()))?
};
```

Then build the `IAmRequest` and call `broadcast_apdu` after the guard has been dropped. In the sample, it would also be worth avoiding `server.lock().await.broadcast_i_am().await` so the outer `Mutex<BACnetServer<_>>` is not held across the same network await.

examples/rust/samples/whois-scan/src/main.rs
```diff
@@ -0,0 +1,199 @@
...
+        .await
+}
+
+#[tokio::main]
+async fn main() {
+    let args = Args::parse();
+    let interface = resolve_interface(&args);
+    let broadcast = args
+        .broadcast
+        .unwrap_or_else(|| subnet_broadcast(interface));
```
jscott3201: This default broadcast calculation breaks when auto-detection falls back to `0.0.0.0`. `resolve_interface` returns `Ipv4Addr::UNSPECIFIED` if `--interface` is omitted and `enp3s0` is not present, then `subnet_broadcast(0.0.0.0)` produces `0.0.0.255`. That is not a useful BACnet/IP broadcast target, so on hosts without an `enp3s0` interface the sample will silently send Who-Is to the wrong address.

This same pattern appears in the other client samples (`point-discover`, `bacnet-write`, and `rpm-read`), while `mini-device-revisited` already handles the unspecified case by using `Ipv4Addr::BROADCAST`. Please apply the same fallback consistently, e.g. use `255.255.255.255` when the resolved interface is unspecified, or fail fast and require `--interface`/`--broadcast`. A small unit test around the default-broadcast helper would keep this from regressing.

jscott3201 (changes requested):
Re-reviewed after the latest fixes and after removing the added CI job. Two actionable issues remain before I would merge this: the new I-Am helper holds locks across await, and the client samples derive an unusable default broadcast address when interface auto-detection fails.

bbartling:
Rebased onto upstream/dev and fixed the remaining CI issues locally:

- mini-device-revisited: wire `MAX_APDU_LENGTH` into `DeviceConfig` (clippy dead_code)
- Rebased to pick up pyo3 0.29 audit fix already merged on dev (#24)

All local CI checks pass (fmt, workspace clippy/check, cargo audit, sample clippy).

Latest commit: fc77795 — GitHub Actions shows `action_required` for fork PR workflow approval. Could you please **Approve and run** the pending workflow? Thanks!

crates/bacnet-server/src/server/lifecycle.rs
jscott3201: Fixed in d77205f.

Changes made:
- `BACnetServer::broadcast_i_am` now delegates to a shared helper that scopes the database read guard only around the Device object lookup. I-Am encoding and `broadcast_apdu(...).await` now happen after that guard has dropped.
- Added `IAmBroadcaster`, a cloneable handle over the bound server transport, so samples can send unsolicited I-Am without wrapping the whole `BACnetServer` in a mutex.
- Updated `mini-device-revisited` to use that broadcaster for startup and periodic I-Am sends, so it no longer holds an outer server mutex across the network await.

Validated with `cargo check --workspace --exclude rusty-bacnet --exclude bacnet-wasm --locked`, `cargo check --locked` for `mini-device-revisited`, `cargo fmt --all --check`, `git diff --check`, and the file-size check.

examples/rust/samples/whois-scan/src/main.rs
jscott3201: Fixed in d77205f.

Changes made:
- Added a `default_broadcast` fallback that returns `255.255.255.255` when the resolved interface is `0.0.0.0`, and otherwise keeps the existing /24 directed broadcast behavior.
- Applied that fallback to `whois-scan`, `point-discover`, `bacnet-write`, and `rpm-read`.
- Added unit coverage for unspecified and bound-interface defaults. In `point-discover`, the helper and test live in `src/net_defaults.rs` so `main.rs` stays within the 700 LOC cap.

Validated with `cargo test --locked` in all four affected client sample crates, plus `cargo fmt --all --check`, `git diff --check`, and the tracked Rust file-size check.

PR #138, opened 2026-07-21
fix(encoding): encode BACnetPriorityValue choice tags

hechuan9:
## Summary

- Encode server-side BACnetPriorityValue values with the standard Null `[0]`, Real `[1]`, Enumerated `[2]`, and Unsigned `[3]` context choice tags.
- Preserve the choice tag explicitly in `PropertyValue::ContextTagged` instead of inferring it from payload bytes.
- Reject unsupported, oversized, or nested context values with typed errors.
- Keep context-value encoding atomic so an error does not leave partial bytes in the destination buffer.

## Root cause and current scope

Priority-array object values were represented as ordinary application values. Server responses therefore emitted application tags where BACnetPriorityValue requires a context-tagged CHOICE, so the value type could not be represented correctly on the wire.

This Draft establishes the encoding-side representation and propagates it through priority-array object implementations. It also includes CLI and Python surface plumbing for the new representation.

The generic client decoder still rejects context-tagged values. Property-aware decoding from a Priority_Array read response into Rust, Python, and CLI values is not yet implemented, so this Draft does not claim end-to-end client round trips.

## API and compatibility

`PropertyValue` gains a public `ContextTagged` variant and Python values expose the context tag. Because the Rust enum is public and not `#[non_exhaustive]`, this can break downstream exhaustive matches even though existing constructors and application-tag encodings are unchanged. Maintainer guidance on the preferred compatibility treatment is welcome before this PR is marked Ready.

Unsupported context tags, tag numbers above 254, and nested context values are rejected rather than silently emitted with ambiguous wire semantics.

## Validation

Local validation on `e8ab7fbd66606667be617e8e2c4d1fa85fafd486`:

- `cargo fmt --all --check`
- `git diff --check origin/dev...HEAD`
- `cargo test -p bacnet-types -p bacnet-encoding -p bacnet-objects --locked`
- `cargo clippy --workspace --exclude rusty-bacnet --exclude bacnet-wasm --all-targets --locked`

The existing missing-docs warning baseline remains unchanged. Upstream PR CI is waiting for maintainer approval to run because this is a first-time external fork contribution.

## Draft gates before Ready

- Decide whether this PR remains an encoding-side fix or adds property-aware Priority_Array client decoding and end-to-end tests.
- Resolve the public enum compatibility approach with maintainer input.
- Add the accepted behavior and API impact to `CHANGELOG.md`.
- Run the upstream Tier-1 CI after workflow approval; add a Python wheel/import smoke if the Python surface remains in scope.

## Review focus

- Context-tag encoding and rejection boundaries in `bacnet-encoding`.
- Priority-array encoding across analog, binary, multistate, lighting, and value objects.
- Preferred public API shape for representing context-tagged BACnet CHOICE values.

### DIFF
--- a/crates/bacnet-cli/src/output.rs
+++ b/crates/bacnet-cli/src/output.rs
@@ -206,6 +206,9 @@ pub fn format_property_value(value: &PropertyValue) -> String {
             let formatted: Vec<String> = items.iter().map(format_property_value).collect();
             format!("[{}]", formatted.join(", "))
         }
+        PropertyValue::ContextTagged { tag_number, value } => {
+            format!("context[{tag_number}]({})", format_property_value(value))
+        }
     }
 }
 
--- a/crates/bacnet-encoding/src/primitives/mod.rs
+++ b/crates/bacnet-encoding/src/primitives/mod.rs
@@ -599,7 +599,9 @@ pub fn encode_property_value(buf: &mut BytesMut, value: &PropertyValue) -> Resul
             }
         }
         PropertyValue::ContextTagged { tag_number, value } => {
-            encode_context_tagged_property_value(buf, *tag_number, value)?;
+            let mut encoded = BytesMut::new();
+            encode_context_tagged_property_value(&mut encoded, *tag_number, value)?;
+            buf.extend_from_slice(&encoded);
         }
     }
     Ok(())
--- a/crates/bacnet-encoding/src/primitives/mod.rs
+++ b/crates/bacnet-encoding/src/primitives/mod.rs
@@ -514,7 +514,68 @@ pub fn decode_application_value(
     Ok((value, content_end))
 }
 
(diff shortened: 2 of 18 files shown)


hechuan9:
Closing after clause-level verification against ANSI/ASHRAE Standard 135-2020 Clause 21. BACnetPriorityValue primitive alternatives such as NULL, REAL, ENUMERATED, and Unsigned are application-tagged; explicit context tags apply only to the separately tagged constructed alternatives. This draft incorrectly mapped the primitive alternatives to context tags [0] through [3]. We will revert the fork-side change and re-investigate the original client behavior from the standard-defined wire contract. Reference: ANSI/ASHRAE Addendum ca to Standard 135-2020, Clause 21 aggregate production, PDF page 26, together with its published errata.

Issue #198, opened 2026-07-27
bacnet-cli/src/main.rs is at exactly 700/700, so any CLI change fails the cap gate


jscott3201:
### Affected revision

Observed on `dev` at commit `33a2dd7`.

### Behavior

`crates/bacnet-cli/src/main.rs` measures exactly 700 non-empty, non-comment lines — the ceiling enforced by `.github/scripts/check-file-size.sh`. Any added line fails CI, so the file cannot be edited at all without a split in the same change.

No batch of the current backlog plan touches `bacnet-cli`, which is why this was not folded into the Batch 1 headroom work (PR #197 and its `bacnet-server`/`bacnet-types` sibling). It is recorded here so the next person to touch the CLI is not surprised mid-change.

The file is dominated by the clap surface: `struct Cli` and `enum Command` occupy lines 24-359, roughly 340 of the 700. The remainder is `setup_tracing`, `resolve_format`, `resolve_target_mac`, `execute_command`, `parse_discover_range`, `execute_bip_command`, `run`, and `main`.

### Suggested direction

- move `Cli` and `Command` into a sibling `args` module, leaving dispatch in `main.rs`;
- note the fields are currently private and are read across the crate, so they need `pub` — harmless here because `bacnet-cli` is a binary, but it is a real diff rather than a pure move;
- `args.rs` additionally needs `std::net::Ipv4Addr`, `std::path::PathBuf`, and the `transport` module paths that `main.rs` currently provides implicitly;
- alternatively split `execute_command` (lines 419-662) by command family, which keeps visibility untouched.

A first attempt at the `args` extraction produced 10 compile errors of exactly these kinds, so budget for the import and visibility work rather than treating it as a mechanical move.

### Relevant code

- `crates/bacnet-cli/src/main.rs`
- `.github/scripts/check-file-size.sh`

PR #199, opened 2026-07-27
refactor(server,types): carve file-size headroom for the handler batches

jscott3201:
Batch 1, PR 5 of 5 — completes the headroom work. **No behavior change**, pure moves.

## Results

| File | Before | After |
|---|--:|---|
| `bacnet-server/src/event_enrollment/tests.rs` | 696 | 7 files by algorithm, max **214** |
| `bacnet-server/src/server/requests.rs` | 695 | mod **547** + unconfirmed **152** |
| `bacnet-server/src/server/event_notifications.rs` | 617 | **281** + tests **333** |
| `bacnet-server/src/handlers/alarm_event.rs` | 240 | 6 files, one per service, max **106** |
| `bacnet-types/src/constructed/event_parameter/mod.rs` | 630 | mod **260** + codec **216** + fault **165** |

## Why these, in this order

`event_enrollment/tests.rs` had **4 lines** of headroom, and **Batch 2 is entirely Event Enrollment work**. It went first.

`alarm_event.rs` was nowhere near the cap at 240 — it is split anyway because **Batch 4 lands seven handler PRs in it**. One file per service means those seven stop serializing on a single module, and stop racing the cap collectively.

## Visibility preserved, not widened

This is the thing worth checking in review:

- `handle_unconfirmed_request` was `pub(super)` in `server/requests.rs` — visible in `server`. From one level deeper it becomes `pub(in crate::server)`: the **same scope**, stated explicitly, so `server::dispatch` still reaches it. Not a widening.
- The event-parameter codec helpers were private to `event_parameter/mod.rs` and are now `pub(super)` in `codec.rs` — again the same module scope.

## Two mistakes, both compiler-caught

- Slicing `alarm_event.rs` at each `pub fn` line orphaned every `///` block at the tail of the previous file (`expected item after doc comment` ×5). Boundaries now walk back over the doc block.
- The trailing `#[cfg(test)] mod tests;` was swept into the `event_parameter` fault segment, where it resolved to a nonexistent `fault/tests.rs`.

Recording these because they are the failure modes of mechanical splitting, and the next person doing one will hit them.

## Verification

- `cargo test --workspace --exclude rusty-bacnet` — exit 0, **2234 tests**, 34 suites
- `cargo clippy --workspace --exclude rusty-bacnet --all-targets --locked` — exit 0, **zero** unused-import warnings (the duplicated header blocks introduced some; cleaned, including one left over from #197)
- `cargo fmt --all --check` — clean; cap gate — green
- **Leaf test-function names identical to `dev`** — no test added, dropped, or renamed; only module paths change

## Out of scope

`crates/bacnet-cli/src/main.rs` is also at exactly **700/700**. No batch touches the CLI, and extracting its clap surface needs visibility and import changes rather than a pure move — a first attempt produced 10 compile errors. Filed as **#198** rather than bundled here.

## No CHANGELOG entry

Nothing user-visible changed.


jscott3201:
Correcting a figure in the description: `event_enrollment/tests/` maxes at **212** (`compat.rs`), not 214. My number was measured before `cargo fix` removed the unused imports the split introduced. Caught in review.

Review also raised a NIT worth recording rather than fixing: moving the codec helpers and `fault_parameter_tag` one module deeper **narrows** their theoretical internal visibility — a sibling of `event_parameter` that could reach them on `dev` would now get a privacy error. There is no such caller, and no public or runtime behavior changes, so this is strictly tighter than before rather than looser. Noting it because the direction is the opposite of the usual refactor hazard.

Everything else verified independently: mechanical reconstruction found no changed statement, expression, literal, or attribute; `handle_unconfirmed_request` retains exactly `crate::server` visibility with `server/dispatch.rs` still its sole caller; `alarm_event` exposes the same six handlers; leaf test-name multisets identical at 2,273 entries including duplicate multiplicities.

Issue #200, opened 2026-07-27
CommandFailureDetector suppresses TO_FAULT distribution regardless of Event_Enable


jscott3201:
## Summary

`CommandFailureDetector::fire` hardcodes `distribute = false` for any state that is
neither NORMAL nor OFFNORMAL, where the two sibling detectors consult the
`TO_FAULT` bit of `Event_Enable`. If a TO_FAULT transition ever reaches this
detector, its notification is unconditionally suppressed — even with
`Event_Enable` bit 1 set.

## Evidence

`crates/bacnet-objects/src/event.rs`, three structurally identical `fire` methods:

```rust
// OutOfRangeDetector::fire — event.rs:317
_ => self.event_enable & Self::TO_FAULT != 0,

// ChangeOfStateDetector::fire — event.rs:475
_ => self.event_enable & Self::TO_FAULT != 0,

// CommandFailureDetector::fire — event.rs:592
_ => false,                                   // <-- diverges
```

## Reachability: latent today, live after #167

All four `fire` call sites pass `desired`, which comes from `compute_new_state`.
`CommandFailureDetector::compute_new_state` (event.rs:597) returns only
OFFNORMAL or NORMAL:

```rust
fn compute_new_state(&self, present_value: u32, feedback_value: u32) -> EventState {
    if present_value != feedback_value { EventState::OFFNORMAL } else { EventState::NORMAL }
}
```

So the `_` arm is unreachable as of today's code — this is a latent defect, not an
active one, and no test can currently exercise it.

That changes with **#167**, which wires Reliability into `Event_State` and makes
TO_FAULT reachable. At that point COMMAND_FAILURE objects would silently drop
every fault notification while OUT_OF_RANGE and CHANGE_OF_STATE objects deliver
theirs — a per-algorithm inconsistency with no basis in the standard.

## Spec basis

ASHRAE 135-2020 Clause 12.12 scopes `Event_Enable` to which transitions are
*distributed*; it does not carve out an exception by event algorithm. Clause
13.2.2.1.4 lists notification distribution as the one transition action gated on
`Event_Enable`, again with no algorithm-specific exemption. A COMMAND_FAILURE
object with the TO_FAULT bit set must distribute its fault notifications.

## Proposed fix

Change event.rs:592 to `_ => self.event_enable & Self::TO_FAULT != 0`, matching
the siblings, and add a test that drives the detector to FAULT and asserts
`distribute == true` with the bit set / `false` with it clear.

[...]

Issue #166, opened 2026-07-27
Event Enrollment evaluator discards same-state transitions that Clause 13.2.2.1.4 requires to be acted on


jscott3201:
### Affected revision

Observed on `dev` at commit `3990f3e`.

### Behavior

`crates/bacnet-server/src/event_enrollment/mod.rs:624-626` drops any evaluation whose result equals the current state:

```rust
if new_state == current_state {
    continue;
}
```

The guard runs before the `Event_Enable` check at `:628`, before the `Event_State` write at `:658`, and before the transition is pushed onto the returned vector at `:659-667`. `updates` (`:640-646`) is the only channel into that vector, so a same-state indication is unrecoverable — no state write, no timestamp, no notification.

Clause 13.2.2.1.4, Transition Actions (p642): "The actions are the same for all transitions and they shall be executed even if the transition does not change the event state (e.g., a transition from the OFFNORMAL event state to the OFFNORMAL event state)."

The clearest consequence is CHANGE_OF_VALUE. Its only condition, Clause 13.3.3 (p652), indicates "a transition to the NORMAL event state" while `pCurrentState is NORMAL`, and Figure 13-10 (p653) shows that single self-transition. The condition is not optional, so a correct CHANGE_OF_VALUE implementation is structurally unreportable behind this guard. CHANGE_OF_BITSTRING condition (c) (13.3.1, p650) and CHANGE_OF_STATE condition (c) (13.3.2, p651) also re-enter OFFNORMAL, but both are prefixed "Optional:" in the standard — leaving those two unimplemented is conformant.

Removing the guard on its own would be wrong, because no evaluator here can currently tell a genuine same-state indication from "nothing changed":

- `eval_change_of_state_struct` (`:282-319`) and `eval_change_of_bitstring_struct` (`:323-345`) take `_current` unused and return OFFNORMAL whenever the monitored value matches any alarm value. Neither retains the value that caused the last transition to OFFNORMAL, which is what condition (c) keys on, so an unguarded evaluator would emit a re-entry on every poll while the value sits unchanged in an alarm value.
- `eval_change_of_value_struct` (`:357-378`, delegating to `eval_change_of_value` at `:230-243`) returns OFFNORMAL when `|value| >= increment`, a state CHANGE_OF_VALUE can never indicate, and holds no change baseline (#137).

### Suggested direction

[...]

Issue #137, opened 2026-07-14
Track a baseline for Event Enrollment change-of-value evaluation


jscott3201:
### Affected revision

Observed on rusty-bacnet v0.10.1 at commit `8b8c0cba3e92caa30e145c91abedfa003c64f2c9`.

### Behavior

The Event Enrollment change-of-value branch compares the absolute current numeric value with the configured increment. It does not retain a prior reported value or compute a delta. A stable value whose magnitude exceeds the increment can remain classified as changed, while a change around zero can be misclassified.

### Suggested direction

- retain an explicit change baseline per enrollment;
- compare the current value with that baseline;
- define when the baseline advances and resets;
- add stable, below-threshold, threshold-boundary, repeated-change, and recovery tests.

### Relevant code

- `crates/bacnet-server/src/event_enrollment/mod.rs`

jscott3201:
Implemented in PR #290 (merged as daab3a8): CHANGE_OF_VALUE evaluates against a retained per-enrollment baseline (first sample establishes, local matter; baseline advances on fired NORMAL transitions), conforming the evaluator to 13.3.3.

Issue #127, opened 2026-07-14
Propagate Event Enrollment transitions into the notification lifecycle


jscott3201:
### Affected revision

Observed on rusty-bacnet v0.10.1 at commit `8b8c0cba3e92caa30e145c91abedfa003c64f2c9`.

### Behavior

The periodic Event Enrollment task evaluates enrollments and stores enabled state changes, but consumes the returned transitions only as debug output. They do not reach Notification Class routing, acknowledgement tracking, timestamp or message storage, or Event Log handling.

### Suggested direction

- forward enrollment transitions into the shared event-notification pipeline;
- maintain transition acknowledgement and timestamp state;
- declare Event Log integration behavior;
- add lifecycle tests for every supported enrollment algorithm.

### Relevant code

- `crates/bacnet-server/src/event_enrollment/mod.rs`
- `crates/bacnet-server/src/server/lifecycle.rs`
- `crates/bacnet-server/src/server/event_notifications.rs`

PR #201, opened 2026-07-27
fix(objects,server): scope Event_Enable to distribution, not detection (#136)

### LINKED ISSUE #136: Use one Event_Enable state-mutation policy across event evaluators
### Affected revision

### Behavior

Per-object detectors update their internal event state before checking the applicable `Event_Enable` bit, so a disabled transition changes state but is not emitted. The Event Enrollment evaluator checks the enable bit first and writes `Event_State` only for enabled transitions.

The two reachable evaluator families therefore give `Event_Enable` different state-mutation semantics.

### Suggested direction

- choose and document one policy for state mutation versus transition emission;
- apply it consistently to per-object and enrollment evaluators;
- keep observable `Event_State`, acknowledgement state, and later recovery transitions coherent;
- add disabled offnormal, fault, and normal transition tests.

### Relevant code

jscott3201:
Batch 2, PR 6. Closes #136.

## The spec position

ASHRAE 135-2020 Clause 13.2.2.1.4, "Transition Actions" (p640), lists **four**
actions a detected transition performs:

1. store the new `Event_State` (and specifically — "it is not acceptable to set
   Event_State to OFFNORMAL when the returned value is HIGH_LIMIT")
2. store the time of the transition in `Event_Time_Stamps`
3. store the message text in `Event_Message_Texts`, if present
4. indicate the transition to **both** the alarm-acknowledgment process
   (Clause 13.2.3) **and** the event-notification-distribution process
   (Clause 13.2.5)

and requires that they *"shall be executed even if the transition does not change
the event state."*

**None of the four is scoped to `Event_Enable`.** The gate sits downstream.
Clause 12.12 defines the property as three flags that *"separately enable and
disable the distribution of TO_OFFNORMAL, TO_FAULT, and TO_NORMAL notifications
(see Clause 13.2.5),"* and Clause 13.2.5 places it inside the distribution
process: *"The external distribution of notifications can be disabled by the
Event_Enable property."* Note "external" — whether local objects such as Event
Logs are affected is explicitly a local matter.

`Acked_Transitions` is **not** a transition action. Clause 13.2.3 has the
alarm-acknowledgment process maintain it from the unconditionally-indicated
transition, gated by `Ack_Required` — never by `Event_Enable`.

Both reporting families conflated detection with distribution — in different
ways, and to very different effect. That distinction is the substance of this PR
and worth checking carefully.

## Event Enrollment — the damaging case

The evaluator wrapped its entire state-persisting update in `if transition_enabled`,
so a suppressed transition never reached `set_event_state_internal`:

```
Event_Enable = TO_NORMAL only (0x04), high_limit 80, deadband 2
  pv 85  ->  HIGH_LIMIT detected, suppressed, Event_State stays NORMAL
  pv 70  ->  evaluator compares NORMAL vs NORMAL, no change, nothing fires
```

The **enabled** TO_NORMAL notification is silently lost. Clearing one
`Event_Enable` bit disabled a different one — and `Event_State`, a readable
property rather than a notification, reported a condition the device was no
longer in.

Pinned by `out_of_range_suppressed_offnormal_still_yields_enabled_return_to_normal`.

## Intrinsic detectors — structural, not observable

Worth being precise about, because the two families are not symmetric. On `dev`,
`fire` already assigned `self.event_state` **before** consulting `Event_Enable`
(the old test even asserted "state still updated"), so detector state was
correct. The defect was the return value:

```rust
enabled.then_some(change)   // "no transition" and "suppressed transition" -> same None
```

A caller could not learn that a suppressed transition had occurred. That matters
because action 4 indicates the transition to *two* processes: withholding it
denies the alarm-acknowledgment process as well as distribution — the very
process #123 will build `Event_Time_Stamps` / `Acked_Transitions` on.

## The change

Detection and distribution become separate results:

```rust
pub struct TransitionOutcome {      // new, bacnet-objects::event
    pub change: EventStateChange,
    pub distribute: bool,
}
```

Both paths always report the transition and always persist `Event_State`; the
`Event_Enable` verdict rides along and is applied at the two points of send
(`fire_event_notifications` and the 1-second `intrinsic_reporting_task`).
#123's bookkeeping hooks that same boundary, ahead of the `distribute` check,
and the comments there mark the spot.

**No notification previously sent is now suppressed, and the notification bytes
are unchanged.** The observable change is confined to Event Enrollment.

## Breaking API changes

- `BACnetObject::evaluate_intrinsic_reporting` and `tick_intrinsic_reporting`
  return `Option<TransitionOutcome>` instead of `Option<EventStateChange>`.
  `None` now means only "no transition fired".
- `EventEnrollmentTransition` gains `pub distribute: bool`.
- Neither struct is `#[non_exhaustive]`.

## Deliberately not fixed here

Recorded rather than hidden:

- **#166** — the enrollment same-state skip is still non-conformant. Removing it
  needs the #137 change baseline to distinguish a genuine same-state indication
  from "nothing changed"; without that it would re-fire every poll.
- **#127** — the enrollment path emits no notifications at all yet, so
  `distribute` is logged rather than acted on (which is also why it is logged —
  otherwise an operator sees state move with no notification and no explanation).
- **#202** (filed by this PR) — all four `EventState` → transition-bit mappings
  end in a catch-all defaulting to TO_FAULT, but Clause 13.2 states that "all
  states that are not normal and not fault are offnormal states," making
  TO_OFFNORMAL the correct residual. Latent today (`LIFE_SAFETY_ALARM` is never
  produced), inverted structurally. The fix is one shared helper, testable
  exhaustively over all six `EventState` values.
- **#200** (filed by this PR) — `CommandFailureDetector::fire` hardcodes
  `distribute = false` for FAULT where its two siblings read `TO_FAULT`;
  subsumed by #202.

## Verification

- `cargo test --workspace --exclude rusty-bacnet` — 2236 passed, 0 failed
- `cargo clippy --workspace --exclude rusty-bacnet --all-targets --locked` — exit 0
- `cargo check -p rusty-bacnet --tests` — clean
- `check-file-size.sh` — all files within the 700-LOC cap
- **Mutation-proven:** restoring the old gate (`if !distribute { … }` at both
  sites) fails all 7 new/rewritten tests and no others.


jscott3201:
**Spec-citation correction, applied before merge.**

The first revision of this PR claimed Clause 13.2.2.1.4's four transition actions were: store `Event_State`, timestamp in `Event_Time_Stamps`, **update `Acked_Transitions`**, and **issue the notification** — with only the fourth scoped to `Event_Enable`. A Codex spec-review lane cross-checking the clause text flagged that as wrong, and it was.

The actual four actions are:

1. store the new `Event_State` (specifically — "it is not acceptable to set Event_State to OFFNORMAL when the returned value is HIGH_LIMIT")
2. store the time of the transition in `Event_Time_Stamps`
3. store the message text in **`Event_Message_Texts`**, if present
4. indicate the transition to **both** the alarm-acknowledgment process (13.2.3) **and** the event-notification-distribution process (13.2.5)

So: there is no "distribution action" in the list, **none** of the four is `Event_Enable`-scoped, and `Acked_Transitions` is not a transition action at all — Clause 13.2.3 maintains it from the unconditionally-indicated transition, gated by `Ack_Required`.

The gate lives in Clause 13.2.5: *"The external distribution of notifications can be disabled by the Event_Enable property."* Clause 12.12 concurs, defining the property as flags that *"separately enable and disable the distribution of TO_OFFNORMAL, TO_FAULT, and TO_NORMAL notifications (see Clause 13.2.5)."*

**The code change is unaffected — and better supported than originally argued.** Since none of the four actions is `Event_Enable`-scoped, and action 4 feeds *two* downstream processes, returning `None` for a suppressed transition denied the alarm-acknowledgment process as well as distribution. Only the justifying prose was wrong; it has been corrected in the commit message, PR body, CHANGELOG, rustdoc, and code comments. Issue #200's spec basis was corrected in place too.

Recording this rather than silently editing, because a wrong clause citation in a merged changelog is exactly the kind of thing that gets trusted later.

jscott3201:
**Adversarial review: 0 blockers, 1 MAJOR finding — accepted and fixed in b41022c.**

The MAJOR finding was a **coverage regression this PR introduced**, not a defect in the change. The single property #136 exists to preserve — that `Event_Enable` still suppresses the outbound notification — was pinned by zero tests.

I reproduced it before fixing rather than taking it on trust. Replacing `if outcome.distribute {` with `if true {` at **both** send sites:

```
cargo test --workspace --exclude rusty-bacnet --no-fail-fast
-> 2236 passed, 0 failed
```

Nothing detected it. And this is a *regression* rather than a pre-existing gap: on `dev` the gate lived inside `fire()`, so a suppressed transition returned `None` and a notification was literally unbuildable — `event_enable_zero_suppresses_all_notifications` caught the equivalent mutation. Moving the gate to the send site moved it out from under the tests.

Three tests added, each proven against the mutation it guards:

| test | mutation | result |
|---|---|---|
| `event_enable_cleared_suppresses_per_write_send` | `if true` | FAILS — "got 1 broadcast(s)" |
| `event_enable_set_permits_per_write_send` | `if false` | FAILS |
| `event_enable_cleared_suppresses_periodic_time_delay_send` | `if true` | FAILS — "got 1 broadcast(s)" |

The second exists because suppression alone is satisfied by a gate stuck permanently closed; both directions together pin the gate to `Event_Enable` rather than to a constant. The third drives the real spawned `intrinsic_reporting_task` on a paused clock — the per-write tests cannot reach that site, because a nonzero `Time_Delay` makes the per-write probe return `None` by design.

Also from the review:

- **#204 filed** — `IntrinsicReportingEngine` is a *third* `Event_Enable` path with the same defect (`pub mod`, zero in-workspace callers). Not fixed here: separate subsystem, own test suite, and whether it should exist at all is a semver call. The CHANGELOG no longer says "both reporting families" unqualified.
- The breaking-change bullet now records the **silent** half of the break: the new `pub` field is a compile error a consumer cannot miss, but `evaluate_event_enrollments` returning previously-filtered transitions is invisible — a per-element consumer starts emitting suppressed notifications with nothing failing to compile.
- Renamed `out_of_range_event_enable_suppresses_notification`; it no longer suppresses anything observable.
- **#203** strengthened: `LimitEnable::to_bits` in the *same file* packs MSB-first correctly (`bits |= 0x80; // bit 0 (MSB first)`), so the LSB-first sites are demonstrably a bug rather than a house convention.

**Verification note.** The mutation-proof command in the PR body aborts at the first failing crate. The 3 enrollment failures only appear with `--no-fail-fast`; without it you see 4 and would conclude the claim is wrong. My own first run hit exactly that.

Gate on b41022c: **2239 passed, 0 failed**, clippy 0, fmt clean, pyo3 check clean, file-size cap OK.

Issue #206, opened 2026-07-27
GetEventInformation ignores selection criterion (b): objects in NORMAL with unacknowledged transitions are never reported


jscott3201:
## Summary

`handle_get_event_information` implements only the first of the two selection
criteria Clause 13.12.1.4 defines. It returns objects whose `Event_State` is not
NORMAL, but never returns an object that is back in NORMAL while still carrying
an unacknowledged transition.

## Spec basis

ASHRAE 135-2020 Clause 13.12.1.4, GetEventInformation Service Procedure
(spec lines 45154-45161):

> After verifying the validity of the request, the responding BACnet-user shall
> search for all event-initiating objects that do not have an
> Event_Detection_Enable property with a value of FALSE and that meet the
> following conditions [...]
>
> (a) have an Event_State property whose value is not equal to NORMAL, **or**
> (b) have an Acked_Transitions property that has at least one of the following
>     bits (TO_OFFNORMAL, TO_FAULT, TO_NORMAL) set to FALSE.

The two are disjunctive. Clause 13.10.1.2.1.1 confirms the same idea from the
other side for GetAlarmSummary's stricter criteria, and Clause 13.12.1.2.1.1
describes the returned set as objects "whose value is not equal to NORMAL **or**
[that have] an Acked_Transitions property that has at least one of the following
bits [...] set to FALSE".

## Evidence

`crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs` selects
on `Event_State` alone:

```rust
if let Ok(PropertyValue::Enumerated(state)) =
    object.read_property(PropertyIdentifier::EVENT_STATE, None)
{
    if state != bacnet_types::enums::EventState::NORMAL.to_raw() {
        // ... build summary
    }
}
```

There is no `Acked_Transitions` read anywhere in the handler.

## Impact

This is the normal operator workflow for an alarm that has cleared but not been
acknowledged. A device returns to NORMAL, the operator has not yet acknowledged
the TO_OFFNORMAL transition, and a workstation polling GetEventInformation is
told there is nothing outstanding — so the unacknowledged event silently
disappears from the alarm console.

GetEventInformation is the one summarization service Clause 13.2.4 (spec line
41509) *requires* notification-servers to support, which makes this the more
consequential of the summarization gaps:

> Notification-servers are required to support execution of the
> GetEventInformation service.

## Reachability: latent until #123

Nothing currently clears a bit in `Acked_Transitions`. The Clause 13.2.3
[...]

Issue #205, opened 2026-07-27
AlertEnrollment ignores the Clause 13.2.2.1 reset when Event_Detection_Enable is written FALSE


jscott3201:
## Summary

`AlertEnrollmentObject` exposes `Event_Detection_Enable` and holds an
`Event_State`, but writing the property FALSE performs none of the reset
ASHRAE 135-2020 requires. The object keeps reporting its last detected alarm
state while detection is off.

`EventEnrollmentObject` got this right in #164/#206; `AlertEnrollmentObject`
lives in the same file and did not.

## Spec basis

Clause 13.2.2.1 (Event-State-Detection State Machine), spec line 41257:

> If the Event_Detection_Enable property is FALSE, then this state machine is
> not evaluated. In this case, no transitions shall occur, **Event_State shall
> be set to NORMAL**, and Event_Time_Stamps, Event_Message_Texts and
> Acked_Transitions shall be set to their respective initial conditions.

Clause 13.2.2 states it as a standing condition rather than a one-shot action
(spec line 41200):

> When the event-state-detection process is disabled via the
> Event_Detection_Enable, both the event algorithm and the Reliability value are
> ignored, and Event_State **remains** NORMAL.

## Evidence

`crates/bacnet-objects/src/event_enrollment/mod.rs`, `AlertEnrollmentObject`:

```rust
// write path
if property == PropertyIdentifier::EVENT_DETECTION_ENABLE {
    if let PropertyValue::Boolean(v) = value {
        self.event_detection_enable = v;   // <-- no reset of event_state
        return Ok(());
    }
    return Err(common::invalid_data_type_error());
}
```

The struct carries `event_state: u32` alongside `event_detection_enable: bool`,
so the two can disagree. Nothing forces `event_state` back to NORMAL, and
nothing prevents it being set to a non-NORMAL value afterwards.

## Impact

`Event_State` is a readable property, so a disabled AlertEnrollment answers
ReadProperty with an alarm state the standard says it must not be in. It also
leaks into the event-summarization services: those are now filtered on
`Event_Detection_Enable` directly (#164), so this particular object is excluded
regardless — but that is the *filter* saving it, not the object being correct.
Any future consumer that reasonably trusts the Clause 13.2.2 invariant
("Event_State remains NORMAL") would be wrong about this object.

## Proposed fix

Mirror what `EventEnrollmentObject` now does:

1. On a write of FALSE, force `event_state` to NORMAL (and reset
[...]

Issue #123, opened 2026-07-14
Maintain Acked_Transitions and Event_Time_Stamps when an event fires


jscott3201:
### Affected revision

Observed on rusty-bacnet v0.10.1 at commit `8b8c0cba3e92caa30e145c91abedfa003c64f2c9`.

### Behavior

Event-capable objects initialize three acknowledgement bits and timestamp slots, but transition evaluators do not clear the fired acknowledgement bit or update the corresponding timestamp. Analog acknowledgement handlers can set a bit, but no reachable transition path first clears it.

This leaves newly fired events observable as already acknowledged and their transition timestamps unchanged.

### Suggested direction

- clear the applicable bit when an acknowledgement-requiring transition fires;
- update the matching timestamp slot;
- preserve the selected `BACnetTimeStamp` alternative;
- add transition and acknowledgement round-trip tests for all three coordinates.

### Relevant code

- `crates/bacnet-objects/src/event.rs`
- `crates/bacnet-objects/src/analog/mod.rs`
- `crates/bacnet-objects/src/common.rs`

jscott3201:
Closing as implemented on clean `dev@3825f9bcd9ecd4849d74a9bfadc49a88a0a4a09d`.

Provenance: PRs #444 (`02ccb783`), #445 (`6acda25a`), and #446 (`b99fc582`) established the atomic transition kernel, intrinsic producer migration, and committed-history EventNotification source. PRs #451 (`f1eb5152`), #493 (`08b05fcd`), #505 (`1b7bd13b`), and #506 (`9c8a3383`) completed Event Enrollment lifecycle/correlation, all supported object-family acknowledgment, and acknowledgment-notification coverage.

Current behavior:
- a fired transition atomically updates `Event_State`, the matching `Event_Time_Stamps` coordinate, and the matching `Acked_Transitions` bit;
- acknowledgement-required transitions clear that coordinate, while non-required transitions set it;
- Time, SequenceNumber, and DateTime timestamp alternatives are preserved through commit, notification encoding, and correlated acknowledgment;
- intrinsic and Event Enrollment producers use the shared commit path, and outbound notifications read committed history;
- AcknowledgeAlarm correlation sets only the exact stored transition coordinate.

Focused validation passed at this SHA for all three commit/timestamp coordinates, exact timestamp-choice acknowledgment, all nine atomic intrinsic families, committed wire history, Event Enrollment service round trips, supported-family correlation, and device-clock DateTime commits.

PR #207, opened 2026-07-27
feat(objects,server): implement Event_Detection_Enable on Event Enrollment (#164)

### LINKED ISSUE #164: Implement Event_Detection_Enable on Event Enrollment
### Affected revision

Observed on `dev` at commit `4ee4d6a`.

### Behavior

`EVENT_DETECTION_ENABLE` is implemented only on the forwarder object (`crates/bacnet-objects/src/forwarder.rs`). The Event Enrollment object neither exposes the property nor is it consulted by the server's enrollment evaluator.

Clause 13.2.2.1 (p641) requires that when `Event_Detection_Enable` is FALSE the event-state-detection state machine is not evaluated and `Event_State` shall be set to NORMAL, with the associated timestamp and acknowledgment state reset accordingly.

Without the property there is no conformant per-object way to disable detection. The server-level `enable_event_enrollment` switch added in #133 stops the evaluation task but performs none of the required reset, so disabling evaluation leaves every enrollment holding whatever state it last detected. That switch is an implementation control over a background task, not the normative property, and is documented as such.

### Suggested direction

- add `Event_Detection_Enable` to the Event Enrollment object with its required default;
- skip evaluation for an enrollment whose flag is FALSE;
- perform the Clause 13.2.2.1 reset on the transition to FALSE rather than freezing state;
- keep the server-level switch as a documented implementation control, distinct from the property;
- test the disabled reset, re-enable behavior, and that a disabled enrollment emits no notifications.

### Relevant code

- `crates/bacnet-objects/src/event_enrollment/mod.rs`
- `crates/bacnet-server/src/server/mod.rs`

jscott3201:
Batch 2, PR 7. Closes #164.

## Spec position

Table 12-14 codes `Event_Detection_Enable` **R** on Event Enrollment — required,
with no footnote qualifier — and it was implemented only on
`NotificationForwarderObject` and `AlertEnrollmentObject`. The one object type
where the standard *requires* it was missing it, so there was no conformant
per-object way to disable detection.

Three requirements, from three different clauses:

| Clause | Requirement |
|---|---|
| 12.12 / Table 12-14 | property is **R** (required, present + readable) |
| 13.2.2 / 13.2.2.1 | while FALSE: not evaluated, no transitions, `Event_State` **remains** NORMAL, history properties at initial conditions |
| 13.10.1.4, 13.11.1.4, 13.12.1.4 + Table 13-2 | all three summarization services exclude the object — each states it independently |

## The disabled state is an invariant, not a one-shot action

Clause 13.2.2 puts it in the continuous tense:

> When the event-state-detection process is disabled via the
> Event_Detection_Enable, both the event algorithm and the Reliability value are
> ignored, and Event_State **remains** NORMAL.

An edge-only reset would be unsound — it never fires for an object constructed
or restored from persistence with the property already FALSE. So it is enforced
at three points, deliberately:

```
write_property(FALSE)        -> reset Event_State to NORMAL immediately, not on
                                some later evaluation pass
set_event_state_internal     -> refuses any non-NORMAL state while detection is
                                off, so the invariant holds by construction
evaluate_event_enrollments   -> skips disabled enrollments, per Clause
                                13.2.2.1's separate "this state machine is not
                                evaluated"
```

**Being explicit about the third one:** it is *not* independently observable.
Deleting the evaluator guard fails no test, because the object-level refusal
already blocks the transition and the evaluator only records one when
`set_event_state_internal` succeeds. I verified that by mutation and documented
it at the guard rather than leaving a reader to assume coverage. It earns its
place by implementing the clause's first sentence literally (the algorithm
genuinely does not run) and by not swallowing an `Err` on every pass, forever.

## Summarization

Each of the three services repeats the exclusion in its own Service Procedure,
so each is filtered and each is tested separately. It is checked **directly**,
not inferred from `Event_State == NORMAL`, for two concrete reasons:

- `GetEnrollmentSummary` applies no default `Event_State` filter, so the
  exclusion cannot fall out of the forced-NORMAL invariant there.
- `AlertEnrollmentObject` exposes the property but does not apply the reset
  (filed as **#205**), so inferring it would leak a disabled object.

**Absence of the property means enabled**, per Clause 13.12.1.4's deliberate
double negative — objects that "do not have an Event_Detection_Enable property
with a value of FALSE". The inverse reading would silently empty these responses
for the ~64 object types that do not model the property. That case has its own
test, proven by mutating the helper to be absence-intolerant.

## Writability

Accepted, and `is_writable_property` is overridden so PICS reports what dispatch
actually accepts (the inherited historical default omitted every property this
object writes). R does not imply writable — but Clause 12.1.2 permits an R
property to be writable "at the implementor's option unless specifically
prohibited in the text describing that particular standard object's property",
and Clause 12.12 prohibits nothing: "is expected to be set during system
configuration" is guidance, not a "shall". Annex K's AE-AVM-A BIBB (K.2.17 with
Table K-17) positively *requires* a conforming workstation to write it, without
the footnote-1 present-only exemption that covers `Event_State`,
`Acked_Transitions`, `Event_Time_Stamps` and `Event_Message_Texts`.

## A correction to the issue text

#164 says to add the property "with its required default". **There is no
required default** — 135-2020 specifies none anywhere, and Clause 15.3.1.4 makes
the initial values of properties not named in a CreateObject request "a local
matter". TRUE is a project choice: it preserves the always-detecting behavior
this object had before the property existed, and matches the two objects that
already implemented it. Recorded on the issue.

## Verification

17 tests. Mutation matrix, each mutation caught by exactly its intended tests
and no others:

| Mutation | Result |
|---|---|
| summarization filter removed | 3 exclusion tests FAIL, absence test passes |
| helper made absence-intolerant | 3 exclusion tests pass, absence test FAILS |
| write-time reset removed | 2 object tests FAIL |
| `set_event_state_internal` guard removed | 1 object test FAILS |
| evaluator guard removed | nothing fails — documented at the guard |

Gates: `cargo test --workspace --exclude rusty-bacnet --no-fail-fast` **2256
passed, 0 failed**; clippy exit 0; `cargo fmt --all --check` clean;
`cargo check -p rusty-bacnet --tests` clean; file-size cap OK (tightest file 542/700).

Conformance ledger needs no regeneration — verified that
`docs/conformance/bacnet-135-2020.json` holds no per-object property lists (35
clause-group rows), so the staleness test at `conformance_ledger.rs:369` is
unaffected.

## Filed while implementing

- **#205** — `AlertEnrollmentObject` ignores the same Clause 13.2.2.1 reset when
  the property is written FALSE.
- **#206** — `GetEventInformation` ignores selection criterion (b): objects back
  in NORMAL with an unacknowledged transition are never reported. That is the
  normal cleared-but-unacknowledged operator workflow. Latent until #123 makes
  `Acked_Transitions` mutable.

## Checked and deliberately *not* filed

`property_list()` includes `OBJECT_IDENTIFIER` / `OBJECT_NAME` / `OBJECT_TYPE`,
which Clause 12.12 says "are not included in the list". A review lane flagged
this as a conformance defect; it is not. `common::read_property_list_property`
already filters all four at the serving boundary, so the wire output is
conformant and the trait method is an internal superset. Verified before filing.


jscott3201:
**Self-review found this PR overclaiming; fixed in 78c7869.**

The PR said the Clause 13.2.2.1 invariant holds "by construction". It did not. `EventEnrollmentObject::set_event_state` is `pub` and routes around both guards — it is neither `write_property` (so no reset fires) nor `set_event_state_internal` (so no refusal). A caller could leave the object with `Event_Detection_Enable` FALSE and `Event_State` HIGH_LIMIT, exactly what Clause 13.2.2 forbids: *"Event_State **remains** NORMAL"*.

The tell was an internal contradiction in my own claims: the summarization test fixture **deliberately built that inconsistent state** through this setter, while the CHANGELOG asserted it was unreachable. Both could not be true.

`set_event_state` now ignores a non-NORMAL seed while detection is off, so all three paths that can change `Event_State` are covered and the claim is accurate:

| Path | Behavior while detection is FALSE |
|---|---|
| `write_property(EVENT_DETECTION_ENABLE, false)` | resets `Event_State` to NORMAL immediately |
| `set_event_state_internal` (evaluator) | returns an error |
| `set_event_state` (public seeder) | ignores a non-NORMAL value |

**Knock-on: the summarization fixture had to change, and is better for it.** With the state now unreachable for a real enrollment, a fixture built from one would prove only that the reset works — not that the services filter on the property. Replaced with a minimal mock reporting `EVENT_STATE=HIGH_LIMIT` and `EVENT_DETECTION_ENABLE=FALSE` together.

That is the right fixture on the merits: the exclusion is stated independently in each of the three Service Procedures rather than derived from `Event_State`, and the inconsistent state **is** reachable in the wild via `AlertEnrollmentObject`, which applies no reset (#205). The filter has to be tested directly against something that can actually be inconsistent.

Mutation-verified after the change:

```
seeder guard removed          -> disabled_detection_ignores_non_normal_seed FAILS
summarization filter -> true  -> 3 exclusion tests FAIL, absence test passes
```

Gate on 78c7869: **2258 passed, 0 failed**, clippy 0, fmt clean, pyo3 clean, file-size OK.

jscott3201:
**Codex spec review: all 8 core claims CONFIRMED, 7 citation defects — fixed in 507632f.**

| Claim | Verdict |
|---|---|
| C1 conformance code R on Event Enrollment | CONFIRMED |
| C2 no default value specified anywhere | CONFIRMED |
| C3 standing condition, not an edge action | CONFIRMED |
| C4 reset targets + `Acked_Transitions` initial = all TRUE | CONFIRMED |
| C5 all three services exclude independently | CONFIRMED |
| C6 absence means INCLUDED | CONFIRMED (qualified: other criteria still apply) |
| C7 writability permitted; AE-AVM-A requires client write capability | CONFIRMED (qualified: server acceptance permitted, not mandated) |
| C8 FALSE suppresses reliability/FAULT detection too | CONFIRMED |

**One finding was substantive and I had it wrong.** Three places said the property is required *"only on Event Enrollment"*. **Table 12-61 codes it R on Alert Enrollment too**, alongside `Event_State` R and `Acked_Transitions` R — I verified that directly against the table rather than taking it on trust. That makes **#205 a required-property conformance gap**, not an optional-property nicety, and the issue has been updated accordingly.

**Five were subclause numbers I could not verify and therefore should not have asserted.** The reviewer says the Service Procedures are 13.10.2 / 13.11.2 / 13.12.2 and CreateObject is 15.3.2; an earlier lane gave me 13.10.1.4 / 13.11.1.4 / 13.12.1.4 / 15.3.1.4. **My authoritative source cannot settle it** — the text extraction strips the number from all 45 "Service Procedure" headings, the table of contents stops at the 13.x level, and the document contains no cross-reference to those subclauses.

Rather than pick a side on unverifiable evidence, these now cite **by title**: "Clause 13.10, Service Procedure", "Clause 15.3's CreateObject Service Procedure". Unambiguous to a reader, and correct under either numbering. Same treatment for the `Event_Algorithm_Inhibit` re-arm rule (13.2.2.2 vs the reviewer's 13.2.2.1.5).

**The seventh was a real over-generalization.** The test module said all three Service Procedures use "shall be ignored". Only 13.10 and 13.11 do — 13.12 folds it into the search predicate as a double negative. That distinction is not cosmetic: it is exactly what makes *absence* of the property mean *included*, which is the behavior `summarization_includes_objects_without_the_property` pins.

Gate on 507632f: **2258 passed, 0 failed**, clippy 0, fmt clean, CI green.

*(This comment was reposted — the first version had shell-mangled content.)*

jscott3201:
**Adversarial pass — I ran it myself.** The spawned review lane went idle twice without delivering a report, so rather than a third round-trip I ran the checks directly. Being explicit about that: the findings below are mine, verified by running, not an independent second opinion. The Codex spec lane *did* deliver and is reported above.

### Mutation matrix, re-run on this head, workspace-wide with `--no-fail-fast`

Baseline: 2258 passed. Working tree verified clean after every mutation.

| Mutation | Failed | Which |
|---|---|---|
| write-time reset removed | **2** | the 2 intended |
| `set_event_state_internal` guard removed | **1** | intended |
| public `set_event_state` seeder guard removed | **1** | intended |
| summarization filter → `true` | **3** | the 3 exclusion tests |
| helper made absence-intolerant | **3** | mine + **2 pre-existing** `device_event` tests |
| evaluator guard removed | **0** | as documented at the guard |

**Correction to the PR body:** I listed the absence-intolerant mutation as failing 1 test. It fails **3** — two pre-existing `device_event` tests catch it as well. Absence-tolerance was better covered than I claimed, and the earlier "no existing test would catch it" assessment was wrong.

### The load-bearing claim, verified by enumeration

I tried to find any reachable path leaving an Event Enrollment with `Event_Detection_Enable` FALSE and `Event_State` non-NORMAL. There are exactly three assignments to `self.event_state`: the reset itself, the public seeder (guarded), and the internal lifecycle setter (guarded). The flag is a **private** field mutated at exactly one site, which triggers the reset. No `serde`, no `Default`, no `From` — no deserialization backdoor. `db.get_mut` hands out `&mut dyn BACnetObject`, so only trait methods are reachable, and those are guarded. The claim holds.

### One real finding, and it was mine

I probed `is_writable_property` against what `write_property` actually accepts, across 14 properties:

```
EventEnrollmentObject   OBJECT_NAME  claims=true  accepts=false   <<< MISMATCH
AnalogInputObject       OBJECT_NAME  claims=true  accepts=true
```

Event Enrollment has no `OBJECT_NAME` arm, so **it cannot be renamed** while every core I/O/V type can. My doc comment claimed the override "mirrors the `write_property` arms" — false for this one property.

**No regression, though:** the inherited `historical_writable_default` (traits.rs:228) already returned `true` for `OBJECT_NAME`, so PICS reports exactly what `dev` reported. I left the behavior alone rather than silently changing PICS, documented the exception at the override, and filed the rename gap as **#208** — making the object renameable needs its own tests (rename, name-index refresh, `DUPLICATE_NAME` rejection) and is not #164's business.

Gate on 608414c: **2258 passed, 0 failed**, clippy 0, fmt clean, file-size OK.

jscott3201:
**Merged as e933bf1 on owner decision, with the review gap recorded rather than papered over.**

To be explicit about what this PR did and did not clear:

| Gate | Result |
|---|---|
| CI (5 jobs) | pass |
| Tests | 2258 passed, 0 failed |
| Clippy / rustfmt / pyo3 check / file-size cap | clean |
| Codex **spec** review | delivered — 8/8 core claims confirmed, 7 citation corrections applied |
| Independent **adversarial** review | **not obtained** |

Four attempts at an independent adversarial pass — three agent lanes and one Codex adversarial run — all terminated without producing a report. The adversarial analysis reported on this PR was run by the same author as the change, which is weaker by construction and is not a substitute. The owner elected to merge on that basis; noting it here so the record is accurate rather than implying a gate that was never met.

That self-review was not empty, for what it is worth. It reproduced the full mutation matrix workspace-wide with `--no-fail-fast`, verified the load-bearing invariant by enumerating every write path to `event_state`, and found two real defects that are now filed:

- **#208** — Event Enrollment has no `OBJECT_NAME` write arm, so it cannot be renamed while every core I/O/V type can, and PICS claims otherwise. Found by empirically probing `is_writable_property` against actual `write_property` behavior across 14 properties.
- **#209** — a `WritePropertyMultiple` writing `Event_Detection_Enable = FALSE` that then fails leaves `Event_State` cleared, because the rollback snapshots only the written property. Reproduced, not reasoned. The invariant still holds; it is an atomicity gap, and it sharpens once #123 lands.

**#209 and #205 both point at a better shape than what merged here:** stop *storing* the disabled condition and compute it at read time instead. Nothing is mutated, so there is nothing to roll back and no bypass to guard — and it matches Clause 13.2.2's standing-condition phrasing ("Event_State *remains* NORMAL") more faithfully than a reset-on-write does. That would make the four guards in this PR unnecessary rather than merely correct. Worth deciding once and applying to both enrollment objects.

PR #213, opened 2026-07-27
feat(client,server): use separate ports for broadcast vs unicast

amab8901:
broadcast port should default to 47808 so that every UDP endpoint knows in advance which port to listen to for broadcasts.

unicast port should default to 0 to assign ephemeral port so that only the intended recipients listen to the unicast instead of everybody getting spammed.

So far, a single port was used for both broadcast and unicast. This forced you to choose between broadcasts working and unicasts not (by setting port to 47808) or unicasts working and broadcasts not (by setting port to 0). This PR fixes it. Tested against Yabe simulator to confirm the fix.

### DIFF
--- a/benchmarks/src/bin/bacnet_device.rs
+++ b/benchmarks/src/bin/bacnet_device.rs
@@ -24,9 +24,13 @@ struct Args {
     #[arg(long, default_value = "0.0.0.0")]
     interface: String,
 
-    /// UDP port (BIP only)
+    /// UDP broadcast port (BIP only)
+    #[arg(long, default_value_t = 0)]
+    unicast_port: u16,
+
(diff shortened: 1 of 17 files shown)


jscott3201:
I'm working through the issues backlog and will check on this. Check the CI it looks like its my file cap rule getting tagged. 700 LOC per file cap on Rust files.

Edit: Tests look good so once that is fixed should be no issue!

amab8901:
don't merge this. It looks like it was a simulator bug on Yabe where I was able to reach the simulator via broadcast the 1st time but not the 2nd time, and switching simulator address resets the bug so u can read it. I was able to both broadcast and unicast with a commit before this one in a later test

PR #215, opened 2026-07-27
fix(objects,server): drive Event_State from Reliability (#167)

### LINKED ISSUE #167: Reliability never drives Event_State, so no TO_FAULT transition is reachable
### Affected revision

### Behavior

`FaultDetector::evaluate` (`crates/bacnet-server/src/fault_detection.rs:55-148`) is the only reliability producer in the stack. It writes `PropertyIdentifier::RELIABILITY` at `:129-135` and returns `ReliabilityChange` records. The periodic task that drives it consumes those records in a `debug!` and nothing else (`crates/bacnet-server/src/server/lifecycle.rs:381-397`, log at `:390-396`). Nothing converts a Reliability other than `NO_FAULT_DETECTED` into `Event_State = FAULT`.

No detector can produce FAULT either. `OutOfRangeDetector::compute_new_state` (`crates/bacnet-objects/src/event.rs:301-340`), `ChangeOfStateDetector::compute_new_state` (`:452-458`) and `CommandFailureDetector::compute_new_state` (`:567-573`) return only NORMAL, HIGH_LIMIT, LOW_LIMIT or OFFNORMAL, and `grep -i reliab crates/bacnet-objects/src/event.rs` returns nothing — the detectors never see the `reliability` field, which lives on the object (`crates/bacnet-objects/src/analog/mod.rs:33`). A repo-wide `grep EventState::FAULT` outside tests yields only consumers (`crates/bacnet-objects/src/event.rs:43`, `crates/bacnet-server/src/handlers/alarm_event.rs:134`), never a producer.

Clause 13.2.2, Event-State-Detection (p640): "The event algorithm determines the normal or offnormal states and the Reliability property determines whether or not the event state will indicate a fault. Fault detection takes precedence over the detection of normal and offnormal states. As such, when Reliability has a value other than NO_FAULT_DETECTED, the event-state-detection process will determine the object's event state to be FAULT." The same clause adds "In objects that support event-state-detection, the reporting of changes in Reliability cannot be disabled while event-state-detection is enabled." Clause 13.2.2.1 (p641-642) then places a ToFault transition out of the Normal, OffNormal and Fault states.

Two consequences:

- No TO_FAULT event notification is ever generated, so `Event_Time_Stamps[2]` and the TO_FAULT bit of `Acked_Transitions` can never move. The round-trip coverage requested in #123 cannot be written for TO_FAULT until this is fixed.
- `compute_status_flags` (`crates/bacnet-objects/src/common.rs:13-39`) sets FAULT from `reliability != 0` at `:25-29` but IN_ALARM from `event_state != 0` at `:20-24`. Because Reliability never reaches `Event_State`, a faulted object reports FAULT TRUE with IN_ALARM FALSE. Clause 12.2 Status_Flags (p169) defines them as "IN_ALARM  Logical FALSE (0) if the Event_State property has a value of NORMAL, otherwise logical TRUE (1)" and "FAULT  Logical TRUE (1) if the Reliability property is present and does not have a value of NO_FAULT_DETECTED", which the standard reconciles by requiring Reliability to drive `Event_State` in the first place.

`enable_fault_detection` defaults to `false` (`crates/bacnet-server/src/server/mod.rs:287`), so on a default server Reliability never changes at all; turning it on changes Reliability but still not `Event_State`.

### Suggested direction

- evaluate reliability first in the event-state-detection path and hold `Event_State` at FAULT while Reliability is not `NO_FAULT_DETECTED`, per 13.2.2;
- route the resulting transition through the same path as offnormal and normal transitions so `Event_State`, `Event_Time_Stamps`, `Acked_Transitions` and the notification are all driven (#123);
- on the transition out of FAULT, re-derive the state from the event algorithm rather than assuming NORMAL;
- feed the `ReliabilityChange` records from the periodic task into that path instead of logging them, and cover the case where Reliability is set by a local or network write rather than by the detector;
- test a Reliability change that drives FAULT, the TO_FAULT notification, the resulting `Status_Flags`, and the return out of FAULT.

### Relevant code

- `crates/bacnet-server/src/fault_detection.rs`
- `crates/bacnet-server/src/server/lifecycle.rs`
- `crates/bacnet-objects/src/common.rs`

### LINKED ISSUE #200: CommandFailureDetector suppresses TO_FAULT distribution regardless of Event_Enable

jscott3201:
## Why

Nothing in the stack converted a `Reliability` other than `NO_FAULT_DETECTED` into `Event_State = FAULT`. `FaultDetector::evaluate` wrote the property and the periodic task consumed its `ReliabilityChange` records in a `debug!` and nothing else; a repo-wide search for a producer of `EventState::FAULT` outside tests found **zero** matches.

Two consequences:

- No TO_FAULT notification could ever be generated, so `Event_Time_Stamps[2]` and the TO_FAULT bit of `Acked_Transitions` could never move.
- A faulted object reported `Status_Flags` FAULT TRUE with IN_ALARM FALSE. Clause 12.2 derives IN_ALARM from `Event_State` and FAULT from `Reliability`, and the standard reconciles them by requiring `Reliability` to drive `Event_State` upstream.

It also means **#211's `CHANGE_OF_RELIABILITY` branch, merged last session, was dead code** — correct, but unreachable. This makes it live.

Closes #167. Closes #200.

## The design changed after a spec pass — read this part

A normative reading of Clause 13.2.2 / 13.2.2.1 was run **before** implementation. It refuted the approach this issue's own text recommends.

Issue #167's "Suggested direction" says: *"on the transition out of FAULT, re-derive the state from the event algorithm rather than assuming NORMAL."* That is backwards. Clause 13.2.2.1 (Fault, ToNormal):

> "If reliability-evaluation indicates a value of NO_FAULT_DETECTED, then perform the corresponding transition actions and **enter the Normal state**."

So the obvious implementation — treat FAULT as an overlay, fall back to the algorithm when it clears — emits `FAULT -> HIGH_LIMIT` when the present value is still out of range, a transition the state machine does not define. `recovery_from_fault_enters_normal_not_the_algorithm_state` is the test that fails under that design.

Other findings folded in:

| Question | Reading |
|---|---|
| Delay on entry to FAULT? | **No.** Clause 13.2.2.1's ToFault is unconditional; `Time_Delay` is an algorithm parameter (13.3.1), and the algorithm is what fault detection overrides |
| Is FAULT standing or edge-triggered? | **Standing.** Clause 13.2.2.1: "In the Fault state reliability-evaluation indicates a value other than NO_FAULT_DETECTED" |
| Pending countdown when a fault arrives? | **Standard is silent.** See below |

## What

FAULT is modeled as a **standing condition derived from `Reliability`**, not a latched state. Nothing about the fault is stored; it is re-derived on every evaluation.

That has a useful consequence: a `Reliability` set by *any* route — the server's fault detector, a local write, or a network write — reaches event-state-detection without that route notifying anything, because the 1-second intrinsic-reporting task re-reads it. No new plumbing was needed in the server at all. `enable_fault_detection` (still `false` by default) governs only whether `Reliability` is *derived* from limits, never whether an existing one is honored.

The rule lives once, in a shared `fault_precedence` function all three detectors consult, so a detector added later that omits it is a visible omission rather than a silently missing clause. It reaches all nine intrinsically-reporting object types through the existing `impl_intrinsic_reporting!` macro.

### One decision that is not a citation

The standard does not say what becomes of an in-flight `Time_Delay` countdown when a fault arrives — Clauses 13.2.2, 13.2.2.1 and 13.3 are all silent. **Cancelling was a project decision**, made because recovery re-enters NORMAL, so a countdown seeded before the fault targets a transition out of a state the machine no longer occupies. The nearest analogous rule agrees in shape — Clause 13.2.2.1.5 restarts the *full* delay when `Event_Algorithm_Inhibit` clears rather than resuming a partial one — but it is scoped to inhibition, not faults. It is labeled as an analogy in the code, not dressed up as a quotation.

### Breaking API change

`OutOfRangeDetector`, `ChangeOfStateDetector` and `CommandFailureDetector` take a trailing `reliability: u32` on `evaluate` / `probe` / `tick`, and `impl_intrinsic_reporting!` takes a third argument naming the object's reliability field. Callers pass `Reliability::NO_FAULT_DETECTED.to_raw()` to preserve prior behavior.

### #200, fixed here rather than deferred

`CommandFailureDetector::fire` hardcoded `distribute: false` for FAULT, ignoring the TO_FAULT `Event_Enable` bit — its `TO_FAULT` constant carried `#[allow(dead_code)]` for exactly that reason. Clause 13.2.5 scopes `Event_Enable` to distribution uniformly across all three directions. The defect was unobservable while FAULT was unreachable and goes live the moment it is not, so leaving it would ship a known bug activated by this very change.

## Verification

**2280 passed / 0 failed**, 34 suites (`--features bacnet-transport/ipv6,bacnet-transport/sc-tls`). 18 new tests. fmt, clippy and the 700-LOC cap all clean.

**Mutation testing** — the reason two of these tests exist at all:

| Mutant | Result |
|---|---|
| Recovery re-derives from the algorithm (the issue's own suggestion) | **killed** by 7 tests |
| Entry to FAULT does not cancel the countdown | **killed** |
| Algorithm keeps running while FAULT holds | **killed** |
| `CommandFailure` TO_FAULT distribution hardcoded off again | **killed** by 2 |
| Precedence keyed on one known value instead of `!= NO_FAULT_DETECTED` | **killed** by 5 |
| Recovery does not clear the countdown | **survives — deliberately** |

The last one is unreachable: `EnterFault` already cleared `pending` and `HoldFault` returns before the algorithm can seed one. The line is kept and documented in place as the invariant the arm relies on, so it is not deleted later on the assumption a test covers it.

The hold guard initially looked tested but was not. The only test asserting it used `OutOfRangeDetector`, whose `compute_new_state` falls through to `_ => self.event_state` and is therefore inert in FAULT — the guard is redundant *there*. `ChangeOfStateDetector` is where it is load-bearing, and that test now exists.

## Deliberately not in scope

Clause 13.2.2.1 defines a FAULT **re-entry** — a transition generated when reliability-evaluation reports a *different* non-`NO_FAULT_DETECTED` value while already in the Fault state. The detectors do not retain the previous `Reliability`, and emitting a same-state transition depends on the same-state handling tracked in **#166**. Left out deliberately, and noted in the code.


jscott3201:
## Spec-accuracy gate: passed, with two follow-ups filed

An independent read-only normative review ran against this branch at `0bfe905`, reading the standard from the PDF rather than from memory. Six questions, all answered with verbatim quotations.

| Q | Verdict |
|---|---|
| Recovery enters NORMAL, not an algorithm-derived state | **CLEAR** — Clause 13.2.2.1 Fault/ToNormal quoted |
| No `Time_Delay` on entry to FAULT | **CLEAR** — ToFault is unconditional; 13.3.1 scopes the timer to offnormal conditions |
| Cancelling an in-flight countdown | **SILENT** — confirms the code's "project decision, not a citation" label is honest |
| `Event_Enable` now consulted for TO_FAULT (#200) | **CLEAR** — Clause 13.2.5 gives no TO_FAULT exemption |
| Steady hold while FAULT stands | **CLEAR** for unchanged Reliability |
| Anything else contradicting a clause | **CLEAR** — none found |

Two findings were raised that are **not** defects in this change, and both are now tracked rather than left as code comments:

- **#217** — the detectors cannot re-enter FAULT when Reliability changes to a *different* non-`NO_FAULT_DETECTED` value, which Clause 13.2.2.1 requires. This PR notes the omission in code; the reviewer's point was that separability does not make the detector fully conformant, which is fair, so it is filed.
- **#216** — Analog, Binary and Multi-state objects omit `Event_Detection_Enable` entirely, though their property tables require it when intrinsic reporting is supported ("These properties are required if the object supports intrinsic reporting"). Same class of gap that #164 closed for Event Enrollment. Worth noting the reviewer verified 3 of the 9 object tables and explicitly declined to generalize to the other 6; that caveat is carried into the issue.

Neither blocks this change. #167 makes FAULT reachable; #216 and #217 are about what else the standard asks for once it is.

jscott3201:
## Adversarial review: FIX-THEN-MERGE

Three independent lenses (state-machine correctness, API blast radius, test integrity) attacked the change, each instructed to refute rather than confirm. An adjudicator then re-verified every finding in an isolated scratch copy and ruled on it.

**No behavioral defect was found in the shipped code.** Ten findings, four rulings that matter.

### Confirmed — fixing here

Both are test gaps in code this PR introduced, reproduced empirically rather than argued:

1. **The `impl_intrinsic_reporting!` `tick` arm has no test coverage at all.** Replacing `self.$reliability_field` with a hardcoded `0` in that arm leaves the *entire workspace* green. `tick_intrinsic_reporting` has one production caller (`lifecycle.rs:508`) and zero test callers — and it is the sole route by which a `FaultDetector`-derived Reliability reaches event-state-detection, i.e. the feature's headline production path. The same mutation on the `probe` arm *is* killed by the object-level tests, so the asymmetry was an oversight.

2. **Both `pending = None` clears in `ChangeOfStateDetector::fault_step` can be deleted with 824/824 still passing.** The adjudicator went further than the reporter and measured non-equivalence directly: with `alarm_values=[1]`, `time_delay=5`, the clean code restarts the full delay and re-fires OFFNORMAL at tick 6, while the mutant resumes the stale countdown and fires at tick 2. So the documented ruling at `event.rs:352-360` is load-bearing and was unpinned. The existing test covers only `OutOfRangeDetector` — and `ChangeOfStateDetector` backs 6 of the 9 wired object types.

### Confirmed, but not this PR's to fix

**FAULT is genuinely unreachable on BI/BO/BV/MSI/MSO/MSV** — no `RELIABILITY` write arm, no setter, and `FaultDetector` covers only the analog three. The macro plumbing is correct and simply inert there. Pre-existing, and the correct fix is a spec question (is Reliability writable only while `Out_Of_Service` is TRUE?), so it is filed as **#218** rather than bolted on here.

### Rejected, with reasoning

- **DCC `DisableInitiation` halts detection** — real, but pre-existing, stalls *all* event-state movement rather than just fault, and this change strictly narrows the prior unconditional inconsistency. Filed as **#220**.
- **`FaultDetector` clobbers an externally written Reliability** — real, but the defect is in untouched code; the new code faithfully tracks the value it is given, and a TO_NORMAL is the correct response to a Reliability that genuinely returned to `NO_FAULT_DETECTED`. Filed as **#219**.
- **"Recovery costs an extra tick"** — the N+1 shape of tick-seeded countdowns predates this change and applies to any object updated through `set_present_value`. It fires late, never early, so Clause 13.3.1's minimum-dwell semantics are satisfied. The finding's supporting claim that the re-seed arm was previously unreachable is false.
- **`CommandFailureDetector` covers only 1 of 4 outcomes** — no object type wires that detector.
- **The vacuous `"algorithm must not run"` assertion** — equivalent mutant; the real hold coverage sits in the adjacent test. Message reworded anyway so it stops implying coverage it does not provide.

Follow-up issues from this review and the spec pass: **#216**, **#217**, **#218**, **#219**, **#220**.

Issue #203, opened 2026-07-27
Bit strings are encoded LSB-first, inverting TO_OFFNORMAL/TO_NORMAL and valid_days on the wire (Clause 20.2.10)


jscott3201:
## Summary

Every BACnet bit string this stack encodes is written **LSB-first within the
octet**, but Clause 20.2.10 requires the **first defined bit to occupy the most
significant bit**. The encoder shifts the internal value up to the top of the
octet without reversing it, so the bits arrive in the wrong order.

For `BACnetEventTransitionBits` this **swaps TO_OFFNORMAL and TO_NORMAL** on the
wire. A conformant client that enables TO_OFFNORMAL causes this server to
consult its TO_NORMAL bit, and vice versa.

## Spec basis

ASHRAE 135-2020, "Encoding of a Bit String Value" (Clause 20.2.10):

> Bit strings defined in this standard, e.g., the Status_Flags property, shall be
> encoded in the order of definition, with the **first defined Boolean value in
> the most significant bit, i.e. bit 7**, of the first subsequent octet. The bits
> in the bitstring shall be placed in bits 7 to 0 of the first subsequent octet,
> followed by bits 7 to 0 of the second subsequent octet [...]

`BACnetEventTransitionBits ::= BIT STRING { to-offnormal(0), to-fault(1), to-normal(2) }`
(Clause 21). So the required octet layout is:

| transition | defined index | required wire bit | required octet |
|---|---|---|---|
| TO_OFFNORMAL | 0 | bit 7 | `0x80` |
| TO_FAULT | 1 | bit 6 | `0x40` |
| TO_NORMAL | 2 | bit 5 | `0x20` |

## What the code emits

The internal representation is LSB-indexed (`bit0 = TO_OFFNORMAL`, `bit1 =
TO_FAULT`, `bit2 = TO_NORMAL`, masks `0x01`/`0x02`/`0x04`), and the encoder is a
bare left shift:

```rust
// crates/bacnet-objects/src/event_enrollment/mod.rs:155
data: vec![self.event_enable << 5],
// decode, same file:196
self.event_enable = byte >> 5;
```

Shifting aligns the 3-bit field to the top of the octet but preserves LSB-first
ordering inside it, so:

| transition | internal | `<< 5` emits | a conformant peer reads it as |
|---|---|---|---|
| TO_OFFNORMAL | `0x01` | `0x20` | **TO_NORMAL** |
[...]

Issue #168, opened 2026-07-27
FAULT transitions are reported with Event Type CHANGE_OF_STATE instead of the mandatory CHANGE_OF_RELIABILITY


jscott3201:
### Affected revision

Observed on `dev` at commit `3990f3e`.

### Behavior

`EventStateChange::event_type()` (`crates/bacnet-objects/src/event.rs:23-33`) returns `OUT_OF_RANGE` if either the from- or to-state is `HIGH_LIMIT` or `LOW_LIMIT`, and `CHANGE_OF_STATE` in every other case. `EventState::FAULT` is never considered. That value goes straight onto the wire at `crates/bacnet-server/src/server/event_notifications.rs:137`, which is the only production caller of `event_type()`.

Clause 13.2.5.3, Fault Event Notifications (p646), in full: "For all transitions to, or from, the FAULT state, the corresponding event notification shall use the Event Type CHANGE_OF_RELIABILITY." Table 13-3 (p646) says the same in its Event Type row: "When 'To State' or 'From State' is FAULT, set to CHANGE_OF_RELIABILITY, / Otherwise the value associated with the event-initiating object's event algorithm."

The transition classifier immediately below (`event.rs:40-48`) does special-case `EventState::FAULT` at `:43-44` to pick the Recipient_List and Priority coordinate, so the omission in `event_type()` reads as an oversight rather than a decision.

`EventType::CHANGE_OF_RELIABILITY` exists as an unused constant (`crates/bacnet-types/src/enums/object_level.rs:162`).

This is latent today: no production path produces `EventState::FAULT`, which is the subject of a separate issue on the reliability-to-`Event_State` gap. It becomes wire-visible the moment that lands or #127 propagates enrollment transitions. The mislabel alone misroutes a fault at the receiver; once #135 populates `event_values`, it also makes the receiver decode context tag [12] against the wrong `BACnetNotificationParameters` alternative (`event_values` is hardcoded `None` at `event_notifications.rs:151` today, so nothing is misdecoded yet).

The existing test does not catch it: `event_notifications.rs:574-600` builds NORMAL -> FAULT and asserts only `priority` and `ack_required`, never `event_type`.

### Suggested direction

- return `CHANGE_OF_RELIABILITY` from `EventStateChange::event_type()` whenever either the from- or to-state is FAULT, before the HIGH_LIMIT/LOW_LIMIT test;
[...]

Issue #204, opened 2026-07-27
IntrinsicReportingEngine is a third, uncalled Event_Enable path with the same detection/distribution defect


jscott3201:
## Summary

`bacnet_server::intrinsic_reporting::IntrinsicReportingEngine` is a third,
published `Event_Enable` code path carrying the same Clause 12.12 / 13.2.5
defect that #136 fixed in the object detectors and the Event Enrollment
evaluator — and in one branch, a worse version of it.

It is `pub mod` in `crates/bacnet-server/src/lib.rs:7`, so a downstream can call
it, but it has **zero callers anywhere in this workspace** outside its own test
module. It is a parallel, unused implementation of machinery that already exists
elsewhere.

## Evidence

Two branches, `crates/bacnet-server/src/intrinsic_reporting/mod.rs`:

```rust
// :242 — evaluate_object
let event_enable = read_event_enable(obj) as u32;
if event_enable == 0 {
    return None;              // object skipped entirely — nothing evaluated, nothing tracked
}
```

```rust
// :303 — apply_time_delay
if !enabled {
    tracker.event_state = desired_state;   // state IS tracked here
    tracker.pending_state = None;
    tracker.time_delay_remaining = None;
    return None;                            // but the transition is withheld from the caller
}
```

The first is the damaging form — the same shape as the Event Enrollment bug
fixed in #136/#201. With `Event_Enable == 0` the object is skipped before any
evaluation happens, so no state is derived or stored at all. Clause 13.2.2.1.4's
transition actions are unconditional and Clause 13.2.5 scopes `Event_Enable` to
external distribution, so nothing licenses skipping detection.

The second is the structural form: state is tracked correctly, but returning
`None` collapses "no transition" and "suppressed transition", denying the caller
the indication Clause 13.2.2.1.4's fourth action requires be delivered to *both*
the alarm-acknowledgment and distribution processes.

## Why it was out of scope for #136

PR #201 scoped itself to the two reachable families — the object detectors in
`bacnet-objects/src/event.rs` and the Event Enrollment evaluator. This engine is
a separate subsystem with its own tracker model and its own test suite; carrying
it would have materially expanded an already-large change, and it is unreachable
in-workspace so it cannot cause an observable defect here today.

That reasoning does not extend to *leaving it unrecorded*, hence this issue.

## Decide the disposition first

Before fixing, settle whether this engine should exist at all. Two paths:

[...]

PR #221, opened 2026-07-27
fix(objects,server): route transition-bit selection through one classifier (#202)

### LINKED ISSUE #202: EventState-to-transition-bit mapping defaults to TO_FAULT where Clause 13.2 requires TO_OFFNORMAL, duplicated across 4 sites
## Summary

The `EventState` → `Event_Enable` transition-bit mapping is duplicated in four
places, uses three different partitions, and every one of them ends in a
catch-all that is **inverted relative to Clause 13.2**. The standard makes
TO_OFFNORMAL the default and FAULT the enumerated case; the code does the
opposite.

## Spec basis

ASHRAE 135-2020 Clause 13.2 "Event Reporting" (p638) is explicit:

> Event-initiating objects identify their "event state" from moment to moment as
> one of any number of possibly unique event states. Notifications are triggered
> by the "transition" of conditions for an object, usually from one unique event
> state to another. In these contexts, **all states that are not normal and not
> fault are offnormal states, and transitions that result in an offnormal state
> are considered to be TO_OFFNORMAL transitions. Transitions to any fault state
> are considered to be TO_FAULT transitions. All other transitions are, by
> definition, TO_NORMAL transitions.**

So the correct partition is:

| target state | transition bit |
|---|---|
| NORMAL | TO_NORMAL |
| any fault state | TO_FAULT |
| **everything else** | **TO_OFFNORMAL** |

The residual case belongs to TO_OFFNORMAL. Note the clause says "any number of
possibly unique event states" — it explicitly anticipates states beyond the
enumerated ones, which is exactly why the default arm matters.

## Evidence

Four sites, three partitions, all defaulting to TO_FAULT:

```rust
// bacnet-objects/src/event.rs — OutOfRangeDetector::fire
s if s == EventState::NORMAL                              => TO_NORMAL,
s if s == EventState::HIGH_LIMIT || s == EventState::LOW_LIMIT => TO_OFFNORMAL,
_                                                          => TO_FAULT,   // also catches OFFNORMAL, LIFE_SAFETY_ALARM

// bacnet-objects/src/event.rs — ChangeOfStateDetector::fire
s if s == EventState::NORMAL    => TO_NORMAL,
s if s == EventState::OFFNORMAL => TO_OFFNORMAL,
_                                => TO_FAULT,   // also catches HIGH_LIMIT, LOW_LIMIT, LIFE_SAFETY_ALARM

// bacnet-objects/src/event.rs — CommandFailureDetector::fire
_ => false,                                     // see #200

// bacnet-server/src/event_enrollment/mod.rs — evaluate_event_enrollments
s if s == EventState::NORMAL => 0x04,
s if s == HIGH_LIMIT || LOW_LIMIT || OFFNORMAL => 0x01,
_                             => 0x02,          // also catches LIFE_SAFETY_ALARM
```

`EventState` has six values (`bacnet-types/src/enums/object_level.rs`):
NORMAL(0), FAULT(1), OFFNORMAL(2), HIGH_LIMIT(3), LOW_LIMIT(4),
LIFE_SAFETY_ALARM(5). Under Clause 13.2, states 2/3/4/5 are all offnormal.
Every site above routes at least one of them to TO_FAULT.

## Reachability: latent today

No current code path produces `LIFE_SAFETY_ALARM` (grep finds it only in the
enum definition and an enum round-trip test — the CHANGE_OF_LIFE_SAFETY
algorithm is not implemented), and each detector's enumerated arms happen to
cover the states its own `compute_new_state` can return. The enrollment
evaluator's `EventState::from_raw` is applied to `current_state` only, not to
the `new_state` that selects the bit, so a malformed stored `Event_State`
cannot reach the bad arm either.

So this is latent — but it is latent in the fragile direction. It activates on
any of:
- **#167**, which makes FAULT reachable through the detectors
- implementing CHANGE_OF_LIFE_SAFETY, which makes LIFE_SAFETY_ALARM reachable
- any new event state added by a future addendum

At that point a life-safety alarm would be gated by the operator's TO_FAULT bit
instead of TO_OFFNORMAL, and would be reported under the wrong transition
coordinate for `Priority` / `Ack_Required` lookup in the Notification Class
(§13.2.1) as well.

## Proposed fix

Replace all four with one shared, spec-shaped helper, so there is one place for
the partition to be right:

```rust
/// Clause 13.2: transitions to a fault state are TO_FAULT, to NORMAL are
/// TO_NORMAL, and every other state is offnormal and therefore TO_OFFNORMAL.
pub(crate) fn transition_bit(state: EventState) -> EventTransitionBit {
    match state {
        EventState::NORMAL => EventTransitionBit::ToNormal,
        EventState::FAULT  => EventTransitionBit::ToFault,
        _                  => EventTransitionBit::ToOffnormal,
    }
}
```

Unlike #200, this **is** testable today without making any new state reachable:
the helper can be unit-tested exhaustively over all six `EventState` values
directly. That is the argument for extracting it rather than patching four
`match` arms in place.

This also subsumes #200 — `CommandFailureDetector` would pick up the correct
mapping for free rather than needing its own one-line fix.

Behavior change today: none. Every state currently reachable at every site is
already enumerated correctly by that site's explicit arms; only the unreachable
residual is being corrected. That is precisely why it should be fixed now, while
it is free, rather than after #167 makes it a live wire.

Found while implementing #136 (Event_Enable gates distribution, not detection),
which rewrote all four of these expressions.

jscott3201:
## Why

Four places independently mapped an `Event_State` to one of the three `Event_Enable` transition bits, using **three different partitions**, and every one ended in a catch-all sending the residual to **TO_FAULT**. ASHRAE 135-2020 Clause 13.2 says the opposite:

> "In these contexts, all states that are not normal and not fault are offnormal states, and transitions that result in an offnormal state are considered to be TO_OFFNORMAL transitions. Transitions to any fault state are considered to be TO_FAULT transitions. All other transitions are, by definition, TO_NORMAL transitions."

Clause 13.2.2.1.2 states it outright in a parenthetical:

> "(Note that the OffNormal state includes all event states other than NORMAL and FAULT)."

| Site | Misrouted to TO_FAULT |
|---|---|
| `OutOfRangeDetector::fire` | OFFNORMAL, LIFE_SAFETY_ALARM |
| `ChangeOfStateDetector::fire` | HIGH_LIMIT, LOW_LIMIT, LIFE_SAFETY_ALARM |
| `CommandFailureDetector::fire` | HIGH_LIMIT, LOW_LIMIT, LIFE_SAFETY_ALARM |
| `event_enrollment` evaluator | LIFE_SAFETY_ALARM |

Closes #202.

## This defect is latent, not live — stated up front

No detector currently *produces* any of the misrouted states, so the catch-all only ever sees `FAULT` and is accidentally correct. **Nothing observable changes today.** This is a preventive fix and a deduplication, not a fix for anything you could trigger on a running device.

It is worth doing anyway, and worth doing as a *shared classifier* rather than four corrected copies, because correcting them in place leaves four things to get wrong again the moment a detector's state vocabulary widens. Clause 13.2's own wording anticipates exactly that — "any number of possibly unique event states" — which is why the residual arm matters.

Same shape as #200, which sat harmless until #167 made FAULT reachable and then became live immediately.

## What

The correct partition already existed in `EventStateChange::transition()`; it just wasn't what the four sites used. It now delegates to a new classifier, so there is exactly one definition:

```rust
pub fn for_target_state(state: EventState) -> Self {
    if state == EventState::NORMAL      { Self::ToNormal }
    else if state == EventState::FAULT  { Self::ToFault }
    else                                { Self::ToOffnormal }
}
```

Each of the four sites becomes:

```rust
let transition_bit = EventTransition::for_target_state(new_state).bit_mask();
let distribute = self.event_enable & transition_bit != 0;
```

Removed: three identical `TO_OFFNORMAL`/`TO_FAULT`/`TO_NORMAL` constant sets (**nine constants**) and the bare `0x01`/`0x02`/`0x04` literals in the Event Enrollment evaluator — `bit_mask()` already yields exactly those values.

**Not touched:** `EventStateChange::event_type()`. The transition *bit* depends on the destination state alone; Clause 13.2.5.3's Event Type rule applies `CHANGE_OF_RELIABILITY` to transitions *to or from* FAULT. Those are two different rules about the same transition, and merging them would silently undo #168.

## Spec basis

A normative pass ran **before** implementation, and an independent second reader re-extracted the PDF and cross-checked every quotation. Five questions, all CLEAR:

- the residual is TO_OFFNORMAL (verbatim above);
- testing exactly `EventState::FAULT` is correct for this six-valued enum, despite the standard's generic "any fault state";
- `LIFE_SAFETY_ALARM` is TO_OFFNORMAL;
- bit order is TO_OFFNORMAL, TO_FAULT, TO_NORMAL per Clause 12.2 — matching `bit_mask()`;
- the transition bit is destination-based, and must not be conflated with the Event Type rule.

Bit-string *encoding* (MSB-first vs the codebase's left-shift convention) was explicitly fenced out of scope; that is tracked as #203.

## Verification

**2290 passed / 0 failed**, 34 suites. fmt, clippy and the 700-LOC cap clean.

Five tests: an exhaustive six-state classifier table (state → transition → bit), plus per-direction selection at both polarities on two detectors.

**Those last four exist because mutation testing found the call sites unpinned.** The classifier test alone did not catch a site bypassing the classifier entirely:

| Mutant | Before fix pass | After |
|---|---|---|
| Classifier residual reverted to `ToFault` | killed | killed by 4 |
| **All three detector sites hardcode TO_FAULT** | **survived** | killed by 4 |
| **One detector site hardcodes TO_FAULT** | **survived** | killed by 2 |
| Classifier mismaps NORMAL | killed | killed by 6 |
| Enrollment site hardcodes TO_FAULT | killed | killed |

The cause of the two survivors: every pre-existing detector test used `event_enable = 0x07`, and with all three bits set `mask & <any single bit>` is true — so selecting the wrong bit was indistinguishable from selecting the right one. The new tests use selective masks. The enrollment site was already pinned only by luck, because one `bacnet-server` test happened to use a selective mask.


jscott3201:
## Adversarial review: FIX-THEN-MERGE — one confirmed finding

Three independent lenses (behavior equivalence, test integrity, API blast radius) attacked the change; an adjudicator re-verified each finding in an isolated scratch copy. **No spec-accuracy problem was found with the classifier, and no behavioral change was found in any reachable path.**

### Confirmed — fixing now

**`CommandFailureDetector`'s call site is not pinned.** Replacing its `transition_bit` line with a hardcoded `0x02` leaves `bacnet-objects` at 834 passed and `bacnet-server` at 252 passed — byte-identical to the unmutated baseline. The same mutant at the `OutOfRangeDetector` site kills two tests.

The cause is the same fixture problem that motivated the other selective-mask tests: every `CommandFailureDetector` test uses `event_enable: 0x07`, under which `mask & <any single bit>` is true. The one exception uses `0x02`/`0x05` against a **FAULT** destination — which a hardcoded `0x02` satisfies exactly, so it cannot discriminate either.

Severity is **low**, and the adjudicator was right to downgrade the reporter's "live behavioural break". No object type embeds `CommandFailureDetector` — `impl_intrinsic_reporting!` is wired only to detectors reaching objects, and `bacnet-server` has its own separate `evaluate_command_failure` — so the site is reachable only by downstream consumers of the published crate. The shipped code is correct; the regression pin was missing. That gap is on my brief, which made this detector's coverage optional.

### Rejected, with verified reasoning

Three findings pointed at `intrinsic_reporting/mod.rs` holding a fifth partition. Real code, but **already disclosed one entry away in the same CHANGELOG section** (line 80, from the #136 entry) as an uncalled duplicate deliberately excluded and tracked as **#204**. The adjudicator re-verified both premises: no in-workspace caller exists, and no `FAULT` desired-state can reach `apply_time_delay`, since the module's evaluators yield only NORMAL/OFFNORMAL/HIGH_LIMIT/LOW_LIMIT and the equality guard filters the residual.

One pointed at `acknowledge_alarm.rs:14` hand-rolling the partition. Accurate quote, but that partition is **already the spec-correct one** — semantically identical to `for_target_state(..).bit_mask()`. It is the shape #202 converts the other sites *to*, not an instance of the defect. The adjudicator even checked the exotic case: an `EventState::from_raw` outside `0..=5` falls to `_` and yields `0x01`, exactly what the classifier returns. A reasonable consolidation follow-up, not a defect here.

### Checks that came back clean

The four-site count is correct. `transition()`'s old inline body was exactly equivalent to the new classifier, so no caller of it — including notification priority and ack-required indexing — changed. No reachable state's bit selection changed at any site.

Issue #223, opened 2026-07-27
Event Enrollment carries its Event_Type onto FAULT notifications, but Clause 12.12 forbids CHANGE_OF_RELIABILITY as an Event_Type value


jscott3201:
### Affected revision

Observed on `dev` at `10d20d8`. Latent today; becomes live when #181 and #127 land.

### Behavior

`evaluate_event_enrollments` reads the enrollment's `Event_Type` property and carries it onto the transition unchanged:

- `crates/bacnet-server/src/event_enrollment/mod.rs:552` reads `PropertyIdentifier::EVENT_TYPE`
- `:589` converts it with `EventType::from_raw(event_type_raw)`
- that value lands in `EventEnrollmentTransition.event_type` (`:26`) and is never replaced for a transition to or from FAULT

There is no `CHANGE_OF_RELIABILITY` override on that path. A `grep` for `EventState::FAULT` in the file returns nothing.

### Why this is a problem

ASHRAE 135-2020 Clause 13.2.5.3 is unconditional:

> "For all transitions to, or from, the FAULT state, the corresponding event notification shall use the Event Type CHANGE_OF_RELIABILITY."

But Clause 12.12 explicitly excludes that value from the property the enrollment path reads:

> "This read-only property, of type BACnetEventType, indicates the type of event algorithm that is to be used to detect the occurrence of events and the event value notification parameters conveyed in event notifications. The value of this property is an enumerated type that may have any standard BACnetEventType value **except CHANGE_OF_RELIABILITY**."

So the correct value for a FAULT transition can never come from `Event_Type` by construction — the override has to be applied by the code that builds the notification. The intrinsic-reporting path now does exactly this via `EventStateChange::event_type(algorithm)`, which applies the FAULT rule and otherwise returns the algorithm (#210). The enrollment path has no equivalent.

### Why it is latent

The enrollment evaluator cannot currently produce `EventState::FAULT` at all — which is **#181** ("Event Enrollment Reliability is permanently NO_FAULT_DETECTED: no MONITORED_OBJECT_FAULT propagation, no configuration-error indication, no FAULT state"). And enrollment transitions are not yet propagated into the notification pipeline, which is **#127**.

[...]

Issue #222, opened 2026-07-27
Binary Output and Multi-state Output apply CHANGE_OF_STATE intrinsic reporting, but their object clauses require COMMAND_FAILURE


jscott3201:
### Affected revision

Observed on `dev` at `10d20d8`.

### Behavior

ASHRAE 135-2020 Clause 13.2.2 makes the intrinsic-reporting algorithm a property of the object type, not a configuration choice:

> "For intrinsic reporting in standard object types, the event algorithm is implied by the object type. For algorithmic reporting, the Event Enrollment object contains the Event_Type property which indicates the event algorithm. An object, standard or proprietary, shall use only a single event algorithm."

All nine Clause 12 object-type descriptions were read and quoted. Seven match this implementation; **two do not**:

| Object type | Clause | Required algorithm | Detector wired here | |
|---|---|---|---|---|
| Analog Input | 12.2 | OUT_OF_RANGE | `OutOfRangeDetector` | ✔ |
| Analog Output | 12.3 | OUT_OF_RANGE | `OutOfRangeDetector` | ✔ |
| Analog Value | 12.4 | OUT_OF_RANGE | `OutOfRangeDetector` | ✔ |
| Binary Input | 12.6 | CHANGE_OF_STATE | `ChangeOfStateDetector` | ✔ |
| **Binary Output** | **12.7** | **COMMAND_FAILURE** | `ChangeOfStateDetector` | ✘ |
| Binary Value | 12.8 | CHANGE_OF_STATE | `ChangeOfStateDetector` | ✔ |
| Multi-state Input | 12.18 | CHANGE_OF_STATE | `ChangeOfStateDetector` | ✔ |
| **Multi-state Output** | **12.19** | **COMMAND_FAILURE** | `ChangeOfStateDetector` | ✘ |
| Multi-state Value | 12.20 | CHANGE_OF_STATE | `ChangeOfStateDetector` | ✔ |

Verbatim, Clause 12.7:

> "Binary Output objects that support intrinsic reporting shall apply the COMMAND_FAILURE event algorithm."

and Clause 12.19:

> "Multi-state Output objects that support intrinsic reporting shall apply the COMMAND_FAILURE event algorithm."

[...]

Issue #181, opened 2026-07-27
Event Enrollment Reliability is permanently NO_FAULT_DETECTED: no MONITORED_OBJECT_FAULT propagation, no configuration-error indication, no FAULT state


jscott3201:
### Affected revision

Observed on `dev` at commit `3990f3e`.

### Behavior

`EventEnrollmentObject` declares `reliability: u32` (`crates/bacnet-objects/src/event_enrollment/mod.rs:35`), initializes it to 0 (`:62`), and never mutates it: there is no setter, no `write_property` branch, and no internal mutation path. It is served by the shared macro (`crates/bacnet-objects/src/common.rs:132-134`) and can only ever return NO_FAULT_DETECTED (`crates/bacnet-types/src/enums/object_level.rs:37`).

`evaluate_event_enrollments` (`crates/bacnet-server/src/event_enrollment/mod.rs:494-673`) never reads or writes any `Reliability` and never produces `EventState::FAULT`. Every unresolvable configuration silently continues instead: undecodable `Event_Parameters` (`:531-539`), missing `Object_Property_Reference` (`:541-543`), missing monitored object (`:545-547`), unreadable monitored property (`:548-551`), and an unresolvable FLOATING_LIMIT setpoint reference (`:573-575`, which includes every remote reference — `read_setpoint` returns `None` whenever `device_identifier.is_some()`, `:388-391`). `Fault_Parameters` is stored and round-tripped on the object (`:164-167`, `:225-234`), but no fault algorithm is ever evaluated.

Clause 12.12 (p234) prescribes both the value and the order in which it is derived: "The reliability-evaluation in an Event Enrollment object is processed in the following order: (1) Evaluate the reliability of the Event Enrollment object itself. (2) If the reliability is NO_FAULT_DETECTED, then if the reliability of the monitored object is not NO_FAULT_DETECTED, then the reliability shall be MONITORED_OBJECT_FAULT. (3) If the reliability is NO_FAULT_DETECTED, then evaluate the fault algorithm, if one is in use, to determine the reliability."

Clause 13.2.2 (p640) states the downstream consequence: "In the event-state-detection process, the event algorithm and the Reliability property together determine the event state of an object. ... when Reliability has a value other than NO_FAULT_DETECTED, the event-state-detection process will determine the object's event state to be FAULT."

[...]

Issue #131, opened 2026-07-14
Restrict GetEnrollmentSummary and report actual event type and priority


jscott3201:
### Affected revision

Observed on rusty-bacnet v0.10.1 at commit `8b8c0cba3e92caa30e145c91abedfa003c64f2c9`.

### Behavior

The GetEnrollmentSummary handler scans every object in the database, reads Notification Class priority index one regardless of transition, and returns entries with hardcoded `CHANGE_OF_STATE` and priority zero.

The response can therefore misclassify object scope, event type, and priority.

### Suggested direction

- define and enforce the candidate object scope;
- derive event type from the enrollment;
- select Notification Class priority by transition coordinate;
- cover each request filter independently.

### Relevant code

- `crates/bacnet-server/src/handlers/alarm_event.rs`
- `crates/bacnet-objects/src/event_enrollment/mod.rs`
- `crates/bacnet-objects/src/notification_class/mod.rs`

PR #224, opened 2026-07-27
fix(objects,server): take Event Type from the event algorithm (#210)

### LINKED ISSUE #210: Event Type is inferred from the state transition instead of the object's configured event algorithm
## Summary

`EventStateChange::event_type()` infers the notification's Event Type from *which
event states the transition involves*, rather than from the event-initiating
object's configured event algorithm. The standard requires the latter.

The heuristic gives the right answer for every object type wired today, purely by
coincidence, so this is latent — but it is wrong by construction and becomes
wire-visible as soon as any of three pending changes lands.

## Spec basis

ASHRAE 135-2020, the 'Event Type' parameter of ConfirmedEventNotification (Clause 13.8, spec line 44593) and UnconfirmedEventNotification (Clause 13.9, spec line 44746):

> If 'Notify Type' is ALARM or EVENT when the 'To State' parameter is FAULT, then
> this parameter shall have a value of CHANGE_OF_RELIABILITY. The Event Type
> CHANGE_OF_RELIABILITY shall be used for reporting a transition from FAULT.
> **Otherwise, this parameter shall have the value associated with the
> event-initiating object's configured event algorithm.**

Table 13-3 says the same: "Otherwise the value associated with the
event-initiating object's event algorithm."

The FAULT half of that sentence is #168. This issue is the "otherwise" half.

## Evidence

```rust
// crates/bacnet-objects/src/event.rs
pub fn event_type(&self) -> EventType {
    if self.from == EventState::HIGH_LIMIT || self.from == EventState::LOW_LIMIT
        || self.to == EventState::HIGH_LIMIT || self.to == EventState::LOW_LIMIT
    {
        EventType::OUT_OF_RANGE
    } else {
        EventType::CHANGE_OF_STATE
    }
}
```

It only ever returns one of two values, out of the ~20 in `BACnetEventType`, and
it derives them from `EventState` — a different axis entirely. The single
production caller puts the result straight on the wire
(`crates/bacnet-server/src/server/event_notifications.rs:153`).

## Why it is latent today

Each wired detector's state vocabulary happens to map 1:1 onto its algorithm:

| Object types | Detector | States produced | Heuristic yields | Correct? |
|---|---|---|---|---|
| AI / AO / AV | `OutOfRangeDetector` | HIGH_LIMIT, LOW_LIMIT | OUT_OF_RANGE | yes |
| BI / BO / BV, MSI / MSO / MSV | `ChangeOfStateDetector` | OFFNORMAL | CHANGE_OF_STATE | yes |

`CommandFailureDetector` exists but is wired to no object (grep finds it only in
`crates/bacnet-objects/src/event/tests.rs`), so its OFFNORMAL never reaches the
wire.

## When it becomes live

1. **Wiring `CommandFailureDetector`** to the commandable outputs it is written
   for (BO, MSO) — its OFFNORMAL would be reported as CHANGE_OF_STATE instead of
   COMMAND_FAILURE.
2. **Implementing FLOATING_LIMIT intrinsic reporting** — its HIGH_LIMIT/LOW_LIMIT
   would be reported as OUT_OF_RANGE instead of FLOATING_LIMIT.
3. **#127**, propagating Event Enrollment transitions into the notification
   pipeline. This is the sharpest case: `EventEnrollmentTransition` *already
   carries the correct value* in its `event_type` field, read from the
   enrollment's `Event_Type` property — and the notification builder would throw
   it away and re-derive from states. A CHANGE_OF_BITSTRING or CHANGE_OF_VALUE
   enrollment would be flattened to CHANGE_OF_STATE.

Once #135 populates `event_values`, a wrong Event Type also makes the receiver
decode context tag [12] against the wrong `BACnetNotificationParameters`
alternative — the same downstream consequence #168 describes.

## Proposed fix

Stop deriving the Event Type from the transition. Carry it from the source:

- intrinsic reporting: each detector knows its own algorithm, so the detector (or
  `TransitionOutcome`) should report it;
- Event Enrollment: use the `event_type` already present on
  `EventEnrollmentTransition`.

`EventStateChange::event_type()` then has no correct implementation left and
should be deleted rather than fixed — the transition alone does not carry enough
information to answer the question. Note it would still be needed for the FAULT
case (#168), which *is* a function of the transition; that suggests the final
shape is "CHANGE_OF_RELIABILITY if either end is FAULT, else the algorithm passed
in".

Sequence after or with #168, which fixes the FAULT half of the same sentence.

Found while implementing #168.

---

**Correction (2026-07-27):** this issue originally attributed the quoted 'Event Type' parameter text to Clause 13.5. That is wrong — Clause 13.5 is the AcknowledgeAlarm Service and has no Event Type parameter. The text is in the ConfirmedEventNotification (13.8) and UnconfirmedEventNotification (13.9) service descriptions, and in Table 13-3. The substance of the issue is unchanged; only the clause number was wrong. Caught by an independent Codex spec-accuracy review on PR #211.

jscott3201:
## Why

`EventStateChange::event_type()` derived the notification's Event Type from **which event states the transition involved** — returning `OUT_OF_RANGE` when either end was `HIGH_LIMIT`/`LOW_LIMIT` and `CHANGE_OF_STATE` otherwise. That is two of roughly twenty `BACnetEventType` values, chosen from `EventState`, a different axis entirely. The single production caller puts the result straight on the wire.

ASHRAE 135-2020 Clauses 13.8.1.1 and 13.9.1.1:

> "Otherwise, this parameter shall have the value associated with the event-initiating object's configured event algorithm."

Clause 13.2.2 says where that algorithm comes from — *"for intrinsic reporting in standard object types, the event algorithm is implied by the object type"* — and Clause 13.3 that *"the event algorithms are indicated by the BACnetEventType value of the same name."*

Closes #210.

## What

`event_type()` now takes the algorithm as a **required parameter** and applies only the FAULT rule:

```rust
pub fn event_type(&self, algorithm: EventType) -> EventType {
    if self.from == EventState::FAULT || self.to == EventState::FAULT {
        EventType::CHANGE_OF_RELIABILITY
    } else {
        algorithm
    }
}
```

Each detector carries `pub const ALGORITHM` — `OutOfRangeDetector` → `OUT_OF_RANGE`, `ChangeOfStateDetector` → `CHANGE_OF_STATE`, `CommandFailureDetector` → `COMMAND_FAILURE` — and `fire()` computes the final value where both the algorithm and the transition are known, carrying it out on `TransitionOutcome`. The server uses the carried value instead of recomputing.

A required parameter rather than an inferred default is deliberate: a detector added later **cannot** silently fall back to a guess.

**Breaking:** `TransitionOutcome` gains `pub event_type: EventType`, and `event_type()` gains a parameter. Not `#[non_exhaustive]`, so exhaustive struct literals and patterns must account for the field.

## A correction to the issue's own quotation

#210 quotes the Event Type definition accurately but **incompletely**, and the omission matters. The sentence it quotes opens *"If 'Notify Type' is ALARM or EVENT..."*, which reads as though the FAULT rule were conditional — implying our unconditional override might be wrong for ACK_NOTIFICATION.

It isn't. The definition has a **third paragraph** the issue omits:

> "If 'Notify Type' is ACK_NOTIFICATION when 'To State' is FAULT, the Event Type shall be CHANGE_OF_RELIABILITY. When 'To State' is NORMAL, and the device can determine reporting acknowledgement of a transition from FAULT, the Event Type shall be CHANGE_OF_RELIABILITY. Otherwise, the Event Type shall be the value associated with the event-initiating object's configured event algorithm."

So the ACK case is handled explicitly rather than falling through, and Clause 13.2.5.3 states the underlying rule symmetrically for transitions "to, or from" FAULT. The unconditional override is correct and is retained. **No defect in the merged #168 work** — which was the live question when this started.

## Why the old code produced right answers

Purely by coincidence — each wired detector's state vocabulary happened to map onto its algorithm:

| Detector | Wired to | States | Old guess | Correct? |
|---|---|---|---|---|
| `OutOfRangeDetector` | AI/AO/AV | HIGH_LIMIT, LOW_LIMIT | OUT_OF_RANGE | by luck |
| `ChangeOfStateDetector` | BI/BV, MSI/MSV | OFFNORMAL | CHANGE_OF_STATE | by luck |
| `CommandFailureDetector` | *nothing* | OFFNORMAL | CHANGE_OF_STATE | **wrong** |

That last row is now an explicit regression test — `command_failure_offnormal_reports_command_failure_not_change_of_state`.

## Two related defects found, filed rather than folded in

Confirming this against the standard turned up two things outside this change's scope:

- **#222** — Binary Output and Multi-state Output apply the CHANGE_OF_STATE algorithm, but Clauses 12.7 and 12.19 require **COMMAND_FAILURE**. All nine object-type clauses were quoted; exactly these two are wrong. After this PR they report `CHANGE_OF_STATE` — truthful about the algorithm actually in use, still non-conformant until rewired. Reporting `COMMAND_FAILURE` from an object running a CHANGE_OF_STATE detector would be worse: a lie on the wire rather than a consistent wrong answer.
- **#223** — the Event Enrollment path carries its `Event_Type` onto FAULT transitions with no override, though Clause 12.12 forbids `CHANGE_OF_RELIABILITY` as an `Event_Type` value. Latent behind #181 and #127.

## Verification

**2296 passed / 0 failed**, 34 suites. fmt, clippy, 700-LOC cap clean, and `cargo check -p rusty-bacnet --tests` passes (that crate is excluded from the normal test run, and a new `TransitionOutcome` field can break struct literals there).

Mutants verified dead:

| Mutant | Killed by |
|---|---|
| `CommandFailureDetector::ALGORITHM` = CHANGE_OF_STATE | 1 |
| `OutOfRangeDetector::ALGORITHM` = CHANGE_OF_STATE | 3 |
| `ChangeOfStateDetector::ALGORITHM` = OUT_OF_RANGE | 2 |
| FAULT branch removed entirely | 5 |
| FAULT branch narrowed to the `to` direction only | 4 |

The last two are mine beyond what the brief asked for — the second confirms the from-FAULT direction is genuinely pinned, not just the to-FAULT one.


jscott3201:
## Adversarial review: FIX-THEN-MERGE

Three lenses (wire correctness, test integrity, API blast radius) attacked the change; an adjudicator re-verified every finding in an isolated `git archive` scratch copy.

**No behavioral defect.** All three `TransitionOutcome` constructions and both server propagation sites carry the detector's value verbatim, no production `event_type()` caller was missed, and `cargo check -p rusty-bacnet --tests` is clean.

### Confirmed — fixing now

**The detector-to-wire seam has zero test coverage.** Three mutations were run against the full workspace suite and *all three survived*:

| Mutation | Result |
|---|---|
| Production `From<(EventStateChange, EventType)>` discards the passed value | **survived** |
| `lifecycle.rs:521` pushes a hardcoded Event Type | **survived** |
| `event_notifications.rs:176` encodes a hardcoded Event Type | **survived** |

The cause is the `#[cfg(test)]` `From<EventStateChange>` shim. All seven `build_and_send_event_notification` calls pass a bare `EventStateChange`, so they resolve through the shim — which recomputes the value locally from a hardcoded `OUT_OF_RANGE`. **The production `From` impl is entered by no test at all.** The only two wire-level `event_type` assertions are FAULT transitions expecting `CHANGE_OF_RELIABILITY`, which any constant satisfies.

There is also a coverage *regression* in a diff that touches no test code: before this PR, `:153` called `change.event_type()` directly, so those two FAULT tests exercised the production derivation. `git diff` on the test file is **0 lines**, yet they now exercise cfg(test) code instead.

Severity is medium, not high — production is correct today; what is lost is the guard. It matters concretely for **#222**: when Binary Output and Multi-state Output are rewired to `CommandFailureDetector`, a regression at either propagation site would ship silently.

Fix in progress: delete the shim, pass explicit `(change, event_type)` tuples at all seven sites, and add a **non-FAULT** wire assertion through the production path. A FAULT transition cannot serve as that guard, since `CHANGE_OF_RELIABILITY` is what a hardcoded constant would produce anyway.

**CHANGELOG accuracy** — the entry introducing `TransitionOutcome` still described it as `{ change, distribute }`. Both entries sit in the same unreleased block, so there is no release in which that shape was ever true. Corrected.

### Rejected, with verified reasoning

- **`non_fault_transitions_are_unchanged` is a tautology** — *refuted by direct mutation.* The adjudicator rewrote `event_type()` to ignore its parameter and return a constant; that test is among the 5 that fail. It has real mutation-killing power against the most likely regression. What survives is that its four rows are redundant with each other — test-data style, out of scope.
- **`GetEnrollmentSummary` hardcodes CHANGE_OF_STATE and ignores `event_type_filter`** — factually accurate and a genuine conformance bug, but the file is untouched by this diff and the inconsistency is byte-for-byte identical before and after. Already tracked as **#131**, which names both the hardcoded value and the per-filter coverage; no duplicate filed.

Issue #241, opened 2026-07-28
bacnet_types::enums::Reliability is missing multi-state-out-of-range (25)


jscott3201:
### Affected revision

`dev` at `20b8a11`. Found during the #218 spec review and independently confirmed by an adversarial reviewer reading the ASN.1 production.

### Behavior

`crates/bacnet-types/src/enums/object_level.rs:35` defines `Reliability` with 24 constants, ending at `REFERENCED_OBJECT_FAULT = 24`. The standard's `BACnetReliability ::= ENUMERATED` production defines one more:

- values 0 through 10;
- value 11 explicitly *"reserved for a future addendum"*;
- values 12 through 24;
- **`multi-state-out-of-range (25)`** — missing here;
- then *"Enumerated values 0-63 are reserved for definition by ASHRAE. Enumerated values 64-65535 may be used by others."*

### Impact today is small but the gap is real

#239 validates written `Reliability` values against the range `0..=10 | 12..=25 | 64..=65535`, so a client **can** write 25 and it is accepted and stored correctly — the validation range was written from the spec text rather than from the enum, precisely because the enum is short. So there is no rejection bug.

What is missing is the named constant. Anything that needs to *produce* or *match on* `MULTI_STATE_OUT_OF_RANGE` has no symbol for it and must use a bare `25`.

That matters for **#226**, which covers `CONFIGURATION_ERROR` / `MULTI_STATE_OUT_OF_RANGE` on the multi-state types — implementing it will want this constant. These two issues should probably be done together, or this one folded into #226 if that is the natural home.

### Suggested direction

- Add `const MULTI_STATE_OUT_OF_RANGE = 25;` to the `Reliability` enum.
- Consider deriving #239's validation range from the enum rather than restating it as a literal, so the two cannot drift again. An adversarial reviewer flagged the literal duplication separately; a single shared predicate would close both.
- Check the other enums in `crates/bacnet-types/src/enums/` for the same kind of truncation against the standard's productions — this one was found incidentally, not by a sweep.

### Relevant code

- `crates/bacnet-types/src/enums/object_level.rs:35`
- the `Reliability` validation range introduced by #239

### Related

#226, #218, #239

Issue #192, opened 2026-07-27
Device object Protocol_Services_Supported is a stale hardcoded constant: four bits are over-declared, twenty executed services are undeclared, and three cannot be represented


jscott3201:
### Affected revision

Observed on `dev` at commit `3990f3e`.

### Behavior

`crates/bacnet-objects/src/device/mod.rs:304` is `let protocol_services_supported = vec![0xA4, 0x0B, 0x80, 0x35, 0x80, 0x00];`, a hardcoded constant with no linkage to the dispatch table, returned verbatim at `:414-419` as `BitString { unused_bits: 7, data }`. Decoded MSB-first it sets bits {0, 2, 5, 12, 14, 15, 16, 26, 27, 29, 31, 32}.

The comment at `:291-303` states the intent as "29=UnconfirmedCOVNotification, 31=WhoHas, 32=WhoIs", but Clause 21 BACnetServicesSupported (p948-949) assigns 28 = unconfirmed-cov-notification, 29 = unconfirmed-event-notification, 31 = unconfirmed-text-message, 32 = time-synchronization, 33 = who-has, 34 = who-is. Three bits are mis-assigned, and Who-Is and Who-Has — the two services this device most needs to advertise — are undeclared.

Clause 12.11 (p226): the property "indicates which standardized protocol services are executed by this device's protocol implementation." Enumerating the actual dispatch set from `crates/bacnet-server/src/server/requests.rs` gives 28 executed services (22 confirmed arms plus WhoIs, WhoHas, TimeSynchronization, UTCTimeSynchronization, WriteGroup and UnconfirmedTextMessage). Against that:

- executed but undeclared: bits 3, 4, 6, 7, 8, 9, 10, 11, 17, 19, 20, 33, 34, 35, 36, 37, 38, 39, 40, 41 — including GetAlarmSummary, GetEnrollmentSummary, AtomicReadFile/AtomicWriteFile, AddListElement/RemoveListElement, CreateObject/DeleteObject, DeviceCommunicationControl, ReinitializeDevice, Who-Has, Who-Is, ReadRange, LifeSafetyOperation, SubscribeCOVProperty, GetEventInformation and WriteGroup;
[...]

PR #244, opened 2026-07-29
feat(types): resolve enumeration, bit-string

amab8901:
`decode_application_value()` returns `PropertyValue` which leaves `Enumerated` and `BitString` as wrapped raw data instead of the human-readable enum variants the value should correspond to according to the specs clause 21. 

This PR introduces the function `resolve_value()`, which transforms the `PropertyValue` into an equivalent `ResolvedValue` where everything is the same except that `Enumerated` and `BitString` are replaced with the human-readable enum variant given in Clause 21. 

Example code for how to use `resolve_value()`:

``` rust
  let p = match bc
      .read_property(&mac_address, ob_id, gi.property, None)
      .await
  {
      Ok(p) => p,
      Err(e) => {
          error!(?e, ?object_type, ?gi.property, ?gi.instance_number);
          continue;
      }
  };

  if p.property_value.is_empty() {
      continue;
  }

  let (v, _) = decode_application_value(&p.property_value, 0).unwrap();

  let r = resolve_value(p.property_identifier, v);

  info!(?p.object_identifier, ?p.property_identifier, ?r);
``` 

## Output Comparison Examples: `decode_application_value()` vs `resolve_value()`

SYSTEM_STATUS v=Enumerated(0)
SYSTEM_STATUS r=Enumerated(DeviceStatus(DeviceStatus::OPERATIONAL))

PROTOCOL_SERVICES_SUPPORTED v=BitString { unused_bits: 1, data: [68, 11, 72, 40, 170] }
PROTOCOL_SERVICES_SUPPORTED r=BitString(ServicesSupported(CONFIRMED_COV_NOTIFICATION | SUBSCRIBE_COV | READ_PROPERTY | READ_PROPERTY_MULTIPLE | WRITE_PROPERTY | DEVICE_COMMUNICATION_CONTROL | REINITIALIZE_DEVICE | I_AM | UNCONFIRMED_COV_NOTIFICATION | TIME_SYNCHRONIZATION | WHO_IS | UTC_TIME_SYNCHRONIZATION | SUBSCRIBE_COV_PROPERTY))

SEGMENTATION_SUPPORTED v=Enumerated(0)
SEGMENTATION_SUPPORTED r=Enumerated(Segmentation(Segmentation::BOTH))

UNITS v=Enumerated(64)
UNITS r=Enumerated(EngineeringUnits(EngineeringUnits::DEGREES_FAHRENHEIT))

### DIFF
--- a/crates/bacnet-types/src/bitstring.rs
+++ b/crates/bacnet-types/src/bitstring.rs
@@ -1,0 +1,339 @@
+//! Named BACnet bit-string types (ASHRAE 135-2020 Clause 21).
+//!
+//! Several properties carry a `BitString` whose bits have a fixed, named
+//! meaning defined by the standard. This module models those `BACnetXxx` BIT
+//! STRING productions so a decoded [`crate::primitives::PropertyValue::BitString`]
+//! can be promoted from raw `(unused_bits, data)` bytes to named flags, the same
+//! way [`crate::enums::ResolvedEnum`] promotes an `Enumerated`.
+//!
+//! ## Bit order
+//!
+//! BACnet bit strings are transmitted most-significant-bit-first: bit 0 of the
+//! string is the top bit (`0x80`) of the first content byte, bit 8 is the top
+//! bit of the second byte, and so on. Every type here reads bit *N* of the
+//! string as "feature N", matching how the device object packs these fields
+//! (see `bacnet-objects`'s `compute_object_types_supported`).
+//!
+//! [`crate::primitives::StatusFlags`] predates this module and keeps its own
+//! right-aligned representation; it is re-exported and mapped by the resolver as
+//! well.
+
+use crate::enums::{ObjectType, ServiceSupported};
+
+/// Read bit `n` of a BACnet bit string (MSB-first) from its content bytes.
+///
+/// `data` is the bit-string payload *without* the leading unused-bits count
+/// (i.e. `PropertyValue::BitString::data`). Bits past the end read as `false`.
+fn wire_bit(data: &[u8], n: usize) -> bool {
+    let mask = 0x80u8 >> (n % 8);
+    data.get(n / 8).is_some_and(|b| b & mask != 0)
+}
+
+/// Render a bitflags value as its set Clause-21 names (` NAME | NAME `), or `()`
+/// when none are set — for both `Display` and `Debug`, never a raw number.
+macro_rules! impl_named_bit_display {
+    ($t:ty) => {
+        impl core::fmt::Display for $t {
+            fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+                let mut first = true;
+                for (name, _) in self.iter_names() {
+                    if !first {
+                        f.write_str(" | ")?;
+                    }
+                    f.write_str(name)?;
+                    first = false;
+                }
+                if first {
+                    f.write_str("()")?;
+                }
+                Ok(())
+            }
+        }
+
+        impl core::fmt::Debug for $t {
+            fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+                core::fmt::Display::fmt(self, f)
+            }
+        }
+    };
+}
+
+/// Write `Display` items as ` A | B | C ` (or `()` when empty).
+fn write_joined<T: core::fmt::Display>(
+    f: &mut core::fmt::Formatter<'_>,
+    items: impl Iterator<Item = T>,
+) -> core::fmt::Result {
+    let mut first = true;
+    for item in items {
+        if !first {
+            f.write_str(" | ")?;
+        }
+        write!(f, "{item}")?;
+        first = false;
+    }
+    if first {
+        f.write_str("()")?;
+    }
+    Ok(())
+}
+
+bitflags::bitflags! {
+    /// `BACnetEventTransitionBits` — 3-bit string for `event-enable`,
+    /// `acked-transitions`, and `ack-required` (Clause 21).
+    #[derive(Clone, Copy, PartialEq, Eq, Hash, Default)]
+    pub struct EventTransitionBits: u8 {
+        /// Transition to an off-normal event state (bit 0).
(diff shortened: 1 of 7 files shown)


jscott3201:
Let me review this and get a CI run. I'll cross check it across the issues board as well as this may close one or two I hope!

jscott3201 (changes requested):
Thanks for this contribution — the direction is exactly right, and the review found the core of it solid: all 64 hardcoded property numbers check out against Clause 21, the `ServiceSupported` bit positions 0–48 reproduce the `BACnetServicesSupported` production exactly (including the removed 13/24/25), and the MSB-first bit math in `ServicesSupported`/`ObjectTypesSupported` was proven correct for every bit position including the byte boundaries (bit 32, bit 63/64). The tests pin real boundary cases, `no_std` stays clean, and Clause 20.2.10's "Undefined bits shall be zero" means the fixed-position decoders are safe against any conformant encoder.

Four changes are needed before merge, one of them blocking:

**1. [blocking] `resolve.rs` — car-door-status (450) is mapped to an enumeration that does not exist in the standard.**
ASHRAE 135-2020 defines no `BACnetLiftCarDoorStatus`. The Lift object's property table types Car_Door_Status as `BACnetARRAY[N] of BACnetDoorStatus`, and `BACnetDoorStatus` is `closed(0), opened(1), unknown(2), door-fault(3), unused(4), none(5), closing(6), opening(7), safety-locked(8), limited-opened(9)`. The crate's `LiftCarDoorStatus` is a different permutation of the same words (`UNKNOWN=0, NONE=1, CLOSING=2, CLOSED=3, …`), so every resolved value gets a wrong name — wire `0` (closed) prints `UNKNOWN`, wire `3` (door-fault) prints `CLOSED`. Please map 450 to `DoorStatus` instead (its 0–2 match the spec; the missing 3–9 constants degrade to raw numbers under the open-enum design, which is the intended fallback). The spec-less `LiftCarDoorStatus` enum itself is a pre-existing defect we'll track separately.

**2. `resolve.rs` — tracking-value (164) contradicts the module's own mapping rule.**
Tracking_Value is object-type-dependent: Life Safety Point/Zone type it `BACnetLifeSafetyState`, but Lighting Output types it `REAL` (Table 12-64). In practice the mapping is safe — a conformant Lighting Output sends REAL, which never reaches the `Enumerated` arm, and no object type in 135-2020 has Tracking_Value as a *different* enumeration — but the module doc currently states that object-type-dependent properties are intentionally left unmapped, and 164 violates that as written. Either drop the arm or (preferred) keep it and amend the module doc to state the carve-out explicitly: 164 is mapped because its only ENUMERATED form is `BACnetLifeSafetyState`.

**3. `resolve.rs:122` — `Segmentation::from_raw(value as u8)` wraps out-of-range values into valid names.**
This is the only narrowing cast in the map. A nonconforming `Enumerated(259)` resolves to `NO_SEGMENTATION` (259 & 0xFF = 3) instead of preserving the raw number, contradicting the module's preserve-the-number contract (which the `vendor_object_type_passes_through` test asserts for ObjectType). Guard it: `if value > u8::MAX as u32 { Self::Unknown(value) }`.

**4. `resolve.rs:260` — StatusFlags decode is the one length-fragile decoder in the PR.**
`data.first() >> unused_bits` is correct for the conformant 4-bit encoding, but it trusts the declared length where every other decoder reads fixed MSB-first positions: a device padding to 8 bits (`unused_bits=0`) silently decodes as empty, and a 3-bit string (`unused=5`) turns in-alarm into OVERRIDDEN. Decoding bits 0–3 via `wire_bit` like the other types makes it robust and consistent, since Clause 20.2.10 fixes bit 0 at the MSB of the first octet regardless of declared length.

Two doc nits, non-blocking: `bitstring.rs` says StatusFlags "is re-exported" by the module, but there is no `pub use` (it lives in `primitives`); and the `ObjectTypesSupported` doc attributes the 0..1023 range to the bit-string production — that range belongs to the `BACnetObjectType` ENUMERATED (the `BACnetObjectTypesSupported` production itself stops at bit 62 with no extension marker). Accepting 1024 bits is fine as robustness; just reword the justification.

One heads-up on sequencing, not a defect in this PR: the decoders here are spec-correct MSB-first, but this stack's own *encode* side for event-enable/acked-transitions/limit-enable is currently bit-reversed (tracked in #203). Until #203 lands, a rusty-bacnet client using `resolve_value` will misread a rusty-bacnet server for those three properties (TO_OFFNORMAL/TO_NORMAL swapped; limit-enable always empty), while reading conformant third-party devices correctly. We'll land #203 right behind this so the stack is self-consistent.

On the issue board: this closes nothing outright, but the `ServiceSupported` enum + `u64` bitset provide exactly the representation #192 says is missing (including the services beyond bit 40 that the current hardcoded constant can't express), and your `services_supported_follows_clause_21` test independently corroborates #192's claim that the device constant's bit-32 comment is mislabeled. #241's missing `Reliability` constant (25) will automatically upgrade your resolver's output when it lands — no change needed here.

jscott3201:
To keep this moving we'll push the fixes above directly to this branch (maintainer edits are enabled) rather than round-tripping — the changes are small and mechanical: retarget the 450 arm to `DoorStatus` (+ extend it to the full Clause 21 production), document the tracking-value carve-out, guard the `Segmentation` cast, and switch the StatusFlags decode to the same MSB-first `wire_bit` path as the other types. Once CI is green on the updated head and our spec gate signs off, we'll merge. Thanks again — the bit-order proofs and boundary tests in this PR made the review straightforward.

jscott3201 (approved):
All requested changes are in (388d5be): car-door-status now resolves as BACnetDoorStatus with the enum extended to the full Clause 21 production, the Segmentation cast is guarded, StatusFlags decodes by MSB-first wire position, and the tracking-value exception is documented. Spec gate re-ran on the fix commit and concurred on every item with clause citations (Clauses 12.11.15, 12.21/12.22/12.54, 20.2.10, 21). CI green on the final head. Follow-ups filed as #245/#246; the encode-side counterpart lands via #203.

PR #268, opened 2026-07-31
feat(types): impl serde::Deserialize for ObjectType & PropertyIdentifier

amab8901:
I need it so I can have JSON-based config file where I specify which properties I wanna read from which object.

Example JSON config file:

``` json
{
  "objects": [
    {
      "group_name": "Object 1",
      "mac_address": "127.0.0.1.224.225",
      "object_type": "DEVICE",
      "group": [
        {
          "description": "-",
          "property": "Object_List",
          "instance_number": 1521976,
          "unit": "#"
        }
      ]
    },
    {
      "group_name": "Object 2",
      "mac_address": "127.0.0.1.224.225",
      "object_type": "ANALOG_INPUT",
      "group": [
        {
          "description": "-",
          "property": "Present_Value",
          "instance_number": 0,
          "unit": "#"
        }
      ]
    }
  ]
}
```


jscott3201:
Let me review this overall structure and see how it fans out. Good add! I'll try to get this rolled in later today. I'll probably release a new version to PyPi and Crates.io this weekend.

jscott3201:
LGTM! I'll be pushing a new release at some point this week. I need to fix a relinquish_default issue before I do.

KnorpelSenf:
@amab8901 why don't you do this in your own code via a zero-cost wrapper type? Could've just implemented `Deserialize` there, no? What am I missing?

amab8901:
> @amab8901 why don't you do this in your own code via a zero-cost wrapper type? Could've just implemented `Deserialize` there, no? What am I missing?

ye I could. But using a wrapper is more verbose than using the type directly. I prefer writing
``` rust
  ObjectIdentifier::new(object_type, gi.instance_number);
  client.read_property(&mac_address, ob_id, gi.property, None).await;
```

rather than:
``` rust
  ObjectIdentifier::new(object_type.0, gi.instance_number);
  client.read_property(&mac_address, ob_id, gi.property.0, None).await;
```

I think both approaches work fine but the small verbosity difference makes library impl the slightly better approach.

I think also overall that the total amount of code everyone has to write will be lower if it's implemented once in the library behind feature gate instead of multiple people having to write the same impl over and over. Since it's feature-gated, it shouldn't affect those who don't want it. 

I think something I could have done better with this PR is to impl serde via `bacnet_enum!` (still via feature gate) and also put serde feature gate on the FromStr since serde is the only thing using it

jscott3201:
I'll double back on this and look at JSON support overall. Trying to get down that issues log I have in the GH issues section for the next release. Thinking this over more I want to trace the JSON stuff out to ensure it doesn't hit the hot path.

I use zero copy techniques on the wire for speed but if anything allocates it makes it a moot speedup. I'll open an issue later today to track a JSON piece in the higher level APIs.

Issue #260, opened 2026-07-30
handle_read_property_multiple forwards array indexes with no is_array_property gate


jscott3201:
### Affected revision

`dev` at `415eb32`. Found by the Event_Time_Stamps family grounding pass (2026-07-30).

### Behavior

The ReadProperty handler gates indexed reads: `crates/bacnet-server/src/handlers/read_property.rs:21-49` checks an `is_array_property` whitelist and rejects an index on a non-array property with `PROPERTY / PROPERTY_IS_NOT_AN_ARRAY` (`:43-48`).

The ReadPropertyMultiple handler has no equivalent. `handle_read_property_multiple` (`:82-183`) selects the per-reference index (`:114-118`), forwards it to the object (`:119`), and echoes it back in the result element (`:126` on success, `:134`/`:150` on error) with no array check of any kind. An indexed read of a scalar property via RPM therefore bypasses the exact check the RP path enforces — and since the object arms ignore the index today (#171), the client receives a full-value response labeled as an element read.

#190 covers reforming the RP whitelist into a per-object query, and cites the RP handler and the write paths — but not RPM.

### Suggested direction

Whatever shape #190 lands must cover the RPM path too; short of that, apply the same gate per property reference so RP and RPM agree. Note the special property identifiers (ALL/REQUIRED/OPTIONAL) never carry an index by construction (`:85-90` only attaches the request index when the identifier matches literally).

### Relevant code

- `crates/bacnet-server/src/handlers/read_property.rs`

### Related

#190, #171

Issue #182, opened 2026-07-27
WriteProperty and WritePropertyMultiple decode only one application-tagged primitive, so no structured property value can be written over the network


jscott3201:
### Affected revision

Observed on `dev` at commit `3990f3e`.

### Behavior

The WriteProperty handler reconstructs the incoming value with a single call to `decode_application_value` (`crates/bacnet-server/src/handlers/write_property.rs:218-219`); WritePropertyMultiple does the same per property (`:73`).

`decode_application_value` (`crates/bacnet-encoding/src/primitives/mod.rs:452-515`) returns exactly one application-tagged primitive. It rejects any context tag outright (`:457-462`) and any opening or closing tag (`:463-465`), and it has no branch that can produce `PropertyValue::List`. The matching encoder shows the asymmetry: `encode_property_value` flattens a `PropertyValue::List` into a bare concatenation of its children with no framing (`:535-539`), so a structured property that reads back correctly cannot be parsed back in.

`Recipient_List` is the visible casualty. `NotificationClassObject::write_property` requires a list of 7-element lists (`crates/bacnet-objects/src/notification_class/mod.rs:203-207`). The handler hands it the first destination's Valid Days bit string instead, so the `if let PropertyValue::List(entries) = value` test fails and the write returns `invalid_data_type_error()` at `:281` — PROPERTY / INVALID_DATA_TYPE (class 2, code 9). Every existing `Recipient_List` write test calls `nc.write_property(...)` directly with an already-built `PropertyValue::List` (`crates/bacnet-objects/src/notification_class/tests/recipient_list.rs:56`, `:77`, `:124`), so nothing in the suite exercises the handler's decode.

AddListElement and RemoveListElement fail the same way: both loop `decode_application_value` over the request payload and push each primitive onto the outer list (`crates/bacnet-server/src/handlers/list.rs:26-36` and `:75-85`), so an added `BACnetDestination` arrives as seven loose scalars rather than as one element. `Event_Parameters` is affected in degraded form — only the legacy `OctetString` to `Opaque` path survives (`crates/bacnet-objects/src/event_enrollment/mod.rs:209-224`).

[...]

PR #271, opened 2026-08-12
fix(encoding,services,objects)!: full ASN.1 framing for event/fault parameters, Recipient_List, and unified BACnetTimeStamp (#154, #152, #259)

### LINKED ISSUE #152: Encode Recipient_List with full ASN.1 CHOICE/SEQUENCE framing (follow-up to #126)
## Context

Follow-up to #126 (PR #151), which preserved the `BACnetAddress` network number in a `NotificationClass` `Recipient_List` address recipient at the **property-value** level (the `Address` variant now encodes as `List[Unsigned(network_number), OctetString(mac_address)]`). That fix closed the round-trip bug but did **not** make the wire encoding spec-correct, by design — the owner scoped #126 to the property-value round-trip and explicitly deferred the larger encoding-model reform to this follow-up.

## The remaining gap

The current `Recipient_List` encoding is a **simplification** of the spec's ASN.1, using flat application-tagged values with no context tags or SEQUENCE framing:

- `BACnetRecipient` is a `CHOICE { device [0] BACnetObjectIdentifier, address [1] BACnetAddress }` (Clause 21), but the code distinguishes `Device` vs `Address` by **application tag type** (`ObjectIdentifier` vs the new `List[Unsigned, OctetString]`), not by the spec's `[0]`/`[1]` context tags.
- `BACnetDestination` is a `SEQUENCE { valid-days, from-time, to-time, recipient, process-identifier, issue-confirmed-notifications, transitions }` (Clause 12.15.5), but the code emits the fields as a **flat `PropertyValue::List`** with no SEQUENCE framing.
- `BACnetAddress` is a `SEQUENCE { network-number Unsigned16, mac-address OCTET STRING }` — #126 preserved both fields but as a flat two-element list, not a framed SEQUENCE.

As a result, a conformant BACnet peer would **not** interoperate with rusty-bacnet's `Recipient_List` encoding. This is an internal-consistent-but-not-wire-spec-correct encoding.

## Proposed direction

- Rework the `Recipient_List` encode/decode to use real ASN.1: `[0]`/`[1]` context-tagged `BACnetRecipient` CHOICE, framed `BACnetAddress` SEQUENCE (`Unsigned16` + `OctetString`), and framed `BACnetDestination` SEQUENCE.
- This requires the `PropertyValue` enum (`crates/bacnet-types/src/primitives.rs:346`) — currently flat / application-tagged-only — to gain **context-tag / structured (opening-closing-tag) variants**, OR a dedicated encode/decode path in `bacnet-encoding` for `BACnetDestination`/`BACnetRecipient`/`BACnetAddress`. The NPDU layer (`bacnet-encoding/src/npdu.rs`) already encodes network+mac for DADR/SADR — assess whether a reusable codec can be shared.
- This is a **wire-format change** with a larger blast radius than #126 (touches the encoding layer and every `Recipient_List` consumer, possibly other flat-tag properties). Needs an owner design discussion before implementation.

## Relevant code

- `crates/bacnet-objects/src/notification_class/mod.rs` (Recipient_List encode/decode)
- `crates/bacnet-types/src/constructed/mod.rs` (`BACnetAddress`, `BACnetRecipient`, `BACnetDestination`)
- `crates/bacnet-types/src/primitives.rs` (`PropertyValue` — may need context-tag variants)
- `crates/bacnet-encoding/src/primitives/mod.rs` + `tags.rs` (context-tag helpers: `encode_ctx_*`, `encode_opening_tag`/`encode_closing_tag` already exist)

## Spec

ASHRAE 135-2020:
- `BACnetDestination ::= SEQUENCE { ... }` (Clause 12.15.5)
- `BACnetRecipient ::= CHOICE { device [0] BACnetObjectIdentifier, address [1] BACnetAddress }` (Clause 21)
- `BACnetAddress ::= SEQUENCE { network-number Unsigned16, mac-address OCTET STRING }` (Clause 21)

Relates to #126.

### LINKED ISSUE #154: Encode Event_Parameters/Fault_Parameters with full ASN.1 CHOICE framing (follow-up to #129)
## Context

Follow-up to #129 (PR #153), which modeled `Event_Parameters` as a structured `BACnetEventParameter` CHOICE and `Fault_Parameters` as a structured `FaultParameters` round-trip at the **property-value** level. The new encoding is a flat `PropertyValue::List` whose first element is the algorithm tag as `Unsigned`, followed by the variant's fields. That fix closed the round-trip/evaluator bug but did **not** make the wire encoding spec-correct, by design — the owner scoped #129 to the property-value round-trip (narrow now + follow-up) and explicitly deferred the larger encoding-model reform to this follow-up. This mirrors the #126→#152 deferred-encoding pattern.

## The remaining gap

The current `Event_Parameters` / `Fault_Parameters` encoding is a **simplification** of the spec's ASN.1, using flat application-tagged values with no context tags or SEQUENCE framing:

- `BACnetEventParameter ::= CHOICE { change-of-bitstring [0] SEQUENCE { time-delay [0] Unsigned, bitmask [1] BIT STRING, list-of-bitstring-values [2] SEQUENCE OF BIT STRING }, change-of-state [1] SEQUENCE { time-delay [0] Unsigned, list-of-values [1] SEQUENCE OF BACnetPropertyStates }, ... }` (ASHRAE 135-2020 Clause 13.5). The code distinguishes alternatives by a **leading `Unsigned` tag** in a flat `List`, not by the spec's `[0]`/`[1]`/.../`[9]` context tags, and the nested fields are flat application-tagged values with no inner context tags or SEQUENCE framing.
- `BACnetFaultParameter ::= CHOICE { fault-none [0] NULL, fault-characterstring [1] SEQUENCE { ... }, ... }` (Clause 13.5) — same flat-`List`-with-leading-`Unsigned` simplification.
- `BACnetDeviceObjectPropertyReference ::= SEQUENCE { object-identifier, property-identifier, property-array-index, device-identifier }` (Clause 21) — the `FloatingLimit` `setpoint-reference` and `FaultLifeSafety` `mode-for-reference` are encoded as a flat 4-element `List`, not a framed SEQUENCE.
- `BACnetPropertyStates ::= CHOICE { boolean-value [0] BOOLEAN, binary-value [1] Unsigned, ... }` (Clause 21) — the `ChangeOfState` `list-of-values` elements are encoded as a flat `List[Unsigned(tag), OctetString(data)]`, not the spec's context-tagged CHOICE.

As a result, a conformant BACnet peer would **not** interoperate with rusty-bacnet's `Event_Parameters`/`Fault_Parameters` encoding. This is an internal-consistent-but-not-wire-spec-correct encoding (same posture as the pre-#152 `Recipient_List`).

## Proposed direction

- Rework the `BACnetEventParameter` / `FaultParameters` encode/decode to use real ASN.1: context-tagged `[0]`..`[9]` CHOICE alternatives, framed inner SEQUENCEs, context-tagged nested fields, and a framed `BACnetDeviceObjectPropertyReference` SEQUENCE. Reuse the `BACnetPropertyStates` CHOICE codec once it too is context-tagged.
- This requires the `PropertyValue` enum (`crates/bacnet-types/src/primitives.rs:346`) — currently flat / application-tagged-only — to gain **context-tag / structured (opening-closing-tag) variants**, OR a dedicated encode/decode path in `bacnet-encoding` for `BACnetEventParameter`/`BACnetFaultParameter`/`BACnetDeviceObjectPropertyReference`/`BACnetPropertyStates`. Assess whether a reusable codec can be shared with the deferred #152 `Recipient_List` work (both need the same context-tag/SEQUENCE primitives).
- This is a **wire-format change** with a larger blast radius than #129 (touches the encoding layer and every `Event_Parameters`/`Fault_Parameters` consumer). Needs an owner design discussion before implementation. The `Opaque`/legacy-LE fallback introduced by #129 should be preserved as a decode-time compatibility path (or removed once full framing lands and a migration policy is decided).
- Coordinate with #152: both depend on the same context-tag/SEQUENCE primitives in `PropertyValue`/`bacnet-encoding`, so they should be designed together (one encoding-model reform, two consumer migrations).

## Relevant code

- `crates/bacnet-types/src/constructed/event_parameter/mod.rs` (`BACnetEventParameter`, `FaultParameters` encode/decode, `take_dopr`, `property_state_pv`/`property_state_from_pv`)
- `crates/bacnet-types/src/constructed/mod.rs` (`FaultParameters`, `BACnetDeviceObjectPropertyReference`, `BACnetPropertyStates`)
- `crates/bacnet-objects/src/event_enrollment/mod.rs` (read/write of `EVENT_PARAMETERS` / `FAULT_PARAMETERS`)
- `crates/bacnet-server/src/event_enrollment/mod.rs` (evaluator decode + `eval_legacy_le` fallback)

### LINKED ISSUE #259: The two BACnetTimeStamp codecs disagree on the time [0] alternative and cannot decode each other
### Affected revision

`dev` at `415eb32`. Found by the Event_Time_Stamps family grounding pass (2026-07-30).

### Behavior

The tree contains two independent BACnetTimeStamp codecs whose `time [0]` CHOICE alternative is encoded incompatibly:

- `encode_timestamp` (`crates/bacnet-encoding/src/primitives/mod.rs:553-560`) encodes `Time` as a context-specific **primitive** tag 0 of length 4 followed by the raw Time octets.
- `GetEventInformationAck::encode` (`crates/bacnet-services/src/alarm_event/get_event_information.rs:284-289`) encodes the same alternative as an **opening tag 0 / closing tag 0 constructed pair wrapping an application-tagged Time**.

Each decoder accepts only its own form: `decode_timestamp` tests `inner_tag.is_context(0)` (`primitives/mod.rs:592`), and `is_context` explicitly excludes opening/closing tags (`crates/bacnet-encoding/src/tags.rs:65-70`), so it rejects the services form; the services decoder accepts only `is_opening_tag(0)` (`get_event_information.rs:134`) and rejects the encoding-crate form. The `sequence-number [1]` alternative agrees between them; only `time [0]` diverges.

At most one of the two is conformant with Clause 20.2.1.4's context-tagging rules. A normative ruling on which is part of the Event_Time_Stamps family spec gate; whichever form survives, there should be exactly one codec — the #171 work wants a shared bare-CHOICE timestamp codec that both call sites use.

### Suggested direction

- Get the Clause 20.2.1.4 ruling (in flight), fix the losing codec, and collapse the two into one shared encode/decode pair.
- Round-trip tests across both call sites, all three alternatives.

### Relevant code

- `crates/bacnet-encoding/src/primitives/mod.rs`
- `crates/bacnet-services/src/alarm_event/get_event_information.rs`

### Related

#171, #169, #128

jscott3201:
## Summary

- `Event_Parameters`, `Fault_Parameters`, and `Recipient_List` are now encoded on the wire with the full ASN.1 CHOICE/SEQUENCE framing of their Clause 21 productions, and the tree's two divergent `BACnetTimeStamp` codecs are collapsed onto the one Clause 20.2.1.5-conformant codec.
- Scope boundary: encoding layer + the three affected object properties + their write/read paths; the event evaluator state machines and the `BACnetNotificationParameters` notification codec are deliberately untouched (separate tranches).

## Standard / ledger scope

- Standard: ANSI/ASHRAE 135-2020
- Anchors: Clause 20.2.1.5-20.2.1.6 (context-tag forms); Clause 21 productions `BACnetTimeStamp`, `BACnetEventParameter`, `BACnetFaultParameter`, `BACnetPropertyStates`, `BACnetDeviceObjectPropertyReference`, `BACnetDestination`/`BACnetRecipient`/`BACnetAddress`; Clause 12.21 `Recipient_List` + Table 12-25; Clause 12.12 `Event_Parameters`/`Fault_Parameters`; Annex K.2.25 (AE-CRL-B) 8-entry minimum posture
- Ledger rows: `BACNET-12-RECIPIENT-LIST-FRAMING`, `BACNET-12-EVENT-PARAMETERS-FRAMING` (split children of `BACNET-12-OBJECT-MODEL`), `BACNET-21-TIMESTAMP-CHOICE` (split child of `BACNET-21-FORMAL-APDUS`) — all at `supported-with-clause-evidence` with golden-vector + negative evidence listed; regenerated `support-summary.md` via `scripts/generate-conformance-docs.py` (`--check` passes)
- Support-scope note: supported for the modeled alternatives/properties with in-tree test evidence; unmodeled `BACnetEventParameter` alternatives remain Opaque-preserved and never reach the legacy little-endian evaluator (byte-for-byte for constructed forms; `none [20] NULL` decodes and re-encodes in the constructed form), not spec-decoded; no public claims broadened beyond the tested rows

## Implementation

Changed files (by commit):

- `crates/bacnet-encoding/src/primitives/mod.rs` — split the timestamp codec into a bare-CHOICE pair (`encode_timestamp_choice`/`decode_timestamp_choice`) + the existing wrapped pair built on them; enforce `Unsigned (0..65535)` for `sequence-number` on encode AND decode; `decode_application_value` maps context-tagged elements to raw bytes; `encode_property_value` emits raw bytes verbatim (#259, #154)
- `crates/bacnet-encoding/src/primitives/tests.rs` — timestamp golden/boundary/negative tests; raw-carrier value codec tests (#259, #154)
- `crates/bacnet-encoding/src/tags.rs` — add public `extract_raw_context` byte scanner for non-TLV vendor/legacy payloads (generalizes the pattern private to `bacnet-services`) (#154)
- `crates/bacnet-encoding/src/constructed/` (new module: `mod.rs`, `event_parameter.rs`, `fault_parameter.rs`, `recipient.rs`, `tests/`) — the full-framing codecs + golden/round-trip/negative tests (#154, #152)
- `crates/bacnet-types/src/primitives.rs` — new `PropertyValue::ApplicationData(Vec<u8>)` raw-carrier variant (one variant chosen over N context-tag variants: smaller public API change; trade-off noted below) (#154, #152)
- `crates/bacnet-types/src/enums/resolve.rs` + `crates/bacnet-types/src/constructed/event_parameter/{mod,fault}.rs` — resolver passthrough; doc refresh (flat form is now documented as the LEGACY compatibility layout) (#154)
- `crates/bacnet-services/src/alarm_event/get_event_information.rs` — deleted its private non-conformant timestamp codec; migrated to the shared bare-CHOICE codec; `GetEventInformationAck::encode` is now fallible (#259)
- `crates/bacnet-services/src/alarm_event/{acknowledge_alarm,event_notification}.rs`, `notification_parameters/encode.rs`, `audit.rs`, `cov_multiple.rs` — `?`-propagate the now-fallible timestamp encode; byte-identical output for in-range values (#259)
- `crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs`, `server/cov_notifications.rs`, `server/event_notifications.rs` — fallible encode plumbing; wrap out-of-range seconds-of-epoch `sequence-number` values mod 65536 instead of emitting non-conformant bytes (#259)
- `crates/bacnet-server/src/handlers/write_property.rs` — `decode_write_property_value`: `Recipient_List` (a BACnetLIST whose framed form is concatenated application-tagged destinations) is handed to the object as the complete `[3]` content; every other property decodes as before (#152)
- `crates/bacnet-server/src/handlers/tests/{mod.rs,framed_properties.rs}` — wire-level framed WriteProperty/WPM -> ReadProperty round-trips (#152, #154)
- `crates/bacnet-server/src/event_enrollment/mod.rs` — evaluator decodes the framed read-arm form (primary) or the legacy flat form (fallback); `Opaque -> eval_legacy_le` compatibility path preserved verbatim (#154)
- `crates/bacnet-objects/src/event_enrollment/mod.rs` — read arms emit framed bytes; write arms accept framed + legacy flat + legacy `OctetString`-to-`Opaque` (#154)
- `crates/bacnet-objects/src/notification_class/mod.rs` — read arm emits framed bytes; write arm decodes framed strictly + legacy flat as fallback; the flat-entry decoder is now ONE helper shared by the write arm and the routing filter (duplicate decode eliminated) (#152)
- `crates/bacnet-objects/src/{event_enrollment,notification_class}/tests/*` — wire-format-changed assertions moved to the framed form; legacy-fallback tests kept (#154, #152)
- `crates/bacnet-cli/src/output.rs`, `crates/rusty-bacnet/src/types/property_value.rs`, `crates/rusty-bacnet/rusty_bacnet.pyi` — exhaustive-match arms + a Python `PropertyValue.application_data(bytes)` constructor/”application_data“ tag/bytes getter for the new variant (#154)
- `docs/conformance/{standard-135-2020-ledger.md,bacnet-135-2020.json,support-summary.md}` — the three new ledger rows (#154, #152, #259)

Wire/API impact:

- Wire bytes changed? **Yes** — `Event_Parameters`, `Fault_Parameters` (CHOICE context-tag framing replaces flat application-tagged lists), `Recipient_List` on Notification Class (NotificationForwarder's property remains a stub) — BACnetDestination application-tagged SEQUENCE with ctx-tagged recipient replaces flat 7-field lists, `GetEventInformation-ACK` `time [0]` timestamps (primitive ctx tag 0 replaces opening-tag-0 wrapper), and server event/COV-notification `sequence-number` timestamps now wrap at 65535 instead of emitting out-of-range values. `time [0]` decode of the old opening-tag-0 form is rejected (that form had no in-repo producer left).
- Public Rust API changed? **Yes** — `PropertyValue` gains `ApplicationData(Vec<u8>)`; `decode_application_value` returns it for context-tagged input instead of erroring; `encode_timestamp`, `GetEventInformationAck::encode`, `COVNotificationMultipleRequest::encode`, and the two COV-multiple APDU encoders are now fallible (`Result`); new `bacnet_encoding::constructed` module (framed codecs); new public `tags::extract_raw_context`.
- Python API changed? **Yes (additive)** — `PropertyValue.application_data(bytes)` constructor; `.tag` may report `"application_data"`; `.value` returns `bytes` for it.
- WASM/JS API changed? No (`bacnet-wasm` not affected by these exports).
- CLI/docs changed? CLI formats the new variant (`application-data([..])`); conformance ledger docs updated as above.

## Tests

Commands run:

```bash
cargo fmt --all --check
cargo clippy --workspace --exclude rusty-bacnet --all-targets --locked
cargo test --workspace --exclude rusty-bacnet --exclude bacnet-wasm   # 2480 passed, 0 failed
cargo check -p rusty-bacnet --tests
python3 scripts/generate-conformance-docs.py --check
```

New/changed tests:

- `primitives::tests::timestamp_*_golden*` + `timestamp_sequence_number_boundaries` — Clause 20.2.1.5 byte vectors for all three alternatives (bare + wrapped forms); 0/65535 accept (#259)
- `primitives::tests::timestamp_sequence_number_over_65535_rejected` — encode-side rejection of 65536/u64::MAX; decode-side rejection of the 3-octet overflow form (#259)
- `primitives::tests::timestamp_decode_non_conformant_time_form_rejected` / `wrong_outer_tag` / `wrong_class_bit` / `truncated` — negative matrix (#259)
- `alarm_event::tests::get_event_information_timestamps::*` — full-ACK golden byte vector, per-alternative round-trips, cross-call-site matrix (primitives-encoded bytes decode through `GetEventInformationAck::decode` and vice versa), non-conformant-time-form / wrong-section-tag / truncation / 65536 rejections (#259)
- `constructed::tests::event_parameter::*` — golden vectors (ChangeOfState, ChangeOfValue x2, FloatingLimit, OutOfRange, Extended, Opaque); per-modeled-alternative round-trips; deprecated/reserved ctx tags 6/7/12/19 rejected in both primitive and constructed form; truncated/unbalanced/wrong-inner-tag negatives (#154)
- `constructed::tests::fault_parameter::*` — golden vectors for all eight `FaultParameters` alternatives; inner-CHOICE REAL/Unsigned/Double/INTEGER discovery; unknown-tag/NULL-with-contents/truncated negatives (#154)
- `constructed::tests::mod` (property states + DOPR) — Clause 21 CHOICE-tag goldens (incl. door/timer/lift 135-2020 tags), all-modeled-variant round-trips, unmodeled-tag `Other` preservation, class-form negatives (#154)
- `constructed::tests::recipient::*` — golden vectors device-form / address-form (nonzero network) / broadcast (zero-length MAC); 8-entry list round-trip; recipient-tag-2, truncated-member, unbalanced-address, Unsigned16-overflow negatives (#152)
- `bacnet-objects` event_enrollment/notification_class test suites — framed read-arm byte assertions, framed write round-trips (byte-identical re-emission), malformed-frame rejections; legacy flat write/fallback and OctetString->Opaque tests kept and passing (#154, #152)
- `bacnet-server` `framed_properties::*` — framed WriteProperty and WritePropertyMultiple bytes into `handle_write_property*` round-trip through `handle_read_property` byte-identically; reserved-tag write rejected with a PROPERTY-class error (#152, #154)
- `event_enrollment::tests::compat::framed_unmodeled_alternative_is_never_le_evaluated` — framed access-event [13] writes round-trip verbatim and produce ZERO transitions (pre-fix: spurious HIGH_LIMIT via the LE path) (#154)
- `notification_class::tests::recipient_list::recipient_list_indexed_write_rejected_list_unchanged` — indexed writes error with PROPERTY/INVALID_DATA_TYPE (dev-parity), list bytewise untouched (#152)
- `handlers::list::tests::{remove_list_element_from_framed_recipient_list_leaves_rest, remove_list_element_non_matching_entry_is_noop, add_list_element_to_framed_recipient_list_appends, remove_list_element_malformed_framed_payload_errors_and_preserves}` — destination-level element editing on the framed list (#152)
- `cov_notifications_tests::cov_multiple_timestamps_wrap_seconds_of_epoch_into_range` — both COV-multiple timestamp fields decode as sequence-number ≤ 65535 through the shared codec (#259)

## Benchmarks

Required? No — correctness tranche (encoding paths, not a measured hotspot; per the tranche mini-contract: "Benchmark question: none").

## Subagent reviews

- `bacnet-reference-researcher`: anchor sheet confirmed (Clause 20.2 tag forms; BACnetTimeStamp time [0] primitive-ctx ruling; full BACnetEventParameter/BACnetFaultParameter productions; BACnetDestination 7-member application-tagged SEQUENCE; BACnetAddress constructed so `address [1]` opens/closes tag 1); implementer re-verified the BACnetPropertyStates tags (restart-reason [14], door 15..=19, timer 43/44, lift 52/53), the fixed-width bit-string members (BACnetDaysOfWeek 7 bits / BACnetEventTransitionBits 3 bits, Clause 20.2.10), and the `none [0]/[20] NULL` primitive forms against the extracted 135-2020 text.
- `review-runtime`: READY after the first blocker round.
- `review-adversary`: 3 blockers found (Opaque->legacy-LE conflation; timeOfChange range bypass; indexed Recipient_List write regression) — all fixed with regression tests in `8ff9938` / `1c18cc8` / `0ac5a0f`.
- `review-dataflow`: 1 blocker found (RemoveListElement/AddListElement empty-fallback wiping framed Recipient_Lists) — fixed in `e503ae0` (destination-level decoding, bytewise no-op/append/remove tests).
- `bacnet-pr-packager`: 3 blockers found (trailing-bytes swallowed on framed writes `7365357`; fixed-width BIT STRING validation `3d629e2`; prefix-truncating stored-list decode now strict `e71f7f4`) plus provenance/wording fixes (`bfcc220`) — final confirmation pending after this push.

## Known limitations

- `BACnetPropertyStates` 135-2020 tag numbers are used by the NEW framed codec and the `Event_Parameters`/`Fault_Parameters` wire form, but the older `bacnet-services` `BACnetNotificationParameters` codec and the legacy flat fallback keep the invented pre-2020 tags for door/timer/lift variants (door 14..=18, timer 38/39, lift 40/42 vs 135-2020's 15..=19/43/44/52/53). Reconciling the notification codec changes notification wire bytes — follow-up (out of this PR).
- `extract_raw_context` skips vendor-defined/opaque bodies by scanning for the closing tag's raw octets; a payload containing the target tag's own open/close octet pattern cannot round-trip (documented on the function; affects only `extended.parameters` and `Opaque` bodies that are hostile/degenerate).
- `none [20] NULL` and other primitive unmodeled event-parameter tags decode to `Opaque { tag, data }` and re-encode in the constructed opening/closing form (value-preserving; the primitive NULL byte pattern itself is only produced for decode).
- The `RECIPIENT_LIST` write-path special case is keyed on the property identifier (Clause 12 fixes the datatype per property, e.g. also NotificationForwarder); the general whole-list decode gap for OTHER list properties remains #182, unchanged.
- Array-indexed reads of `Recipient_List` still return the whole list, and indexed writes are rejected with PROPERTY/INVALID_DATA_TYPE (restoring dev @ 6b9ac4f behavior; full indexed-array Recipient_List semantics are Tranche K / #260 family).
- `CreateObject` initial-value decode was not migrated (the network factory cannot create EventEnrollment/NotificationClass objects today).
- Server `sequence-number` timestamps wrap seconds-of-epoch mod 65536 (conformant range; previously emitted out-of-range values until encoding failed).

- Review-hardening behavior now in-tree: a framed UNMODELED spec alternative of `Event_Parameters` (e.g. access-event [13]) is stored and read back byte-identical but never routed to the legacy little-endian evaluator (only true raw-octet legacy writes via the 0xFF sentinel reach it); `AddListElement`/`RemoveListElement` edit framed recipient lists by decoded destination (malformed payloads are determinate `PROPERTY/INVALID_DATA_TYPE`, never a silent wipe); a stored Recipient_List that fails to decode now fails routing closed — no notification is delivered to that class (previously a decodable prefix would have been notified); COVNotificationMultiple wraps both its timestamps into the Clause-21 sequence-number range.

## Follow-ups

- Fix `bacnet-services` `property_states.rs`/`notification_parameters` to the 135-2020 property-state CHOICE tags (paired with notification wire checks; Tranche C territory).
- #182: multi-element whole-list writes for the remaining list properties.
- Second-stack interop evidence for the new framed forms (noted on the ledger rows).
- NotificationForwarder `Recipient_List` modeling (property currently a stub returning an empty list).

Closes #154
Closes #152
Closes #259


crates/bacnet-server/src/server/cov_notifications.rs
```diff
@@ -285,7 +285,12 @@ impl<T: TransportPort + 'static> BACnetServer<T> {
...
         let total_secs = now.as_secs();
-        let timestamp = BACnetTimeStamp::SequenceNumber(total_secs);
+        // Clause 21 constrains BACnetTimeStamp's sequence-number to
+        // Unsigned (0..65535) and the shared codec enforces it on encode.
+        // Wrap seconds-of-epoch into the valid window rather than letting a
+        // conformant-but-larger raw value fail encoding (previously this
+        // emitted a >65535 sequence number — non-conformant bytes).
+        let timestamp = BACnetTimeStamp::SequenceNumber(total_secs % 65_536);
         let mut timestamp_choice = BytesMut::new();
         encode_ctx_unsigned(&mut timestamp_choice, 1, total_secs);
```
jscott3201: [Adversary] BLOCKING — the wrap was applied to the request-level `timestamp` (line 293) but **not** to the per-item `timeOfChange` bytes built here: `encode_ctx_unsigned(&mut timestamp_choice, 1, total_secs)` still emits the raw seconds-of-epoch as a 4-octet ctx-[1] (sequence-number alternative) inside the `opening tag 3` wrap (cov_multiple.rs:333-337). With `total_secs ≈ 1.78e9` this emits `sequence-number > 65535` on the wire on every timestamped COV-notification-multiple — the exact non-conformance #259 claims to have eliminated ("server event/COV-notification `sequence-number` timestamps now wrap at 65535 instead of emitting out-of-range values" in the PR's wire-impact section; the code comment 3 lines above states the codec "enforces it on encode", which is only true for values routed through `encode_timestamp_choice` — this path bypasses it with the unchecked `encode_ctx_unsigned`).

Note the asymmetry: this tree's own `decode_timestamp_choice` (primitives/mod.rs:658-663) now **rejects** sequence-number > 65535, so a conformant receiver (including this stack acting as a COV client) rejects what this server emits. The fix is one line (`total_secs % 65_536` here, or better, route through `encode_timestamp_choice`). I verified the unwrapped form is pre-existing on dev, but it sits squarely inside this PR's declared #259 scope, so the "wrap" claim is only half-true as shipped.

crates/bacnet-server/src/event_enrollment/mod.rs
```diff
@@ -567,6 +567,17 @@ pub fn evaluate_event_enrollments(db: &mut ObjectDatabase) -> Vec<EventEnrollmen
         };
 
         let params = match enrollment.read_property(PropertyIdentifier::EVENT_PARAMETERS, None) {
+            // Framed wire form (the EventEnrollment object's read arm emits
+            // full ASN.1 CHOICE framing).
+            Ok(PropertyValue::ApplicationData(bytes)) => {
+                match bacnet_encoding::constructed::decode_event_parameter(&bytes, 0) {
```
jscott3201, line 573: [Adversary] BLOCKING — the `Opaque -> eval_legacy_le` fallback (line 665 below) claims to be "preserved verbatim", but the *data* flowing into it changed semantics with this PR, and the mismatch is silent.

Pre-PR, `Opaque` payloads could only originate from the legacy flat codec paths, so `eval_legacy_le`'s little-endian reinterpretation matched the bytes it received. Now the read arm emits full ASN.1 framing and **every conformant framed-but-unmodeled alternative** (`command-failure [3]`, `change-of-life-safety [8]`, `access-event [13]`, `change-of-discrete-value [21]`, `change-of-timer [22]`, vendor tags) decodes to `Opaque { tag, data }` where `data` is the framed TLV body (bacnet-encoding/src/constructed/event_parameter.rs:401-415). `eval_legacy_le` (line 414) then reads those TLV bytes as LE f32 limits / LE u32 alarm values depending on the enrollment's own `Event_Type`.

Concrete scenario: a peer writes a conformant `Event_Type = OUT_OF_RANGE` enrollment with framed `Event_Parameters = access-event [13]` (body ≥ 12 octets — two ctx-unsigneds plus a DOPR easily qualifies). The object's write arm (bacnet-objects/src/event_enrollment/mod.rs:292-296) accepts it — correctly. The evaluator then interprets the first 12 TLV bytes as `[high_limit LE][low_limit LE][deadband LE]` (line 96-98) and the enrollment spuriously transitions HIGH_LIMIT/LOW_LIMIT and fires event notifications. Nothing rejects or logs; the "compatibility path preserved verbatim" wording describes intent, not behavior.

Expected: gate `eval_legacy_le` to `Opaque` values produced by the *legacy* write forms (e.g. the `tag: 0xFF` sentinel from `OctetString`, objects/event_enrollment/mod.rs:287-290) and treat framed-decoded `Opaque` as inert (`continue`), or make `Opaque` carry a marker of its provenance.

crates/bacnet-server/src/handlers/write_property.rs
```diff
@@ -222,6 +222,29 @@ pub fn handle_write_property_multiple(
...
+/// discard the rest (the whole-list decode gap tracked as #182). The object
+/// layer owns the framed codec, so hand the complete content over verbatim;
+/// the property's Clause 12 datatype is fixed by its identifier wherever it
+/// appears (Notification Class, Notification Forwarder).
+pub(crate) fn decode_write_property_value(
+    property: PropertyIdentifier,
+    bytes: &[u8],
+) -> Result<PropertyValue, Error> {
+    if property == PropertyIdentifier::RECIPIENT_LIST {
+        return Ok(PropertyValue::ApplicationData(bytes.to_vec()));
```
jscott3201, line 236: [Adversary] BLOCKING — the `RECIPIENT_LIST` verbatim bypass ignores `property_array_index`, turning a previously-rejected indexed write into a silent whole-list replacement.

A WriteProperty of `RECIPIENT_LIST` with `propertyArrayIndex = n` carries the bytes of ONE `BACnetDestination` (a legal array-element write). This special case passes those 7-member concatenated bytes through as `ApplicationData`, and `NotificationClass::write_property` (bacnet-objects/src/notification_class/mod.rs:164 takes `_array_index` and ignores it) decodes the value with `decode_destination_list` and **replaces the entire list** with that single destination.

Regression check against dev: on dev the same indexed write reached `decode_application_value`, which consumed only the first member (the valid-days BIT STRING) and produced `PropertyValue::BitString`, which the write arm then rejected with INVALID_DATA_TYPE. So: dev = clean rejection; this PR = silently truncating a peer-configured recipient list to one entry — notifications for all other recipients stop with no error returned to the writer. The doc comment frames the special case as a whole-list decode fix (#182) but doesn't mention the indexed-write case at all. Either reject `array_index.is_some()` for this property or implement indexed element replacement.

crates/bacnet-encoding/src/constructed/recipient.rs
```diff
@@ -0,0 +1,297 @@
...
+    if tag.class != TagClass::Application || tag.number != tags::app_tag::BOOLEAN {
+        return Err(Error::decoding(
+            pos,
+            format!("{what}: expected application-tagged BOOLEAN"),
+        ));
+    }
+    let issue_confirmed_notifications = tag.length != 0;
+    // transitions: BACnetEventTransitionBits (3 bits).
+    let ((_, transitions_data), pos) = decode_app_bit_string(data, pos, what)?;
+    let transitions = unpack_octet(&transitions_data, 3);
```
jscott3201, line 192: [Adversary] Non-blocking — silent default on nonconformant input in the new recipient decoder (recipient.rs:154 and here): `decode_app_bit_string` (line 277) accepts a `BIT STRING` with **zero content octets** (length 1 = unused-bits count only), and `unpack_octet(&[], 3)` then yields `0`. A destination whose `valid-days` or `transitions` arrives as this truncated form is silently stored as "no days" / "no transitions" — i.e. it will never receive notifications, with no error and no log. Since `BACnetDaysOfWeek` is defined as exactly 7 bits / 1 content octet (and EventTransitionBits 3 bits / 1 octet), rejecting `unused_bits/0-payload` or `data.len() != 1` would fail loudly instead of creating a dormant destination. Note the write arm is documented as "strict — one malformed entry rejects the whole write", but this malformed form passes.

crates/bacnet-encoding/src/tags.rs
```diff
@@ -334,6 +334,61 @@ pub fn extract_context_value(
...
+            continue;
+        }
+        pos += 1;
+    }
+
+    Err(Error::decoding(
+        value_start,
+        format!("missing closing tag {tag_number}"),
+    ))
+}
```
jscott3201, line 390: [Adversary] Non-blocking — the documented "hostile/degenerate" framing of this limitation understates how easily it triggers. For narrow (single-octet) tag contexts the sentinel is a common byte: context tag 2's closing octet is `0x2F` = ASCII `'/'`, and tag 9's is `0x9F` (a UTF-8 continuation byte that can appear inside any multi-byte CharacterString). Since `extract_raw_context` is used for `extended.parameters` (event_parameter.rs:389, fault_parameter.rs:233) and for Opaque capture of unmodeled alternatives (event_parameter.rs:407), a *conformant* vendor payload containing `'/'` inside `parameters [2]` mis-slices the capture. Concrete consequence today: the mis-slice shifts the returned offset, so the caller's subsequent `expect_closing` (e.g. event_parameter.rs:391) fails and the whole framed write is rejected — loud, thankfully — but re-emission of an accepted `Extended`/`Opaque` value whose captured `data` was truncated mid-body can also emit unbalanced framing on the next read. Also note the balancing logic only counts *same-numbered* nested opens; a body containing opening tag 2 nested under opening tag 3 is balanced here even though the TLV structure says otherwise. Suggest at minimum a debug/warn log on the truncation path, and (if round-trip fidelity of Extended matters) deriving the end offset from the enclosing value's known bounds instead of raw scanning.

jscott3201 (commented):
[Adversary] Security/reliability panel — **NOT READY** (3 blocking findings). (Submitted as COMMENT because GitHub rejects REQUEST_CHANGES on an own-authored PR; these findings are blocking, not advisory.)

Scope of my pass: hostile-input exercise of the new decoders (event_parameter / fault_parameter / recipient), `extract_raw_context`, the framed-vs-legacy/Opaque fallback dispatch, the write-path routing, recipient CHOICE discrimination, timestamp range enforcement, and dev-baseline verification of claimed pre-existing behavior. (Static analysis only — no `cargo clippy` runner available; all findings are code-traced with base/HEAD comparison via the API.)

## BLOCKING

1. **Opaque → legacy-LE evaluator conflation** (`crates/bacnet-server/src/event_enrollment/mod.rs:573,665`). The "Opaque → eval_legacy_le compatibility path preserved verbatim" claim is false as shipped: the *path* is preserved, but the bytes it receives changed meaning. Every conformant framed-but-unmodeled alternative (`access-event [13]`, `change-of-life-safety [8]`, …) now decodes to `Opaque { data = framed TLV body }` and gets reinterpreted as little-endian limits/alarm-values keyed on `Event_Type` — spurious HIGH/LOW_LIMIT transitions and outbound notifications on conformant input, silently. Gate the legacy path to sentinel-tagged Opaque (legacy writes) only. (Inline comment at `event_enrollment/mod.rs:573`.)

2. **COV-multiple `timeOfChange` still emits out-of-range sequence-number** (`crates/bacnet-server/src/server/cov_notifications.rs:295`). The wrap-mod-65536 was applied to the request `timestamp` (line 293) but not to the per-item `timeOfChange` bytes, which bypass the new range check via unchecked `encode_ctx_unsigned(ctx 1, total_secs)` — permanently non-conformant wire bytes (~1.78e9 today, growing), directly contradicting the PR's wire-impact claim for #259 and rejected by this tree's own new `decode_timestamp_choice`. Pre-existing defect, but squarely in the PR's declared fix scope.

3. **Indexed `RECIPIENT_LIST` write silently replaces the whole list** (`crates/bacnet-server/src/handlers/write_property.rs:242-244` + `crates/bacnet-objects/src/notification_class/mod.rs:164`). `propertyArrayIndex` is ignored end-to-end; a legal single-destination element write now destroys the rest of the list (dev rejected it cleanly with INVALID_DATA_TYPE). Silent data loss introduced by this PR.

## NON-BLOCKING follow-ups

- **Pre-existing panic verified** (not a regression): `GetEventInformationAck::decode` `more_events = data[pos] != 0` (bacnet-services/src/alarm_event/get_event_information.rs:212) indexes out of bounds when the trailing ctx tag has zero content octets — a one-byte hostile-ACK crash. Byte-identical on dev @ 81848f78; fix in passing since this codec is already being reworked in this PR.
- Truncated `BIT STRING` (unused-count octet only) silently decodes `valid_days`/`transitions` to 0 (`crates/bacnet-encoding/src/constructed/recipient.rs:154,180`) — dormant destination instead of write rejection. (Inline at recipient.rs:180.)
- `extract_raw_context` limitation is more reachable than "hostile/degenerate" suggests: `'/'` (0x2F) is tag-2's closing octet and `0x9F` (tag 9's) is a valid UTF-8 continuation byte; mis-slice is loud at write time but can emit unbalanced framing on re-emission of truncated captures. (Inline at tags.rs:390.)
- `decode_event_parameter`/`decode_fault_parameters` call sites discard the consumed offset (`Ok((ep, _))`, bacnet-objects/src/event_enrollment/mod.rs:293,311) — garbage after the first CHOICE alternative is accepted and silently dropped on read-back.
- Routing filter `decode_destination_list_pv` (bacnet-objects/src/notification_class/mod.rs:515) truncates at the first malformed entry vs. the strict write arm — documented divergence, but recipients after a bad entry silently vanish.
- Sequence-number wrap inherently duplicates every 65536 s; comments note the wrap but not the duplicate-window behavior for Table 13-3 consumers.
- Verified clean: `Recipient` ctx0/ctx1 discrimination (length≠4 device form rejected, zero-length MAC = broadcast OK), loop monotonicity, MAX_FRAMED_ITEMS caps, no quadratic scans, timestamp seq-number encode/decode rejection (no panic path), fallible-encoder plumbing, and the two claimed "pre-existing" spots are genuinely on dev.

crates/bacnet-objects/src/event_enrollment/mod.rs
```diff
@@ -277,6 +288,15 @@ impl BACnetObject for EventEnrollmentObject {
                     tag: 0xFF,
                     data: bytes,
                 },
+                // Framed wire form: full ASN.1 CHOICE framing per Clause 21.
+                PropertyValue::ApplicationData(bytes) => {
+                    match bacnet_encoding::constructed::decode_event_parameter(&bytes, 0) {
```
jscott3201, line 296: **Blocking: the framed write path accepts trailing bytes and then drops them.** `decode_event_parameter(&bytes, 0)` returns both the decoded value and the consumed offset, but this call discards the offset and accepts any suffix. For example, a valid `OutOfRange` frame followed by `0x5e 0x09` is stored as the first frame; `read_property` re-encodes only that first alternative, so the peer's trailing wire data is silently lost. The same issue exists for `FAULT_PARAMETERS` at line 311. This contradicts the PR's byte-preservation/framing contract and turns malformed or concatenated property input into a successful lossy write. Require the returned offset to equal `bytes.len()` (or explicitly preserve/parse the complete property grammar) before storing.

crates/bacnet-objects/src/notification_class/mod.rs
```diff
@@ -504,102 +438,142 @@ pub fn get_notification_recipients(
...
+/// stops the walk, keeping the entries before it (read path stays
+/// best-effort). Legacy flat form ([`PropertyValue::List`]):
+/// [`destination_from_flat_fields`], skipping malformed entries.
+fn decode_destination_list_pv(value: &PropertyValue) -> Vec<BACnetDestination> {
+    match value {
+        PropertyValue::ApplicationData(bytes) => {
+            let mut destinations = Vec::new();
+            let mut pos = 0;
+            while pos < bytes.len() {
+                match bacnet_encoding::constructed::decode_destination(bytes, pos) {
```
jscott3201: **Blocking: malformed framed list suffixes are silently truncated by the routing filter.** `decode_destination_list_pv` stops at the first decoder error and returns the valid prefix, while the framed write arm is strict. A stored/read `Recipient_List` containing one valid destination followed by a malformed/truncated destination therefore causes routing to deliver only the prefix and silently drops later configured recipients, without rejecting or surfacing the corrupt property. This conflicts with the row's strict framing/data-integrity claim and the e503ae0 handler fix. Either make this path strict and fail closed, or define/test an explicit error-preserving behavior rather than treating malformed suffixes as a valid list.

crates/bacnet-encoding/src/constructed/recipient.rs
```diff
@@ -0,0 +1,297 @@
...
+        ),
+    ))
+}
+
+/// Decode one [`BACnetDestination`] at `offset`; returns it and the offset
+/// past the last member.
+pub fn decode_destination(data: &[u8], offset: usize) -> Result<(BACnetDestination, usize), Error> {
+    let what = "BACnetDestination";
+    // valid-days: BACnetDaysOfWeek (bit 0 = Monday, MSB-first on the wire).
+    let ((_, days_data), pos) = decode_app_bit_string(data, offset, what)?;
```
jscott3201: **Blocking: the documented strict Recipient_List decoder accepts invalid fixed-size BIT STRINGs.** `decode_destination` calls `unpack_octet` on `valid_days` and `transitions` without checking that the decoded BIT STRING has its required one data octet and the expected unused-bit count. A wire value with only the unused-bits octet (or an incorrect bit width) is accepted as an all-zero field, so a malformed destination can be stored successfully and then never route notifications. The write arm is explicitly documented as strict at `crates/bacnet-objects/src/notification_class/mod.rs:187-194`, but this malformed input passes the codec. Reject nonconformant fixed-width `BACnetDaysOfWeek`/`BACnetEventTransitionBits` forms before unpacking, and add a negative test.

jscott3201:
[Luna] Follow-up readiness: **READY** at `a29d4f2410f0f41b914641099aa19a77192e7e4d`.

Verified fixed: trailing framed-write bytes are rejected with consumed-offset checks and regression tests (`7365357`); Recipient destination BIT STRING widths are validated with negative coverage (`3d629e2`); malformed stored Recipient_Lists now fail whole-list/closed routing with regression coverage (`e71f7f4`). Provenance and generated conformance docs are aligned, with the documented convention that `repo_sha` pins the last code commit `265b267` before the docs-only refresh; the PR body now records the actual review outcomes and conservative limitations.

CI is green for the required validation jobs: Test (ubuntu-latest), Clippy, Rustfmt, no-secret scan, and file-size cap all succeeded. Matrix/release/audit jobs are skipped by workflow conditions, not failed.

Residual non-blocking follow-ups remain the documented second-stack interop evidence, NotificationForwarder Recipient_List modeling, migration of the older notification-parameters property-state tags, and the raw opaque-context scanner limitation. Recommend **squash merge** to match repository history.

Issue #277, opened 2026-08-12
Enumerated values longer than 4 octets silently truncate on decode


jscott3201:
Found by adversarial review of PR #276; pre-existing on dev, decode layer.

`crates/bacnet-encoding/src/primitives/mod.rs` (~line 524) decodes `Enumerated` as `decode_unsigned(content)? as u32`. A non-minimal 5-8 octet enumerated wire value silently truncates to its low word, so a hostile/buggy peer can encode `0x00000001_00000002` and have it stored as the valid value `2`, bypassing intake validation (e.g. the Notify_Type / Reliability / DoorValue range gates from #276).

Fix direction: strict length check (`u32::try_from` equivalent: reject when content > 4 octets, or when leading octets are non-zero beyond the 32-bit range) producing an INVALID_DATA_ENCODING / decoding error. Audit the same pattern for `Unsigned` decodes that truncate to u32/u16.

Issue #270, opened 2026-08-06
Feature request: writable/settable Relinquish_Default on commandable objects


jscott3201:
## Current behavior

`Relinquish_Default` is fixed at construction and unreachable afterward: hardcoded per constructor (AO/AV `0.0`, BO/BV `0`, MSO/MSV `1`), no setter, no constructor parameter, and the write arm returns `WRITE_ACCESS_DENIED` (`common.rs` notes RELINQUISH_DEFAULT / CURRENT_COMMAND_PRIORITY are read-only in every implementation; confirmed by runtime replay at dev/6b9ac4f).

## Why it matters

A site-configured safe default is ordinary BACnet practice, and Standard 135 permits Relinquish_Default to be writable. Today a consumer modeling configured defaults must permanently occupy priority slot 16 as a workaround — which works, but makes the wire-readable RELINQUISH_DEFAULT property disagree with the effective default, an observable infidelity for any client (or fault-detection rule) that reads the property.

## Ask

A `set_relinquish_default` method (or constructor parameter) on the commandable objects, validated the same way PV writes are (finite Real for analog, enum range for binary, 1..=Number_Of_States for multistate), with `recalculate_present_value()` after the change. A WriteProperty arm could follow separately if you want wire-writable semantics; the programmatic setter alone unblocks the consumer pattern.

jscott3201:
## Correction: the slot-16 workaround does not work, and this is blocking rather than a convenience

Verified against ANSI/ASHRAE 135-2020 directly. The "Why it matters" paragraph above understates this — it says occupying priority slot 16 "works, but" produces one observable infidelity. It produces three, and the third makes the workaround unusable rather than merely imperfect.

**1. `Priority_Array` reads back wrong.** It is a fixed 16-element `SEQUENCE SIZE (16) OF BACnetPriorityValue` (§21, p.925), each element a CHOICE carrying an explicit `null NULL` tag distinct from every value tag (§21, p.926). A real device with a configured default and nothing commanding it encodes all 16 slots NULL; the workaround encodes slot 16 with an application-tagged value. `Priority_Array` is required and readable on every commandable object (§12.3 Table 12-3, p.172), so any client sees this with no interpretation needed.

[...]

PR #278, opened 2026-08-12
feat(server,encoding,objects): structured property values over WriteProperty/WPM (#182)

### LINKED ISSUE #182: WriteProperty and WritePropertyMultiple decode only one application-tagged primitive, so no structured property value can be written over the network

jscott3201:
## Summary

- **Generic multi-element write decode (#182 core):** `decode_write_property_value` consumed exactly one application-tagged primitive and silently dropped the rest, so no structured or `BACnetLIST` property value could reach an object write arm whole. It now loop-decodes with `decode_application_value` until the payload is exhausted — the mirror of `encode_property_value`'s `List` flattening: one element → scalar `PropertyValue` (unchanged), more than one → `PropertyValue::List`. Full consumption is required: a partial or undecodable trailing element (and an empty payload) is PROPERTY / INVALID_DATA_ENCODING per Clause 15.9.1.3, never a silent drop. The RECIPIENT_LIST verbatim special-case keeps its precedence, and the context-tagged → `ApplicationData` behavior is preserved (framed EE/Fault writes flow exactly one `ApplicationData`, unaffected).
- **Structured reference properties:** Loop `CONTROLLED_VARIABLE_REFERENCE` / `MANIPULATED_VARIABLE_REFERENCE` / `SETPOINT_REFERENCE` (Clause 12.17), Pulse Converter `INPUT_REFERENCE` (Clause 12.23), and — after the panel round — Averaging `OBJECT_PROPERTY_REFERENCE` (Clause 12.5) accept their Clause 21 wire form per the tranche-J pattern: a strict typed codec in `bacnet-encoding::constructed` plus a shared arm helper accepting both that and the legacy local `List` form (property-id member tolerated as Unsigned OR Enumerated — Averaging's flat convention carries Unsigned, Loop/Pulse carry Enumerated). The optional array-index member is carried end to end; device-qualified members ([3]) are rejected (Loop/Pulse Converter productions have no device member; Averaging's `BACnetDeviceObjectPropertyReference` typing leaves remote sampling optional and it is unmodeled — refused, not silently local-ized).
- **Unlocks:** MSI `ALARM_VALUES` whole-list writes (pin flipped); DateTime Value `PRESENT_VALUE` and `PRIORITY_ARRAY` entries over the network; datetime `Relinquish_Default` on DateTime Value + DateTime Pattern Value — the two temporal exclusions tranche L1 deferred pending this decode, completing #270 across all twelve commandable value types.
- **Conformant absent-member frame:** the empty `BACnetSetpointReference` frame (`0x0E 0x0F`) clears `Setpoint_Reference` exactly like a `Null` write — the production's member is OPTIONAL and Clause 12.17 defines its absence as the fixed-setpoint state (runtime + dataflow panel finding).
- Scope boundary: write decode + the five reference arms + the three named unlocks + the panel-round hardening. No encode-path changes beyond what round-trip tests need (reference encoders + read arms carrying the array index), no unrelated-arm refactors, no CI changes, no support-claim broadening. #277 (Enumerated truncation) explicitly untouched and not regressed.

## Standard / ledger scope

- Standard: ANSI/ASHRAE 135-2020 (local extract `/var/folders/.../std135-2020.txt`; line cites verified against it).
- Anchors:
  - `BACnetObjectPropertyReference ::= SEQUENCE { object-identifier [0], property-identifier [1], property-array-index [2] Unsigned OPTIONAL }` (Clause 21; extract line 66943) — no device member exists in this production.
  - `BACnetSetpointReference ::= SEQUENCE { setpoint-reference [0] BACnetObjectPropertyReference OPTIONAL }` (Clause 21; extract line 70304); Loop `Setpoint_Reference` text: "contains zero or one references. The absence of a reference indicates that the setpoint for this control loop is fixed and is contained in the Setpoint property." (extract lines 18718-18720) — **empty frame = conformant clear**, no write prohibition exists in the property text.
  - Loop Object Type Clause 12.17 / Table 12-20 (extract line 18334).
  - Pulse Converter Clause 12.23 (extract line 21100) / Table 12-27 (extract line 21172); `Input_Reference` text at extract line 21344ff. **Brief correction:** the tranche brief said "Accumulator INPUT_REFERENCE (accumulator.rs:~371)"; `accumulator.rs` also holds `PulseConverterObject`, and `Input_Reference` is the Pulse Converter property (Accumulator has none in Table 12-79).
  - Averaging Clause 12.5 (extract line 13142) / Table 12-5 (extract line 13154): `Object_Property_Reference  BACnetDeviceObjectPropertyReference  R1`; the clause text makes remote-device sampling OPTIONAL ("Optionally, the object property to be sampled may exist in a different BACnet device").
  - Clause 15.9.1.3 WriteProperty error table: "The encoding is not valid for the datatype of the property." (extract line 53214); Clause 15.10 WritePropertyMultiple in-order application.
  - Value-object tables: DateTime Value Clause 12.38 / Table 12-45 (extract lines 29458/29467); DateTime Pattern Value Clause 12.46 (extract line 32765).
- Ledger rows: **new** `BACNET-15-STRUCTURED-WRITE-DECODE` and `BACNET-12-REFERENCE-PROPERTY-WRITABILITY`, **updated** `BACNET-12-RELINQUISH-DEFAULT-WRITABILITY` (datetime unlock folded in; pinned-denial negative replaced), `BACNET-15-ARRAY-INDEX-GATING` (reference-property indexed-read pins) — all `supported-with-clause-evidence`; `repo_sha` = last code commit `0c7be9744ffa8540ef95a7b84951ff376831c532`; `reviewed_at` 2026-08-12.
- Support-scope note: draft evidence only; no new public claims.

## Implementation

Changed files (grouped by commit):

- `f62cc85` feat(server): loop-decode multi-element WriteProperty values with full consumption (#182)
  - `crates/bacnet-server/src/handlers/write_property.rs` — `decode_write_property_value` loop-decodes to scalar/List with the INVALID_DATA_ENCODING mapping; RECIPIENT_LIST special-case ordering and context-tag widening preserved.
  - `crates/bacnet-server/src/handlers/tests/multi_element_writes.rs` (new) — the trailing-garbage matrix.
- `40aae9b` feat(encoding,objects): BACnetObjectPropertyReference wire codec + reference write arms (#182)
  - `crates/bacnet-encoding/src/constructed/object_property_reference.rs` (+tests) — `encode/decode_object_property_reference` (bare members; decode reuses the tranche-J `decode_dopr_body` then narrows: full consumption, [3] rejected) and `encode/decode_setpoint_reference` (the [0] opening/closing frame).
  - `crates/bacnet-objects/src/reference.rs` (+tests) — shared arm decode helper.
  - `crates/bacnet-objects/src/loop_obj.rs`, `accumulator.rs` — the four arms route through the helper; read arms carry the optional array index as a third list element.
- `8fad15a` style: cargo fmt over the #182 reference-codec files
- `1f29ca8` feat(objects): unlock datetime-paired writes + whole-list writes over WP (#182, #270) — `rd_access: writable` on the datetime-paired pair; both pins flipped; datetime wire tests.
- `96a5e4c` + `39ed0e9` docs: ledger rows + CHANGELOG (citation corrected to Clause 12.23 after extract verification).
- **`715803e` fix(objects)!: strict reference-arm decode — exact shape, >u32 guards, absent BACnetSetpointReference member (#182)** — the adversary blocker:
  - `crates/bacnet-objects/src/averaging.rs` — the arm routes through the shared decode: exact 2-or-3 member shape (4+-member writes refused instead of silently dropped), `u32::try_from` on the numeric members (`as u32` truncation gone), wrong-typed members refused, framed-form full consumption, device-qualified [3] refused INVALID_DATA_ENCODING with the local-sampling posture cited. (The Averaging `is_writable_property` override rode in this commit — see next entry for the advertising half.)
  - `crates/bacnet-objects/src/reference.rs` — flat-form property member accepted as Unsigned OR Enumerated (compat decision, documented); framed decode returns `Option` for the setpoint frame.
  - `crates/bacnet-encoding/.../object_property_reference.rs` — `decode_setpoint_reference` returns `Option`: empty frame → `None` (absent OPTIONAL member), with golden vectors for the indexed-in-frame encoding and the empty frame.
  - Tests: averaging exact-shape matrix (2/3 members, both member typings, read-back fidelity) + refusal matrix (4-member, wrong-typed third, non-OID first, >u32 member, device frame — state preserved each time) + framed-Local acceptance; `loop_write_empty_setpoint_frame_clears_reference`; helper-level empty-frame scope test.
- **`2398902` fix(objects,server): mirror PICS writability advertising to the Pulse Converter write arms (#182)** — the dataflow item: Pulse Converter override advertises exactly PRESENT_VALUE / SCALE_FACTOR / ADJUST_VALUE / INPUT_REFERENCE + DESCRIPTION / OUT_OF_SERVICE / COV_INCREMENT; Averaging (override in `715803e`, flagged here because no amend/rewrite per tranche rules) advertises OBJECT_PROPERTY_REFERENCE + DESCRIPTION / OUT_OF_SERVICE; both STOP advertising OBJECT_NAME (no arm routes it). Pinned in `pics/truth_source_tests.rs::is_writable_property_matches_write_property_on_pulse_converter_and_averaging` (advertised accepted with a good value, unadvertised rejected with a plausible one, both directions).
- **`0c7be97` test(server): split reference wire tests into reference_writes.rs + indexed-read pins (#182)** — `multi_element_writes.rs` crossed the 700-LOC cap; the seven reference wire tests live in self-contained `reference_writes.rs` (framed writes per arm, flattened-local form, rejections, empty-frame + Null clears, averaging cases, WPM commit+rollback); `array_index_gating.rs` pins the five reference-typed properties rejecting indexed ReadProperty (PROPERTY / PROPERTY_IS_NOT_AN_ARRAY).
- `badda94` docs: review-round ledger sync + changelog.

Wire/API impact:

- Wire bytes changed? Only on paths that were previously wrong: multi-element writes now reach arms (previously the tail was dropped and only element 1 applied); malformed value payloads map to PROPERTY / INVALID_DATA_ENCODING instead of SERVICES/OTHER from a propagated raw decoding error; reads of an indexed reference show a third (index) list element; the empty `BACnetSetpointReference` frame now clears instead of erroring; 4+-member averaging writes now refuse instead of silently storing a prefix. Accepted single-element writes encode identically; no other encoding/layout change.
- Public Rust API changed? Yes, additive: `bacnet_encoding::constructed::{encode,decode}_{object_property_reference,setpoint_reference}` (`decode_setpoint_reference` returns `Option`); `pub(crate)` `bacnet_objects::reference` helper module.
- Behavior changes (flagged): (1) trailing bytes on WriteProperty/WPM values decide the write now — previously dropped. (2) The reference arms reject legacy-List shapes with 4+ members that they previously truncated silently (`items.len() >= 2` → exact shape). (3) Reference reads carry the array index when set. (4) Previously-denied datetime RD writes accepted (additive). (5) Empty setpoint frame clears (was refused). (6) Pulse Converter / Averaging PICS writability advertising corrected toward arm reality (INPUT_REFERENCE / OBJECT_PROPERTY_REFERENCE advertised; OBJECT_NAME unadvertised on both).
- Python API changed? No (`cargo check -p rusty-bacnet --tests` clean).
- WASM/JS API changed? N/A (crate removed).
- CLI/docs changed? No CLI change; conformance docs regenerated from the ledger.

## Tests

Commands run (all on branch `codex/write-path-decode`, final tree):

```bash
bash .github/scripts/check-file-size.sh        # OK: all tracked .rs files within the 700 LOC cap
cargo fmt --all --check                        # pass
cargo clippy --workspace --exclude rusty-bacnet --all-targets --locked   # exit 0 (pre-existing warnings only)
cargo test --workspace --exclude rusty-bacnet --exclude bacnet-wasm      # exit 0: 34 suites, 2580 passed, 0 failed
                                               # (objects 959; server 298; encoding 232)
cargo check -p rusty-bacnet --tests            # exit 0
python3 scripts/generate-conformance-docs.py --check                     # pass
```

New/changed tests (highlights):

- Codec: `golden_vector_unindexed` / `golden_vector_indexed` / `setpoint_reference_golden_vector` / **`setpoint_reference_golden_vector_indexed`** (exact bytes incl. the [2] member inside the frame); `bare_decode_rejects_{device_qualified_reference,unknown_trailing_context_tag,partial_and_empty}`; `setpoint_reference_empty_frame_is_the_absent_alternative` (`0x0E 0x0F` → `None`); `setpoint_decode_requires_the_frame_and_full_consumption`.
- Arm helper matrix (`reference/tests.rs`): legacy 2-/3-member round-trips; `legacy_list_accepts_unsigned_or_enumerated_property_member`; shape violations (4 members, wrong-typed members, >u32 members) → INVALID_DATA_TYPE; framed negatives → INVALID_DATA_ENCODING; `empty_setpoint_frame_clears_only_on_the_setpoint_arm` (Setpoint → None; Bare → INVALID_DATA_ENCODING).
- Averaging (blocker): `averaging_reference_write_accepts_exact_shapes_and_both_member_typings`, `averaging_reference_write_rejects_bad_shapes_and_preserves_state` (4-member / wrong-typed third / non-OID first / >u32 member / device-qualified frame, every refusal state-preserving), `averaging_reference_write_accepts_the_framed_local_form`.
- Objects: `loop_write_empty_setpoint_frame_clears_reference`, `loop_set_references_read_back` (consolidated), plus the framed/indexed/device/wrong-type arm tests from the first round; `pulse_converter_write_framed_indexed_input_reference_lands` + malformed-preservation test.
- PICS truth: `is_writable_property_matches_write_property_on_pulse_converter_and_averaging` — both directions, including OBJECT_NAME non-acceptance.
- Wire (`reference_writes.rs`, 7 tests): framed writes per arm incl. indexed + wrapped-setpoint; flattened-local write-back lands; rejections preserve state; `setpoint_reference_empty_frame_and_null_both_clear_over_the_wire` (both clear paths); `averaging_object_property_reference_over_the_wire`; `wpm_reference_write_commits_in_order_and_rolls_back_on_failure`.
- Wire (`multi_element_writes.rs`, 8 tests): the trailing-garbage full-consumption matrix; the datetime unlocks.
- Indexed reads: `reference_properties_reject_indexed_read_property` (all five reference-typed properties → PROPERTY_IS_NOT_AN_ARRAY).
- Preservation evidence: the framed EE/Fault/Recipient_List suites run unchanged and green — context-tagged single-element behavior byte-identical.

## Benchmarks

Required? No — decode-loop generalization + arm hardening only; the loop is O(elements) over payloads bounded by APDU size, no benchmarked hot path touched, and no measured before/after claim is made.

## Subagent reviews

- Pre-implementation panel directives: this PR implements the four-item review-round disposition against the first push (`39ed0e9`):
  - **Runtime lane: ready** — flagged (with dataflow) the empty-`BACnetSetpointReference`-frame refusal; fixed in `715803e` and pinned at codec/helper/arm/wire levels (`setpoint_reference_empty_frame_is_the_absent_alternative`, `empty_setpoint_frame_clears_only_on_the_setpoint_arm`, `loop_write_empty_setpoint_frame_clears_reference`, `setpoint_reference_empty_frame_and_null_both_clear_over_the_wire`). Runtime's cheap-pin requests landed: indexed setpoint frame golden vector (`setpoint_reference_golden_vector_indexed`) and the indexed-ReadProperty gate pins (`reference_properties_reject_indexed_read_property`).
  - **Dataflow lane: ready with 2 items** — (1) Pulse Converter PICS incoherence (arm accepted INPUT_REFERENCE, historical default advertised read-only): fixed by exact overrides in `2398902` (+ Averaging in `715803e`) with the both-directions truth pin; (2) the Pulse Converter citation was challenged and adjudicated as CORRECT by orchestrator verification (extract TOC line 288; clause at line 21100; Table 12-27 at line 21172 with the BACnetObjectPropertyReference row) — no doc change beyond what `39ed0e9` already carried.
  - **Adversary lane: NOT READY, 1 blocker (FIXED)** — Averaging `OBJECT_PROPERTY_REFERENCE` arm's silent member drops / member retype / `as u32` truncation: the arm is newly network-reachable via this PR's loop decode, so the PR owns it; fixed in `715803e` (shared strict decode, exact shape, `u32::try_from`, INVALID_DATA_ENCODING on device-qualified members with the Clause 12.5 OPTIONAL-remote-sampling citation), refusal matrix proves state preservation per case; `as u32` search of the reference paths confirms none remain.
- Self-review discipline: no approve/merge actions taken by the implementation lane.

## Known limitations

- Reads of the reference properties still emit the flat application-tagged list form (this stack's existing convention) rather than the context-tagged Clause 21 frame; the encode side of these properties was left wire-minimal in this tranche. A peer writing back our own read emission is handled (the flat list is the legacy local form) but byte-framing interop for reads is future work.
- `SETPOINT_REFERENCE` accepts the bare member sequence in addition to the standard's [0]-wrapped `BACnetSetpointReference` frame (defensive; the two are byte-unambiguous). The wrapped form is the conformant emission shape; the empty frame and Null are both supported clear paths.
- A value payload TLV-truncated far enough desynchronizes the service request's own [3]/[4] framing walk and is refused there (a framing error, SERVICES class at dispatch) rather than by the value decode — refusal + preserved state either way; the layering is pinned in `list_write_with_tlv_truncated_tail_is_refused_before_the_value_decode`.
- AddListElement/RemoveListElement still `break` their `listOfElements` decode loop on an undecodable element instead of erroring — the non-framed (non-Recipient_List) trailing-garbage behavior on those two services is issue #182's wider-scope residue, reported here rather than silently widened into this PR.
- `SETPOINT_REFERENCE`'s BACnetSetpointReference production allows zero references (modeled as clear); the Loop's own `Setpoint` property continues to hold the actual fixed setpoint — no attempt is made to cross-validate.
- The reference arm's shared decode covers the five modeled reference properties; other context-tagged-member properties (e.g. `OBJECT_PROPERTY_REFERENCE` on Event Enrollment, `LIST_OF_OBJECT_PROPERTY_REFERENCES`) keep their pre-existing arm-level behavior — those arms were NOT silently-truncating (EE decodes via its own DOPR list codec; Schedule denies writes), verified while auditing residual `.0` offset-dropping and `items.len() >=` patterns.

## Follow-ups

- #277 (Enumerated >4-octet truncation) — untouched per the scope fence; the loop does not newly truncate.
- Add/RemoveListElement trailing-element erroring (above) — belongs to a wider follow-up with the destination-list element semantics from #152's review.
- Reference-property reads: emit the Clause 21 context-tagged frame (encode-side change), with interop evidence against a second implementation.
- Averaging remote sampling: if ever modeled, widen the stored reference to BACnetDeviceObjectPropertyReference and admit member [3] — the current local-only posture is the standard's OPTIONAL branch consciously not taken.
- Pulse Converter's PRESENT_VALUE arm accepts writes unconditionally (in-service included) — Clause 12.23 requires writability when Out_Of_Service is TRUE but does not forbid in-service writes; advertised as-is. A Stricter in-service gate (analog-input style) would be a separate behavior decision.
- Pre-existing wrong clause citations in `crates/bacnet-objects/src/accumulator.rs`'s header ("Clauses 12.1/12.2" — actually Accumulator 12.61, Pulse Converter 12.23) and `crates/bacnet-objects/src/averaging.rs`'s header ("Clause 12.4" — actually 12.5): reported, not fixed here.

Closes #182


crates/bacnet-objects/src/reference.rs
jscott3201, line 1: Packaging verification: the shared reference decoder and round-2 strictness are present on the reviewed head; no additional change requested.

Issue #229, opened 2026-07-27
Event_Enable, Time_Delay and Notify_Type are unreachable on Binary Input, Binary Value, Multi-state Input and Multi-state Value


jscott3201:
## The defect

The event-configuration properties that commission intrinsic reporting are not writable — and in two cases not even readable — on four of the six binary and multi-state object types.

| Property | Analog AI/AO/AV | BI / BV / MSI / MSV |
|---|---|---|
| `Event_Enable` | read + write | **read only** |
| `Time_Delay` | read + write | **absent entirely** |
| `Notify_Type` | read + write | **absent entirely** |
| `Notification_Class` | read + write | read only (BI/BV) |
| `Acked_Transitions` | read + write | read only (BI/BV) |

The analog types get all of these from the shared macros in `crates/bacnet-objects/src/common.rs`. The binary and multi-state types were never wired to them.

## Why it matters

`Event_Enable` gates *distribution* (see #136). Every detector defaults to `event_enable: 0`, i.e. all three transition bits false. With no write path, the bits are stuck at (F,F,F) permanently, so **no event notification can ever reach the wire from these object types**, regardless of what their event algorithm detects.

Clauses 12.6, 12.8, 12.18 and 12.20 are explicit that this is not a permitted restriction:

> "A device is allowed to restrict the set of supported values for this property but shall support (T, T, T) at a minimum."

The implementation supports exactly (F, F, F).

`Time_Delay` being absent means `pTimeDelay` is permanently 0, so any transition fires on the same write that causes it, with no damping available.

## Prerequisite is already done

PR #227 split the shared macros into generic and analog halves precisely so this could be wired without dragging analog-only properties along:

- `read_generic_event_properties!` / `write_generic_event_properties!` — `Event_State` (read only), `Event_Enable`, `Notify_Type`, `Notification_Class`, `Time_Delay`, `Acked_Transitions`. These touch only detector fields every detector carries.
- `read_analog_event_properties!` / `write_analog_event_properties!` — `High_Limit`, `Low_Limit`, `Deadband`, `Limit_Enable`, plus the analog `Event_Time_Stamps` / `Event_Message_Texts` reads.

PR #227 wired the generic halves into Binary Output and Multi-state Output only. This issue is the same wiring for the remaining four types.

## Scope

- `crates/bacnet-objects/src/binary/mod.rs` — Binary Input, Binary Value
- `crates/bacnet-objects/src/multistate/input.rs`, `multistate/value.rs`
[...]

PR #288, opened 2026-08-12
feat(objects): add Time_Delay_Normal with normal-direction delay selection on intrinsic reporters (#225)

### LINKED ISSUE #225: Time_Delay_Normal is unimplemented: every event algorithm uses one delay in both directions
### Affected revision

Observed on `dev` at `01c32c4`.

### Behavior

ASHRAE 135-2020 gives the event algorithms two independent delays. Clause 13.3.4 (COMMAND_FAILURE) defines both:

> "pTimeDelay — This parameter, of type Unsigned, represents the time, in seconds, that the offnormal conditions must exist before an offnormal event state is indicated."

> "pTimeDelayNormal — This parameter, of type Unsigned, represents the time, in seconds, that the Normal conditions must exist before a NORMAL event state is indicated. If no value is available for this parameter, then it takes on the value of the pTimeDelay parameter."

and applies them asymmetrically:

> "(a) If pCurrentState is NORMAL, and pFeedbackValue is not equal to pMonitoredValue for **pTimeDelay**, then indicate a transition to the OFFNORMAL event state."

> "(b) If pCurrentState is OFFNORMAL, and ... for **pTimeDelayNormal**, then indicate a transition to the NORMAL event state."

None of this exists in the implementation:

- `PropertyIdentifier::TIME_DELAY_NORMAL = 356` is defined in `crates/bacnet-types/src/enums/property_id.rs:370` and referenced **nowhere** — a repo-wide grep for `TIME_DELAY_NORMAL` or `time_delay_normal` returns only that definition.
- All three detectors in `crates/bacnet-objects/src/event.rs` carry a single `pub time_delay: u32` field (`:274`, `:509`, `:645`) and use it for transitions in both directions.

So an object configured with a long offnormal delay and a short return-to-normal delay — or the reverse — behaves as though both were the offnormal value.

### Scope

This is **not** specific to COMMAND_FAILURE. The parameter is defined per event algorithm, and the property appears in the object tables for the types that support intrinsic reporting. Both Clause 12.7 (Binary Output) and Clause 12.19 (Multi-state Output) list it, marked present-only-if-intrinsic-reporting rather than required:

> "These properties shall be present only if the object supports intrinsic reporting."

The other intrinsic-reporting object types should be checked the same way before implementing; the algorithm clauses for OUT_OF_RANGE and CHANGE_OF_STATE need reading to confirm each defines `pTimeDelayNormal`, rather than assuming it from COMMAND_FAILURE.

### Why it is invisible today

Both delays are equal whenever `Time_Delay_Normal` is absent, and it is always absent because the property does not exist in any object. So current behavior is indistinguishable from a conformant implementation that happens to have the property unset. It becomes observable the moment the property is added and given a different value.

Note the standard's fallback — "If no value is available for this parameter, then it takes on the value of the pTimeDelay parameter" — means the current single-delay behavior is the *correct* behavior for the absent case. This is a missing feature, not a wrong computation.

### Suggested direction

- confirm from Clause 13.3 which algorithms define `pTimeDelayNormal`, and from the Clause 12 object tables which object types list the property;
- add `time_delay_normal: Option<u32>` to the detectors that need it, implementing the documented fallback to `time_delay` when `None` rather than duplicating the value at construction — the fallback is normative and should be visible as such;
- add the `Time_Delay_Normal` property to the qualifying object types, and route it through `is_writable_property` so PICS matches dispatch;
- test asymmetric delays in both directions, and test that an absent property still yields symmetric behavior.

### Relevant code

- `crates/bacnet-objects/src/event.rs:274`, `:509`, `:645`
- `crates/bacnet-types/src/enums/property_id.rs:370`
- Related: #222 (which surfaced this while scoping COMMAND_FAILURE), #120 (the original Time_Delay work)

jscott3201:
## Summary

Adds `Time_Delay_Normal` (property identifier 356) — the Clause 13.3 `pTimeDelayNormal` parameter — to the nine intrinsic-reporting object types, and wires the normal-direction delay selection through all three intrinsic event detectors, closing the single-delay-in-both-directions gap from #225. `Closes #225`.

## Normative anchors (verified against the local 135-2020 extract)

- **pTimeDelayNormal definition** (identical text per algorithm): "the time, in seconds, that the Normal conditions must exist before a NORMAL event state is indicated. If no value is available for this parameter, then it takes on the value of the pTimeDelay parameter." — CHANGE_OF_BITSTRING extract 45663-45667, CHANGE_OF_STATE 45746-45750, OUT_OF_RANGE 46195-46200 ("If no value is **provided**" variant), COMMAND_FAILURE quoted in the issue body.
- **Direction rule from the condition letters**: CHANGE_OF_STATE (a) pTimeDelay / (b) pTimeDelayNormal / (c) offnormal→offnormal re-indication pTimeDelay (extract 45752-45759); COMMAND_FAILURE (a)/(b) as quoted in #225; OUT_OF_RANGE (a)/(b) INTO HIGH_LIMIT/LOW_LIMIT pTimeDelay, (d)/(g) offnormal→offnormal re-cross pTimeDelay, (e)/(h) deadband returns to NORMAL pTimeDelayNormal (extract 46202-46219); FAULT carries no delay term (Clause 13.2.2 fault precedence, unchanged).
- **Intrinsic tables**: AI Table 12-2 Time_Delay=O3,5 row at extract 11782/11845, Time_Delay_Normal=O5 at 11801/11860; description text at 12076-12078.

## Conformance-code verification notes (two-column interleave risk)

The extract interleaves the property/datatype column with the conformance-code column across page breaks, so every table was verified by **count-based alignment** (property/datatype pairs == conformance codes) plus footnote-body adjacency, not by visual row matching:

| Type | Table | Alignment | Time_Delay | Time_Delay_Normal |
|---|---|---|---|---|
| AI | 12-2 | 42↔42 | O3,5 (11845) | **O5** (11860) |
| AO | 12-3 | 45↔45 | O2,4 | **O4** |
| AV | 12-4 | 44↔44 | O3,6 | **O6** |
| BI | 12-6 | 37↔37 | O5,7 | **O7** |
| BO | 12-8 | 36↔36 | O4,6 | **O6** |
| BV | 12-10 | 44↔44 | O6,8 | **O8** |
| MSI | 12-21 | 37↔37 | O3,5 | **O5** |
| MSO | 12-22 | read in exact name/type/code triples (19412-19414) | O1,3 | **O3** |
| MSV | 12-23 | 39↔39 | O4,6 | **O6** |

Every second footnote body was confirmed as "These properties shall be present only if the object supports intrinsic reporting" adjacent to its number (extract 11888-11889, 12333-12334, 12819-12820, 13555-13556, 14078-14079, 14650-14651, 19019-19020, 19473, 19939-19940). MSO Table 12-22 footnote 3's numeral was lost to the interleave; it is pinned by cross-checking the other codes (Event_Algorithm_Inhibit=O3,4 matches footnote 4's body, Reliability_Evaluation_Inhibit=O5 matches footnote 5).

**Writability**: no table marks Time_Delay_Normal W — like Time_Delay it is O-coded, so writability is *permitted, not required*. It is made writable mirroring Time_Delay (the #229 rationale: otherwise the Clause 13.3 delay asymmetry is not commissionable at all), validated as Unsigned within the u32 span with the Clause 15.9.1.3 pairings (wrong datatype → PROPERTY/INVALID_DATA_TYPE; unrepresentable → PROPERTY/VALUE_OUT_OF_RANGE; stored value preserved).

## Per-detector behavior

- Direction selection lives once in `event.rs` (`delay_toward`): target NORMAL → `time_delay_normal.unwrap_or(time_delay)` (the normative fallback, visible as `Option`); any other target — including offnormal→offnormal — → `time_delay`.
- `probe()` fires immediately only when the *direction-appropriate* delay is zero; otherwise seeds `PendingTransition` with it (no re-seed for a redundant same-target write, as before).
- `tick()` decrements unchanged; a mid-delay target change re-seeds with the *current* target's delay. Condition-revert cancellation unchanged (revert to the confirmed state leaves nothing pending).
- FAULT: Clause 13.2.2 precedence drops any in-flight countdown and fires immediately — both delays ignored (test-pinned).
- `impl_intrinsic_reporting!` macro: no change — it delegates probe/tick; the delay lives inside the detectors.
- Objects: `Time_Delay_Normal` added to all nine `Property_List`s and to the shared generic read/write macros + `is_generic_event_property_writable` (so all nine types + PICS share the truth source). Read-back of an unwritten property returns the effective (fallback) delay. Reset semantics mirror `Time_Delay`: the EDE=FALSE restoration path touches neither delay; nothing invented.

## Tests

- **Detector-level** (`event/time_delay_normal_tests.rs`, new file — keeps `tests.rs` under the 700-LOC cap): `time_delay_normal_selects_the_return_to_normal_delay` (TD=5 offnormal / TDN=10 NORMAL tick counts), `time_delay_normal_absent_falls_back_to_time_delay`, `time_delay_normal_zero_fires_the_return_to_normal_immediately`, `time_delay_normal_does_not_govern_offnormal_to_offnormal`, `time_delay_normal_mid_delay_reseed_uses_the_current_targets_delay`, `fault_transitions_ignore_both_delays`, `time_delay_normal_change_of_state_asymmetric_round_trip`, `time_delay_normal_command_failure_asymmetric_round_trip`, `time_delay_normal_delayed_return_still_cancels_on_revert`.
- **Object-level**: shared round-trip helpers extended (BI/BV in `binary/tests/generic_event_properties.rs`, MSI/MSV/MSO in `multistate/tests/generic_event_properties.rs`) with TDN write/read/listed/writable + INVALID_DATA_TYPE / VALUE_OUT_OF_RANGE refusals and value preservation; fallback readback tests `binary_/multistate_/analog_time_delay_normal_defaults_to_time_delay_when_unwritten` (all nine types); analog round-trip `analog_time_delay_normal_round_trips_and_matches_pics` (AI/AO/AV); asymmetric gating at object level `ai_time_delay_normal_gates_the_return_to_normal` (OOR) and `bo_time_delay_normal_selects_the_return_to_normal_delay` (CommandFailure); MSO inline loop extended in `multistate/output.rs` 's `generic_event_properties_round_trip_and_match_pics`.
- **Wire-level**: `crates/bacnet-server/src/handlers/tests/write_validation.rs::time_delay_normal_round_trips_over_write_property_and_read_property` — RP fallback readback, WP round-trip, Time_Delay independence, both refusal pairings over WriteProperty/ReadProperty on Analog Input.

## Gates (all green)

- `.github/scripts/check-file-size.sh` — OK (new file `event/time_delay_normal_tests.rs` split out to stay under the cap)
- `cargo clippy --workspace --exclude rusty-bacnet --all-targets --locked` — exit 0 (pre-existing warnings only; none in changed code)
- `cargo test --workspace --exclude rusty-bacnet --exclude bacnet-wasm` — all suites ok (objects 977 incl. 16 new)
- `cargo check -p rusty-bacnet --tests` — clean
- `python3 scripts/generate-conformance-docs.py --check` — clean (JSON + hand ledger synced; `repo_sha` = code commit `53dd962c7a85e2fe8a16b1dc6f1e836d47f864f0`, docs commit after)

## Deviations / limitations

- OUT_OF_RANGE's *immediate* (no-delay) returns when a limit is disabled (13.3.6 (c)/(f)) were never separately modeled by the detector; a LIMIT_ENABLE write disabling an active limit still returns through the deadband path, now gated by pTimeDelayNormal. Pre-existing approximation, flagged in the ledger row; EE-evaluator tranche (C2) territory.
- Event_Time_Stamps / Event_Message_Texts stamping on transitions stays out of scope (#123 family), as does the EE evaluator, DCC gating, and AlertEnrollment (C2-C4).


jscott3201 (commented):
[Luna] READY — merge recommendation: merge PR #288 into `dev`.

The implementation, CI, PR packaging, conformance artifacts, scope, and documented follow-ups were verified. Issue #289 remains a non-blocking C3 data-integrity follow-up for preserving the unset/effective-value distinction during WPM rollback.

Issue #264, opened 2026-07-30
Event Enrollment and Alert Enrollment lack Event_Time_Stamps, which their property tables mark R (required)


jscott3201:
### Affected revision

`dev` at `415eb32`. Found by the Event_Time_Stamps family grounding pass (2026-07-30); conformance codes verified by the family's spec gate (Codex + Opus concur).

### Behavior

Both enrollment objects are required to carry `Event_Time_Stamps` unconditionally — not via the intrinsic-reporting footnote pair the six I/O types use, but as plain `R`:

> Table 12-14 (Event Enrollment, Clause 12.12, p. 231): `Event_Time_Stamps BACnetARRAY[3] of BACnetTimeStamp R`
> Table 12-61 (Alert Enrollment, Clause 12.52, p. 503): `Event_Time_Stamps BACnetARRAY[3] of BACnetTimeStamp R`

Neither object models it: `EventEnrollmentObject` has no field, no read arm, and no `Property_List` entry (`crates/bacnet-objects/src/event_enrollment/mod.rs:368-388` — the list carries EVENT_TYPE, EVENT_STATE, EVENT_ENABLE, ACKED_TRANSITIONS, EVENT_DETECTION_ENABLE, but no EVENT_TIME_STAMPS). `AlertEnrollmentObject`'s property_list (`:528-535`) carries only EVENT_DETECTION_ENABLE and EVENT_ENABLE from that group.

The spec gate also confirmed (Clauses 13.2/13.2.1) that Event Enrollment maintains Event_Time_Stamps and Acked_Transitions by the same Clause 13.2.2.1.4 / 13.2.3 process as intrinsic reporting, with Ack_Required from its referenced Notification Class — and that Alert Enrollment shares the process except its TO_NORMAL Ack_Required bit is ignored with FALSE conveyed (Clause 12.52.8, p. 504).

### Suggested direction

- Add the array to both types using whatever shared storage shape the #235/#230/#258 slice lands, listed in `Property_List`.
- The EE evaluator (`crates/bacnet-server/src/event_enrollment/mod.rs:688-709`) applies transitions with `set_event_state_internal` only — it needs the same transition-maintenance hook #123 adds for the intrinsic paths.
- Respect the Alert Enrollment TO_NORMAL exception when wiring Ack_Required.

### Relevant code

- `crates/bacnet-objects/src/event_enrollment/mod.rs`
- `crates/bacnet-server/src/event_enrollment/mod.rs`

### Related

[...]

Issue #179, opened 2026-07-27
Event Enrollment Status_Flags never reflects Event_State in IN_ALARM, and can set OUT_OF_SERVICE which 12.12 requires to be FALSE


jscott3201:
### Affected revision

Observed on `dev` at commit `3990f3e`.

### Behavior

`EventEnrollmentObject::read_property` calls `read_common_properties!(self, property, array_index)` first and returns its result (`crates/bacnet-objects/src/event_enrollment/mod.rs:119-121`), so `STATUS_FLAGS` is answered by the shared macro and never reaches an object-specific arm. The macro passes a hard-coded `0` for `event_state` (`crates/bacnet-objects/src/common.rs:118-128`); its own comment says objects with event detection should handle `STATUS_FLAGS` before delegating.

`compute_status_flags` (`crates/bacnet-objects/src/common.rs:13-39`) therefore always takes the else branch at `:22-24` and clears IN_ALARM. This is reachable, not theoretical: the server evaluator writes a non-NORMAL `Event_State` through `set_event_state_internal` (`crates/bacnet-server/src/event_enrollment/mod.rs:658`), which the object stores verbatim (`crates/bacnet-objects/src/event_enrollment/mod.rs:252-255`). An enrollment sitting in HIGH_LIMIT reports IN_ALARM FALSE.

Separately, `compute_status_flags:30-34` sets OUT_OF_SERVICE from `self.out_of_service`, which on an Event Enrollment is network-writable (`crates/bacnet-objects/src/event_enrollment/mod.rs:235-239`).

Clause 12.12 Status_Flags (p239):

- "IN_ALARM  Logical FALSE (0) if the Event_State property has a value of NORMAL, otherwise logical TRUE (1)."
- "OVERRIDDEN  The value of this flag shall be logical FALSE (0)."
- "OUT_OF_SERVICE The value of this flag shall be logical FALSE (0)."

Both flag requirements are unconditional for this object type. `Status_Flags` is conformance R in Table 12-14 (p234) and directly readable over the wire, so both deviations are observable. Table 12-14 lists no `Out_Of_Service` property for Event Enrollment at all, so the writable flag is itself nonstandard — see #164, which removes it.

OVERRIDDEN is FALSE today only by accident: `status_flags` is initialised to `StatusFlags::empty()` (`:60`) and has no writer, so the required value happens to hold.

### Suggested direction

[...]

Issue #130, opened 2026-07-14
Separate internal Event Enrollment transitions from network-writable Event_State


jscott3201:
### Affected revision

Observed on rusty-bacnet v0.10.1 at commit `8b8c0cba3e92caa30e145c91abedfa003c64f2c9`.

### Behavior

The Event Enrollment evaluator mutates `Event_State` by calling the same `write_property` branch exposed to network requests. That branch also accepts direct `Event_State` writes.

Internal lifecycle state and public property mutation therefore share an access path without a distinct invariant or authorization boundary.

### Suggested direction

- add an internal transition-state method;
- define the network write policy for `Event_State`;
- validate values at both boundaries;
- test that periodic evaluation does not depend on a public write route.

### Relevant code

- `crates/bacnet-objects/src/event_enrollment/mod.rs`
- `crates/bacnet-server/src/event_enrollment/mod.rs`

PR #290, opened 2026-08-13
fix(server,objects): Event Enrollment delay honoring, same-state transition actions, COV baseline (#163, #166, #137)

### LINKED ISSUE #137: Track a baseline for Event Enrollment change-of-value evaluation

### LINKED ISSUE #163: Honor Time_Delay in the Event Enrollment evaluator
### Affected revision

Observed on `dev` at commit `4ee4d6a`.

### Behavior

`BACnetEventParameter` carries a per-algorithm `time_delay` field for all five evaluated algorithms (added in #129), but `crates/bacnet-server/src/event_enrollment/mod.rs` never reads it — `grep -n time_delay` over that file returns no matches. Every algorithm branch evaluates statelessly, so a condition first observed by an evaluation pass transitions immediately even when the enrollment configures a nonzero delay.

Clause 13.3.2 (p649) defines `pTimeDelay` as "the time, in seconds, that the offnormal conditions must exist before an offnormal event state is indicated". Equivalent requirements appear in 13.3.1, 13.3.3, 13.3.5, and 13.3.6.

The evaluation interval is not a substitute. #133 makes the cadence configurable, but cadence and `Time_Delay` are independent quantities — a 10-second poll neither implements nor approximates a 60-second delay.

Every existing evaluator test sets `time_delay: 0`, so this is untested rather than covered.

#120 fixed the equivalent defect on the *intrinsic* reporting path via a probe/tick split on the per-object detectors. The Event Enrollment (algorithmic) evaluator was outside that change and still needs the same treatment.

### Suggested direction

- carry a per-enrollment pending-transition countdown, as the intrinsic detectors do;
- decrement it on the evaluation tick and fire only at zero;
- cancel a pending transition when the condition clears before the delay elapses;
- do not let a redundant qualifying observation re-seed or shorten an in-flight countdown;
- test nonzero delays across offnormal, fault, and normal transitions, plus mid-delay recovery.

### Relevant code

- `crates/bacnet-types/src/constructed/event_parameter/mod.rs`
- `crates/bacnet-objects/src/event.rs` (the #120 probe/tick precedent)

### LINKED ISSUE #166: Event Enrollment evaluator discards same-state transitions that Clause 13.2.2.1.4 requires to be acted on

jscott3201:
## Summary

The Event Enrollment evaluator is brought to Clause 13 conformance on three fronts: it honors `Time_Delay`/`Time_Delay_Normal` instead of discarding every arm's `time_delay` (#163), it executes Clause 13.2.2.1.4's transition actions for *indicated* transitions whether or not they change the event state (#166), and CHANGE_OF_VALUE tracks the Clause 13.3.3 detection baseline instead of comparing absolute magnitudes (#137). `Closes #163`. `Closes #166`. `Closes #137`.

The evaluation core moves from `mod.rs` to the new `algorithms.rs` (700-LOC cap) behind an explicit `Option<Indication>` spine: the pre-#166 code folded "no condition true" and "indicated, same state" into one `EventState` return, which made the same-state drop unfixable without a baseline; both are now distinguishable, per the Clause 13.3 introduction ("If no condition evaluates to true, then no transition shall be indicated").

## Subagent reviews (record)

The panel's second pass on this PR returned three blocking findings plus agreed fixes; all were reproduced, fixed, and re-verified. Both re-verification lanes then returned READY with non-blocking notes; the absorbed micro-round (`108ae4f`) fixed the row-notes' stale per-pass wording (now seconds + ceil-per-interval) and added the FLOATING_LIMIT-high-limit and COS-foreign-with-matching-alarm recovery pins:

| # | Finding | Fix commit(s) |
|---|---|---|
| B1 (runtime + adversary) | Countdown counted evaluation PASSES, but Clause 13.3's delays are SECONDS — Time_Delay=5 at the default 10s interval gated ~50s. | `7b42d39` — `evaluate_event_enrollments(db, interval_secs)`; countdown seeded `ceil(delay_secs / interval_secs)` (never-fire-early), saturated at u32::MAX; residual behavior documented (interval is builder-config, pending is in-memory → no runtime rescale; restart re-evaluates with a fresh conversion). |
| B2 (adversary) | Foreign `Event_State` (e.g. HIGH_LIMIT left after `Event_Parameters` were rewritten to CHANGE_OF_STATE) wedged the arms — no condition names a state outside the algorithm's set, so the ghost state persisted forever (a regression vs. the stateless base evaluators). | `9dc4b86` — per-arm reachable sets; outside the set the arm evaluates as from NORMAL and indicates the computed state through the ordinary actions path incl. direction-selected delay gating. |
| B3 (dataflow) | `AcknowledgeAlarm` on an EE could never succeed (trait default `OPTIONAL_FUNCTIONALITY_NOT_SUPPORTED`), so the 13.2.3 ack-owed bit could never be acknowledged. | `1c90c90` — `acknowledge_alarm` override: unconditional idempotent set per 13.2.3; detection-disabled EE refuses OBJECT/NO_ALARM_CONFIGURED (Table 13-10, "object exists but does not support or is not configured for event generation"), which also preserves Clause 12.12's disabled-state `Acked_Transitions` invariant. OOS does not gate ack (no clause bars it). |
| F4 (adversary N1 / dataflow F2) | Fingerprint-mismatch cancellation was lost across early exits — params round-trip A→B→A resumed a stale countdown. | `7b42d39` — cancel is flushed to the object as soon as the mismatch is observed, before any later exit; the unreadable-reference exit deliberately precedes the check (transient unreachability retains the countdown). |
| F5 (adversary N2) | Retargets invisible to the fingerprint. | `7b42d39` — monitored object + property folded into `params_fingerprint` (the FLOATING_LIMIT setpoint reference was already covered: it serializes inside the framed `Event_Parameters`). |
| F2 (runtime N2) | Structured COBS matcher truncated to `min(mask, alarm, value)` — false OFFNORMAL when the mask is wider than the monitored bitstring (alarm bits beyond the value's width were never examined). | `9dc4b86` — compare over `max(mask, value)` with zero-filled value bytes (matches `masked_value_hash` and the legacy evaluator). |
| F7 (runtime N3) | `Indication.condition` doc conflated 13.3.2 (a)'s "equal to ANY of the values … for pTimeDelay" with (c)'s "remains equal to THAT value". | `9dc4b86` — comments now anchor the strictness on (c) and call (a)'s application the deliberate stricter-than-required choice. |
| F8 (dataflow F4) | `addenda_errata_status` in the JSON ledger still described the #225 tranche. | `6d591a6` (+ `81faf70` clippy drop) — refreshed, summaries regenerated, hand-ledger synced. |

Runtime residual gaps made cheap while in the file: TO_NORMAL-direction `Ack_Required` coverage and the `Event_Enable`-vs-`Acked_Transitions` independence pin (`same_state.rs`). (Pending-across-read-failure is exactly F4's regression test.)

## Normative anchors (verified against the local 135-2020 extract)

- **Table 12-14 row**: `Time_Delay_Normal` — Unsigned, conformance **O** (extract 16444-16446, read in name/type/conformance triples adjacent to `Event_Algorithm_Inhibit` O1 and `Status_Flags` R because of the extract's two-column page-break interleave); property text "the pTimeDelayNormal parameter for the object's event algorithm" (16887-16889). No table marks it W → writability is the Clause 12.1.2 *permitted* option, exercised mirroring the intrinsic types (the #229 rationale: otherwise the asymmetry is uncommissionable).
- **Table 12-15 mapping** (16530-16679): every evaluated algorithm (CHANGE_OF_BITSTRING / CHANGE_OF_STATE / CHANGE_OF_VALUE / FLOATING_LIMIT / OUT_OF_RANGE) maps `Time_Delay` → pTimeDelay; the EE object has no `Time_Delay` property, so pTimeDelay comes from `Event_Parameters`. `Time_Delay_Normal` is NOT in `Event_Parameters` — it is the object property.
- **13.2.2.1.4** (45051-45062): actions "shall be executed even if the transition does not change the event state"; first action stores the SPECIFIC state ("it is not acceptable to set Event_State to OFFNORMAL when the returned value is HIGH_LIMIT", 45055-45057).
- **13.2.3** (45140-45141): "If the corresponding bit in Ack_Required is set, then the bit in Acked_Transitions is cleared, otherwise it is set." `Ack_Required` is read from the referenced Notification Class object; an unresolvable class reads as not-required (the clause's own "otherwise it is set" half).
- **Direction rule**: CHANGE_OF_STATE (a)/(c) pTimeDelay vs (b) pTimeDelayNormal (45749-45758); OUT_OF_RANGE (a)/(b)/(d)/(g) pTimeDelay vs (e)/(h) pTimeDelayNormal (46195-46215); pTimeDelayNormal fallback text (45663-45667). Delay units are SECONDS ("the time, in seconds, that the offnormal conditions must exist", e.g. 45663-45667's sibling pTimeDelay texts).
- **13.3.3 CHANGE_OF_VALUE** (45797-45885): baseline = "the value of the monitored value when a transition to NORMAL is indicated"; "initialization … before the first transition to NORMAL is indicated is a local matter"; REAL condition (a) targets NORMAL gated by pTimeDelayNormal (Figure 13-10 admits no offnormal transition).
- **Clause 13.9 / Table 13-10** (48157-48300): AcknowledgeAlarm's Result(-) mapping — `OBJECT/NO_ALARM_CONFIGURED` for an object "not configured for event generation".

## Per-arm behavior (post-review state)

| Algorithm | Delay applied (seconds, ceil vs. interval) | Same-state outcome | Foreign-state recovery |
|---|---|---|---|
| OUT_OF_RANGE / FLOATING_LIMIT | TD toward HIGH/LOW_LIMIT; TDN(→TD fallback) toward NORMAL | None defined (13.3.6 (a)–(h) all state-changing): persistence = silence — pinned | normalized to NORMAL's perspective; computed state indicated (recovery), delay-gated |
| CHANGE_OF_STATE | TD: (a) NORMAL→OFFNORMAL, (c) OFFNORMAL→OFFNORMAL; TDN(→TD): (b) →NORMAL | (c) **implemented** (Optional: in 13.3.2); pending identity per matched value (driven by (c)'s "THAT value"; (a)'s "ANY" gets the stricter reading deliberately) | matched → OFFNORMAL indicated; else NORMAL indicated |
| CHANGE_OF_BITSTRING | TD: (a); TDN(→TD): (b); full-width zero-padded compare | (c) deliberately **not** implemented (Optional:, no bitstring baseline retained) — documented | same shape as COS |
| CHANGE_OF_VALUE | TDN(→TD fallback) only — 13.3.3's conditions reference pTimeDelayNormal exclusively | Only possible transition: NORMAL→NORMAL on `|current − baseline| ≥ pIncrement` (positive REAL only) / significant masked-bit change; baseline advances at each indicated transition | indicates NORMAL + installs the sample as baseline on fire |
| Legacy `Opaque` 0xFF | 0 (no delay slot in the LE layouts) — behavior byte-identical to before | state-diff mapping, unchanged | via the shared byte evaluators' normalization |

## Pending-ownership decision

The pending countdown, the COV baseline, and the COS last-offnormal-causing value live **in the EE object** (in-memory only — never network-readable), behind new guarded `BACnetObject` internal methods (`enrollment_eval_state_internal` / `set_enrollment_eval_state_internal` / `set_acked_transitions_internal`, each defaulting to fail-closed), mirroring the `set_event_state_internal` precedent (#130). The Clause 13.2.2.1 detection-disabled reset clears all three; restart semantics match the intrinsic detectors. The EE pending type deliberately does **not** reuse `bacnet_objects::event::PendingTransition` — structurally similar mechanics, no shared code across the objects/server boundary — and the countdown's tick is one `event_enrollment_task` pass while its *duration* is wall-clock seconds via `ceil(delay_secs / interval_secs)`. Parameter-change-mid-pending (incl. monitored-reference change): fingerprint mismatch cancels and re-gates; the cancellation is persisted immediately.

## First-sample policy decision (#137)

The first sample observed initializes the COV baseline and **never indicates a transition** — the clause's explicit "local matter" (a zero-initialized baseline would false-fire on any opening value larger than the increment). A criterion-type change leaving the stored baseline incomparable re-establishes under the same policy. Consequence recorded in the changelog/ledger: two pre-existing tests pinned the removed absolute-magnitude behavior and were rewritten. Custom downstream EE-typed objects skipping the new internal channel degrade fail-closed (TD=0 transitions fire as before; nonzero delays and COV inert) — documented in the trait rustdoc.

## Fences honored

- No notification sending — the lifecycle task still only logs (`#127`, tranche E).
- No `Event_Time_Stamps` / `Event_Message_Texts` on the EE object (tranche E/#264); the remaining 13.2.2.1.4 actions are named in code comments.
- No `Status_Flags` algorithm input (Table 12-15.1): conditions referencing pStatusFlags take the spec's all-FALSE absent-parameter fallback (tranche-E follow-up).
- No intrinsic-detector changes; no DCC work; the legacy `Opaque` paths' behaviors are unchanged.
- The intrinsic detectors' missing `Acked_Transitions` maintenance stays tracked under #123.

## Known limitations (also in the ledger row)

- **#179 — EE `Status_Flags` IN_ALARM never set**: explicitly NOT this tranche (tranche E); recorded here so the panel record is honest.
- OUT_OF_RANGE 13.3.6 (c)/(f) no-delay returns on a limit-disabled edge: inherited from `BACNET-12-TIME-DELAY-NORMAL`'s limitation; unchanged.
- EE fault precedence (no Reliability term in the evaluator): unchanged, out of scope.
- CHANGE_OF_BITSTRING condition (c): deliberately unimplemented (Optional:).
- `TimeStamp` matching in AcknowledgeAlarm (Table 13-10's INVALID_TIMESTAMP situation) is unenforceable until `Event_Time_Stamps` exist (#264); a stale-timestamp ack is currently accepted (13.2.3's set is unconditional and idempotent).
- Retargeting `Object_Property_Reference` mid-run has no network/property route (Table 12-14 codes it R); the retarget-regate mechanism is proven through the internal channel.

## Tests

- `event_enrollment/tests/delays.rs` — TD gating at seeded cadence (kills restart-mutants), TDN-vs-TD asymmetry, absent-TDN fallback, cancel-on-revert, target-change re-seed, param-change regates + persisted-cancel (A→B→A), `passes_for_delay` ceiling/saturation pins, interval=10 conversion through the evaluator, fingerprint-covers-retarget (+transplant-driven behavioral regate).
- `event_enrollment/tests/same_state.rs` — COS (c) emission/no-refire/delayed/value-discriminated; OOR no-same-state pin + across-band specific-state storage; `Acked_Transitions` clear/set incl. no-NC fallback, **TO_NORMAL-direction ack-required**, and **`Event_Enable`-independence**.
- `event_enrollment/tests/change_of_value.rs` + `compat.rs` — first-sample policy, threshold crossing NORMAL→NORMAL, baseline advancement, repeated crossings, non-positive increment, delay gating, bitmask significance.
- `event_enrollment/tests/foreign_state.rs` — per-arm wedge recovery (HIGH_LIMIT under COS gated by TDN fallback; OFFNORMAL under OOR; OFFNORMAL under COV installing the baseline; HIGH_LIMIT under COBS; COBS width pin; OOS gate precedes recovery).
- `server/event_enrollment_task_tests.rs` — `start_paused` lifecycle: TD=2 at 1s interval; **at the DEFAULT 10s interval** TD=5 fires at the t≈10s tick, TD=15 at ≈20s, TD=25 at ≈30s, TD=0 on the immediate first tick; nothing distributed (#127).
- `handlers/tests/acknowledge_alarm_ee.rs` — wire-level: fired ack-required transition → GEI shows the owed bit cleared → AcknowledgeAlarm succeeds → bit set → GEI confirms; idempotent duplicate; TO_NORMAL path; detection-disabled → OBJECT/NO_ALARM_CONFIGURED with `Acked_Transitions` untouched.
- `objects/…/event_enrollment/tests/time_delay_normal.rs` — TDN read/write/validation (15.9.1.3 pairings)/`Property_List`/PICS, eval-state channel round-trip, disable-reset clearing + fail-closed guards, ack-channel bit maintenance, trait-default rejections.
- Manually killed mutants: swapped delay directions (3 tests), always-reseed countdown (11 tests).

## Verification (real gate runs, not wrapper exit codes)

`check-file-size.sh` OK; `cargo fmt --all --check` clean; `cargo clippy --workspace --exclude rusty-bacnet --all-targets --locked` exit 0; `cargo test --workspace --exclude rusty-bacnet --exclude bacnet-wasm` exit 0 (**2666 tests passed, 0 failed**); `cargo check -p rusty-bacnet --tests` exit 0; `python3 scripts/generate-conformance-docs.py --check` exit 0. Conformance ledger: `BACNET-13-EVENT-ENROLLMENT-EVALUATOR` (supported-with-clause-evidence) with extract line anchors + review-round notes; `repo_sha` = 8680203 (last commit of the initial tranche; review-round commits extend it per the branch-tip convention).


jscott3201 (commented):
[Luna] READY. HEAD 108ae4f is package-complete for merge: the three reviewed blockers have concrete fixes and regression tests; all triggered CI checks are green; the PR body records both review rounds, all three closing directives, scope fences, and honest limitations; the JSON, hand ledger, and generated support summary agree at repo_sha 8680203 without widening public claims; the changelog attributes the rewritten pre-existing COV assertions to #137 semantics; and the remaining tranche-B/tranche-E gaps are explicitly tracked rather than silently dropped. No line findings.

jscott3201:
[Luna] Package verification at HEAD `108ae4f11af0ca61f3e53b9f36491e8b32979f5f`: READY. The three reviewed blockers are fixed and pinned (`delays.rs:436` `passes_for_delay_ceiling_semantics` plus `server/event_enrollment_task_tests.rs:218/240/259/278` default-10s wall-clock tests; `foreign_state.rs:372` `foreign_high_limit_recovers_under_floating_limit_params` and `:445` `foreign_high_limit_with_matching_alarm_indicates_offnormal`, alongside the foreign-state suite; `handlers/tests/acknowledge_alarm_ee.rs:87` `ee_acknowledge_alarm_round_trip_over_services` plus the disabled-refusal and TO_NORMAL wire tests). All triggered CI checks are green: Test (ubuntu-latest), Clippy, Rustfmt, file-size cap, and no-secret scan; release/heavy jobs are correctly skipped for a feature→dev PR. The PR body is template-complete in substance, records both review rounds and the fixes, has `Closes #163`, `Closes #166`, and `Closes #137`, and documents #264/INVALID_TIMESTAMP and #179 as limitations. Conformance JSON, hand ledger, and generated support summary are synced at `repo_sha` `8680203e8cb986afb58d0ad53b84eb75aec0959a`; `addenda_errata_status` is refreshed by `6d591a6`, notes are corrected by `108ae4f`, and no public claim is widened. CHANGELOG is truthful and explicitly attributes the two rewritten pre-existing COV assertions to #137 semantics. Scope is limited to the requested EE tranche plus its evidence; `codex-overnight-status.md` is absent from the branch (local ignored workspace copy only). Remaining paused work is explicit/tracked: INVALID_EVENT_STATE validation remains tranche B (#132), and Table 12-15.1/EE Status_Flags fetching remains tranche E (#179); neither is silently dropped. Merge recommendation: merge PR #290 after the normal branch protection gate; no code findings.

Issue #291, opened 2026-08-13
Alert Enrollment model diverges from Table 12-61 beyond Event_Time_Stamps


jscott3201:
### Affected revision

`dev` at `daab3a8`, confirmed while auditing C3 (`#205`, `#209`, `#289`) against ASHRAE 135-2020 Clause 12.52 / Table 12-61.

### Gaps

1. `Present_Value` has the wrong datatype and meaning. Table 12-61 defines a read-only `BACnetObjectIdentifier` identifying the object that last provided an alert. `AlertEnrollmentObject` stores `pub present_value: u32` as an AlertState and returns `PropertyValue::Enumerated`.
2. `Notify_Type` is unconditionally required (`R`) but has no field, read/write arm, or `Property_List` entry.
3. `required_properties()` is not overridden. RPM `REQUIRED` and generated PICS therefore classify every type-specific Table 12-61 property as optional. A partial override would still be false while required `Notify_Type` and `Event_Time_Stamps` are absent.
4. `Status_Flags`, `Out_Of_Service`, and `Reliability` are exposed even though Table 12-61 does not list them. Whether to remove them is an API-compatibility decision.

`Event_Time_Stamps` is already tracked by #264 and should not be duplicated here.

### Impact

The `Present_Value` defect is wire-visible and prevents clients from identifying the alert source. Missing `Notify_Type` leaves the object below its required property model. PICS and RPM `REQUIRED` under-report required properties.

### Suggested direction

- Retype `Present_Value` to `ObjectIdentifier`, define its initial/local behavior, and preserve source compatibility only if there is a concrete consumer requirement.
- Implement and validate `Notify_Type` using the existing `NotifyType::ALL_NAMED` pattern.
- Land #264 for Alert `Event_Time_Stamps`.
- Then override `required_properties()` with the complete Table 12-61 R set and add exact RPM/PICS tests.
- Make an explicit compatibility decision for the three properties absent from Table 12-61.

### Relevant code

- `crates/bacnet-objects/src/event_enrollment/mod.rs`
- `crates/bacnet-objects/src/event_enrollment/tests/alert.rs`
- `crates/bacnet-objects/src/traits.rs`
- `crates/bacnet-server/src/pics/mod.rs`
- `crates/bacnet-server/src/handlers/read_property.rs`

### Standard source

ASHRAE 135-2020 Clause 12.52, Table 12-61 (standard pp. 503-505).

Issue #220, opened 2026-07-27
DeviceCommunicationControl halts event-state-detection, though storing Event_State is a transition action rather than an act of communication


jscott3201:
### Affected revision

Pre-existing on `dev` at `118d0db`. Untouched by #167, which merely makes the consequence observable in a new way.

### Behavior

While DeviceCommunicationControl is in effect, the server continues to execute confirmed requests but stops running event-state-detection entirely.

- The request drop gate fires only on `state == 1` (`crates/bacnet-server/src/handlers/requests/mod.rs:37-47`), so under `DisableInitiation` (`state == 2`) a `WriteProperty` is still executed.
- Both detection routes bail on `>= 1` — `crates/bacnet-server/src/server/event_notifications.rs:21-23` (the per-write path) and `crates/bacnet-server/src/server/lifecycle.rs:493-495` (the 1-second intrinsic task).

So a client can write `Reliability` or drive a value out of range, the write lands, and no `Event_State` transition is ever computed. `Status_Flags` then reports FAULT TRUE with IN_ALARM FALSE for the duration, and on restore the object's `Event_State`, `Event_Time_Stamps` and `Acked_Transitions` reflect none of what happened in between.

### Why this looks wrong

ASHRAE 135-2020 Clause 13.2.2.1.4 lists the transition actions: storing the new `Event_State`, storing the timestamp in `Event_Time_Stamps`, storing message text, and indicating the transition to the alarm-acknowledgment and notification-distribution processes. Only the last of those is an act of communication. DCC governs whether the device *communicates*, so suppressing the first three conflates "do not transmit" with "do not detect" — the same distinction #136 drew for `Event_Enable`, where distribution is gated but detection and the Clause 13.2.2.1.4 actions are not.

The natural shape is the one `TransitionOutcome { change, distribute }` already provides: let detection run and the transition actions happen, and gate only distribution on the DCC state.

### Not a regression

Before #167 the FAULT/IN_ALARM inconsistency held unconditionally in every DCC state, because `Reliability` never reached `Event_State` at all. After it, the inconsistency survives only while DCC is active. The change strictly narrows the problem; this issue is about closing the remainder.

There is also a documentation residue to fix alongside it: the comment at `crates/bacnet-server/src/server/lifecycle.rs:468-477` says a written `Reliability` "reaches detection through this tick within a second" without noting that DCC suspends the tick's effect.

### Suggested direction

[...]

PR #292, opened 2026-08-13
fix(server,objects): preserve event state across WPM rollback (#205, #209, #289)

### LINKED ISSUE #205: AlertEnrollment ignores the Clause 13.2.2.1 reset when Event_Detection_Enable is written FALSE

### LINKED ISSUE #209: WritePropertyMultiple rollback does not restore Event_State cleared by an Event_Detection_Enable write
## Summary

A `WritePropertyMultiple` that writes `Event_Detection_Enable = FALSE` and then
fails on a later write leaves the object's `Event_State` cleared, even though the
operation reports failure. The rollback restores `Event_Detection_Enable` but not
the state the write reset as a side effect.

This does **not** break the Clause 13.2.2.1 invariant — the post-rollback state
(`Event_Detection_Enable` TRUE, `Event_State` NORMAL) is internally consistent.
It is an atomicity deviation, not a correctness one.

## Evidence

Reproduced directly against `EventEnrollmentObject`, simulating the WPM
snapshot/write/rollback sequence:

```
ATOMICITY after-rollback Event_State = Enumerated(0)   (pre-WPM was HIGH_LIMIT)
ATOMICITY detection = Boolean(true)
```

Sequence:

1. Enrollment sits in HIGH_LIMIT with `Event_Detection_Enable` TRUE.
2. WPM writes `Event_Detection_Enable = FALSE`. Per Clause 13.2.2.1 this resets
   `Event_State` to NORMAL (and, once #123 lands, `Acked_Transitions` to its
   initial condition).
3. A later write in the same WPM fails.
4. Rollback restores `Event_Detection_Enable = TRUE` from its snapshot.
5. `Event_State` is left at NORMAL. The alarm state is gone.

## Root cause

`handle_write_property_multiple`
(`crates/bacnet-server/src/handlers/write_property.rs:95-150`) snapshots only the
property being written:

```rust
object.read_property(*prop_id, *array_index)
    .ok()
    .map(|v| (*prop_id, *array_index, v))
```

That is correct for properties whose write affects only themselves. It does not
cover a property whose write has a mandated side effect on *other* properties —
which `Event_Detection_Enable` now does. The commandable `PRESENT_VALUE` case
already needed special handling for a related reason (snapshot the priority-array
slot, not the resolved value), so this is the second instance of the same shape.

ASHRAE 135-2020 §19.1.2 is the atomicity requirement the existing rollback code
cites.

## Impact: narrow and self-healing

The window requires a WPM that writes this property FALSE *and* fails afterwards.
The state is re-derived by the periodic evaluator on its next pass (default 10 s),
because with detection re-enabled and the monitored value still out of range the
algorithm computes the alarm state afresh. So the observable damage is a
transient loss of `Event_State` plus a spurious later transition, not a permanent
wrong value.

Once #123 implements `Acked_Transitions` maintenance, the same window also
silently clears acknowledgment state, which is less benign — an operator's
unacknowledged alarm would be marked acknowledged.

## Two possible fixes

1. **Extend the snapshot.** Have the rollback capture `Event_State` (and later
   `Acked_Transitions`) whenever the write targets `Event_Detection_Enable`.
   Localized, but it grows the special-case list in the WPM handler, which is
   already carrying one.

2. **Stop storing the disabled condition; compute it at read time.** Leave
   `Event_State` alone on the write, and have `read_property(EVENT_STATE)` return
   NORMAL whenever `Event_Detection_Enable` is FALSE (same for
   `Acked_Transitions`). Nothing is mutated, so there is nothing to roll back and
   the gap disappears.

Option 2 is the better shape and matches the standard's own phrasing more
closely: Clause 13.2.2 states the disabled case as a standing condition
("Event_State **remains** NORMAL"), not as an action performed at a moment. It
would also resolve #205 — where `AlertEnrollmentObject`'s `event_detection_enable`
is a `pub` field, so there is no write path to hook a reset onto at all — and
would make the guards added in #164/#207 unnecessary rather than merely correct.

Worth deciding 1 vs 2 once, and applying the same shape to both objects.

Found by probing the WPM rollback path while reviewing PR #207.

### LINKED ISSUE #289: WPM rollback replays effective (fallback) read-back of Time_Delay_Normal as a stored write, severing the mandated TD-coupling
Found by dataflow review of PR #288 (bounded severity; deliberately not blocking that PR).

Time_Delay_Normal is the crate's first Option-backed property whose read arm substitutes the effective value (stored, else Time_Delay — the mandatory 13.3.x parameter fallback). The generic WPM rollback snapshot (`write_property.rs:180-187`) reads the property and replays it through the write arm, so a rolled-back multi-write converts unset → `Some(TD)`: the failed request silently changes stored state, and a later Time_Delay write no longer propagates per the fallback rule.

Fix candidates: snapshot preserves unset-ness (internal restore path writing the Option), or a snapshot special-case for fallback-computed properties. Same family as #209 (snapshot/derived-state consistency) — natural fit for tranche C3.

jscott3201:
## Summary

Alert Enrollment now exposes `Event_State` and `Acked_Transitions` and keeps their Clause 13.2.2.1 initial conditions while `Event_Detection_Enable` is FALSE. This closes the modeled reset gap without claiming a complete Table 12-61 implementation.

WritePropertyMultiple now uses object-owned rollback tokens where property readback cannot reconstruct state. Detection-disable writes restore private detector/evaluator state and event history, and an unset `Time_Delay_Normal` restores as unset rather than persisting its effective fallback. Tokens supplement readable property snapshots. Rollback failures are returned, and affected objects still run event/COV reconciliation. This all-or-nothing behavior is repository policy; Clause 15.10 permits preceding writes to remain applied.

Closes #205. Closes #209. Closes #289.

## Behavior

- Alert Enrollment projects NORMAL `Event_State`, all-acknowledged `Acked_Transitions`, and corresponding `Status_Flags` while disabled.
- Alert internal event-state and acknowledgment updates fail while disabled.
- Intrinsic objects, Event Enrollment, and Alert Enrollment restore detection state after a later WPM write fails.
- Duplicate-name failures unwind already-applied writes.
- Out-of-service rollback preserves evaluated and simulated Reliability state without masking the original write error.
- Token-backed `Object_Name` rollback resynchronizes the database name index.
- Unreadable writes and restoration failures report rollback failure and identify residual objects for event/COV reconciliation.

## Review Round

The first independent panel found three actionable rollback-contract gaps. Commit `98e8170` fixes them:

| Finding | Resolution |
|---|---|
| A rollback token replaced the readable property snapshot | Tokens now supplement property snapshots; property replay runs before object-state restoration |
| Token-backed `Object_Name` restoration could leave the name index stale | The name index is refreshed after the complete write rollback |
| A restoration failure could leave residual state without event/COV reconciliation | The handler returns affected object IDs internally and production processes them even when WPM returns an error |

The panel also flagged direct public `event_detection_enable = false; true` assignment on Alert Enrollment. The field remains public for source compatibility; its rustdoc and changelog state that callers must use `set_event_detection_enable` for re-enable reset. Network writes, setter writes, reads while disabled, and internal lifecycle paths enforce the invariant.

## Scope

Alert Enrollment support remains partial. #264 tracks `Event_Time_Stamps`; #291 tracks the `Present_Value` datatype, `Notify_Type`, required-property/PICS metadata, and extra-property decisions. No complete Alert Enrollment or PICS claim is made. DCC behavior remains out of scope (#220).

## Verification

- `cargo test -p bacnet-server handlers::tests::wpm_ --locked`
- `cargo check -p bacnet-server --tests --locked`
- `cargo fmt --all --check`
- `git diff --check`


crates/bacnet-objects/src/event_enrollment/mod.rs
```diff
@@ -672,7 +738,16 @@ pub struct AlertEnrollmentObject {
...
     pub present_value: u32,
     /// Whether event detection is enabled.
+    ///
+    /// Prefer [`Self::set_event_detection_enable`] so disabling also clears
+    /// stored event state. This field remains public for compatibility;
+    /// property reads and internal transition hooks still enforce the
+    /// disabled-state invariant after a direct assignment. Re-enable through
+    /// the setter as well: a direct FALSE-to-TRUE assignment cannot run the
+    /// reset and may expose state stored before the direct disable.
     pub event_detection_enable: bool,
```
jscott3201, line 749: **High – the public flag still lets a disable/re-enable cycle resurrect pre-disable alarm state.**

Concrete sequence: create an Alert Enrollment, set its internal state to `OFFNORMAL` and clear an ack bit; assign this public field `false`, then assign it directly back to `true` (both assignments remain valid source-compatible API use). No setter runs on either assignment, so `event_state`/`acked_transitions` retain their old values. With the flag true, `read_property(EVENT_STATE)` at lines 835-841 returns the stale `OFFNORMAL`; the summarization helper also considers the object enabled, so it can reappear as an alarm despite having been disabled.

The disabled read projection only hides the stale fields while the flag is false and cannot restore the reset when callers re-enable directly. Make the field non-public and require the transition method (or otherwise make direct mutation incapable of bypassing the reset), then add the direct FALSE→TRUE regression test including summary visibility. If retaining the public `bool` is a hard compatibility constraint, this needs to remain an explicit unsupported/invariant-breaking API path rather than being described as enforcement.

crates/bacnet-server/src/handlers/write_property.rs
```diff
@@ -131,9 +226,11 @@ pub fn handle_write_property_multiple(
         // In that case fall back to snapshotting PRESENT_VALUE directly — the
         // same value the old code snapshotted — so the write still rolls back.
         // Without this fallback a failed multi-write would leave the
-        // non-commandable PRESENT_VALUE changed despite reporting failure
-        // (ASHRAE 135-2020 §19.1.2).
-        let rollback_records = if *prop_id == PropertyIdentifier::PRESENT_VALUE {
+        // non-commandable PRESENT_VALUE changed despite this implementation's
+        // all-or-nothing WPM policy.
+        let rollback_records = if object_rollback.is_some() {
```
jscott3201: **High — a rollback token silently disables rollback of the written property for downstream objects.** `Some(token)` selects an empty `rollback_records` vector. A custom object can reasonably use the new hook to save a private side effect while relying on the existing readable-property replay for the property itself: let `capture_write_property_rollback(DESCRIPTION)` return a token for a private counter, let `write_property(DESCRIPTION)` change the description, and let `restore_write_property_rollback` restore only that counter and return `Ok(())`. A WPM `[DESCRIPTION = "new", OBJECT_TYPE = ...]` then fails at the second write, reports only `WRITE_ACCESS_DENIED`, and leaves DESCRIPTION as `"new"` with no rollback error. The hook documentation says it captures state readback cannot preserve; it does not say it replaces restoration of readable state. Capture/replay the visible property as well, or make an explicit combined rollback contract with a regression test for a downstream object that has both visible and private state.

jscott3201: **High — unreadable snapshots still fail open, so WPM is not all-or-nothing for a transiently unreadable or write-only custom property.** When no token is provided, the fallback at `write_property.rs:287-292` converts a `read_property` error to no rollback record. Reproduce with a custom object whose writable DESCRIPTION is temporarily unreadable: WPM `[DESCRIPTION = "new", OBJECT_TYPE = 0]` accepts the first write, cannot snapshot it, then rejects OBJECT_TYPE and returns the original error while DESCRIPTION remains `"new"`; nothing reports a restoration failure because no restoration was attempted. This is a silent partial commit. Snapshot/preflight all writes before mutation and reject the request if any rollback state cannot be captured (or require a transactional object hook), with a regression test.

crates/bacnet-objects/src/event_enrollment/mod.rs
jscott3201, line 749: **Blocker — the public flag still bypasses the promised reset.** A downstream caller can set an Alert Enrollment to OFFNORMAL and clear an ack bit through the public `BACnetObject` internal methods, assign `event_detection_enable = false`, then assign it back to `true`. The first direct assignment only changes this field; it never calls the reset at lines 780-785. Once re-enabled, `read_property(EVENT_STATE)` exposes the stale OFFNORMAL value and `ACKED_TRANSITIONS` exposes the stale cleared bit. This is a concrete false → true sequence through the supported public API that resurrects state which disabling is supposed to reset. The projection while false does not repair the stored state. Make transitions through this public API resettable (rather than exposing a mutable bool), and add this exact false→true regression test.

crates/bacnet-server/src/handlers/write_property.rs
```diff
@@ -131,9 +226,11 @@ pub fn handle_write_property_multiple(
...
+            Vec::new()
```
jscott3201: [high] A rollback token can leave the database name index stale even when restoration succeeds. `BACnetObject::capture_write_property_rollback` is public and permits a token for any lossy write. A valid custom object can therefore return a token for `OBJECT_NAME`, write `old -> new`, and restore `old` from that token after a later WPM element fails. This branch suppresses the normal property snapshot, so rollback uses `ObjectState`; that arm never calls `update_name_index`. The object then reports `old`, while `find_by_name("new")` still resolves it and `old` is incorrectly available to another object. Preserve the property identifier on token records and resync the name index after restoring an `OBJECT_NAME` token (including the error/partial-restore case), or prohibit/special-case tokens for this property. Validate with a trait-object fixture whose name write has token-only side effects, followed by a failing read-only write.

crates/bacnet-server/src/handlers/write_property.rs
```diff
@@ -25,10 +25,99 @@ fn check_and_prepare_name_write(
...
+}
+
+fn error_after_rollback(
+    db: &mut ObjectDatabase,
+    applied: Vec<RollbackRecord>,
+    write_error: Error,
+) -> Error {
+    match rollback_writes(db, applied) {
+        Ok(()) => write_error,
+        Err(rollback_error) => Error::Encoding(format!(
```
jscott3201: [high] A surfaced rollback failure can leave changed runtime state without triggering the post-write paths. For example, a `supports_cov()` object can successfully change `PRESENT_VALUE`, return a rollback token whose restore fails/downcasts incompatibly, and then encounter a later read-only WPM write. This returns only `Err` even though the value remains changed (the new `FailingRollbackObject` test establishes that residual-mutation outcome). In `server/requests/mod.rs:134-145`, `written_oids` is populated only on `Ok`, while event/COV processing runs only over that list at lines 521-544. Subscribers therefore remain stale after the server has accepted a mutation it could not undo. Carry the dirty OIDs alongside the rollback error (or otherwise schedule event/COV reconciliation for every object whose rollback failed). Validate with a COV-capable failing-restore object and assert a failed WPM still reconciles/notifies its residual value, plus a multi-object case proving successful later rollbacks continue.

crates/bacnet-objects/src/event_enrollment/mod.rs
jscott3201, line 749: [high] The public field still permits a stale-state re-enable that defeats the new invariant. Concrete sequence: seed `event_state = OFFNORMAL`, assign `event_detection_enable = false` directly (the compatibility path does not call `set_event_detection_enable`), then assign it back to `true` directly. The reads project NORMAL while false, but line 748 exposes the original OFFNORMAL again after re-enable; `STATUS_FLAGS` likewise becomes IN_ALARM. This is an externally reachable state transition because the field is public, and the doc's recommendation cannot enforce it for existing callers. Remove the public mutable escape hatch (private field plus getter/setter), or make the re-enable/read path normalize state whenever a direct assignment was observed; add a direct-false/direct-true regression test.

crates/bacnet-server/src/handlers/write_property.rs
```diff
@@ -101,20 +198,18 @@ pub fn handle_write_property_multiple(
...
         // Enforce Object_Name uniqueness against the database index before
         // mutating the object, so a rejected duplicate leaves no trace.
         if *prop_id == PropertyIdentifier::OBJECT_NAME {
-            check_and_prepare_name_write(db, oid, value)?;
+            if let Err(error) = check_and_prepare_name_write(db, oid, value) {
+                return Err(error_after_rollback(db, applied, error));
+            }
         }
         let object = db.get_mut(oid).unwrap();
+        let object_rollback = object.capture_write_property_rollback(*prop_id);
```
jscott3201: [high] A rollback token suppresses the generic property snapshot, but the token API does not require the token to restore the property itself. Concrete downstream implementation: `capture_write_property_rollback(EVENT_DETECTION_ENABLE)` returns a token containing only private side effects, while `write_property` changes the public detection flag; if a later WPM write fails, `rollback_records` is empty at lines 231-232 and only `restore_write_property_rollback` runs. Unless every implementation duplicates the written property's old value in its opaque token, the flag remains false even though WPM returns failure. This is especially dangerous because `BACnetObject` is public and the default restore rejects tokens. Make the contract explicit/enforceable: either always retain the normal property snapshot in addition to the token, or define the token as a complete property-state snapshot and validate/require that before skipping the generic record; add a custom-object WPM test where the token preserves side effects but not the property.

crates/bacnet-server/src/handlers/write_property.rs
```diff
@@ -185,34 +341,42 @@ pub fn handle_write_property_multiple(
...
-                        // the database index so it no longer maps the rolled-back
-                        // name and once again maps the restored name.
-                        if rb_prop == PropertyIdentifier::OBJECT_NAME {
-                            db.update_name_index(&rb_oid);
-                        }
-                    }
+                // Object implementations should not mutate on error, but a
+                // captured snapshot lets rollback contain a partial write from
+                // a custom implementation rather than trusting that contract.
+                if has_rollback {
```
jscott3201, line 396: **[blocker] Do not drop an errored write merely because it was unreadable.** A custom `BACnetObject` can have `read_property(DESCRIPTION)` return `Err`, no rollback token, and have `write_property(DESCRIPTION, ...)` mutate state before returning `Err`. Here `has_rollback` is false, so no `RollbackRecord` is appended; `error_after_rollback` receives no record for that object, returns only the original write error, and leaves `residual_oids` empty. Production then sets `written_oids = residual_oids` at `server/requests/mod.rs:142`, so neither event nor COV reconciliation runs for the residual mutation. The same sequence on an unreadable `OBJECT_NAME` also leaves the database name index at the old name while the object has changed, allowing a later duplicate-name write. Record an unknown/empty rollback entry for every errored write (or otherwise mark that OID residual); let `rollback_writes` surface the unavailable snapshot and refresh the name index. Add a regression with an unreadable, mutate-then-error custom object, including `OBJECT_NAME`, that asserts rollback failure, residual reconciliation eligibility, and a synchronized name index.

jscott3201 (commented):
[Dataflow] READY

No actionable dataflow defect found in the requested WPM/Alert paths.

Residual risk / missing coverage: focused handler tests cover token-plus-visible replay, rejected replay validation, name-index recovery, and residual OID collection, but I did not execute tests in this read-only pass. The production WPM path assigns residual OIDs before response generation and runs event then COV reconciliation afterward (`crates/bacnet-server/src/server/requests/mod.rs:135-149, 525-548`); successful WPM uses only committed unique OIDs. The documented public Alert field’s direct FALSE→TRUE compatibility boundary was accepted as requested.

jscott3201 (commented):
[Dataflow] Immutable-target review completed. No actionable findings.

Stale check passed: `HEAD` = `cdd0faa3a98c6563a36081cddb3c6e4007250c79`; `merge-base HEAD origin/dev` = `daab3a8053a3960ea66ceaa7975c167be8e5d123`.

Reviewed full `daab3a...cdd0faa` diff and surrounding write, rollback, object, name-index, request reconciliation, and conformance-guard flows. Residual risk: rollback remains best-effort for third-party/write-only `BACnetObject` implementations; the documented side-effect-free `Err` contract and residual event/COV reconciliation are the containment boundary.

jscott3201 (commented):
[Dataflow] READY

No actionable findings from an independent read-only dataflow review at immutable head `46d63a72edafd2eafc33435450c17ec395bca2d0` (merge-base `daab3a8053a3960ea66ceaa7975c167be8e5d123`).

Verified: WPM applies/reverts records in reverse order; readable snapshots replay before object tokens; residual OIDs are propagated only for failed restoration and drive the request-path event/COV handoff; Object_Name index refresh follows complete rollback; log buffers, Network Port changes-pending, Channel priority metadata, Access Door slots, event state/history, and raw Time_Delay_Normal are token-covered. The rejected PRIORITY_ARRAY replay is checked after token restoration.

Residual gaps: Alert Enrollment remains intentionally partial (#264/#291); direct public-field compatibility is explicitly bounded; conformance evidence SHA remains the documented reviewed implementation SHA `272f86680f7de884dfd5928aa62682516bc6cdda`, not the documentation-only final head.

crates/bacnet-objects/src/event_enrollment/mod.rs
```diff
@@ -672,7 +739,16 @@ pub struct AlertEnrollmentObject {
...
```
jscott3201, line 749: [high] This public mutable flag still bypasses the reset. A caller can put the object in OFFNORMAL with an unacknowledged bit, assign `event_detection_enable = false`, then assign it directly back to `true`. Neither assignment calls `set_event_detection_enable`, so the old internal state is exposed again on the next `Event_State`/`Acked_Transitions` read without a new transition. That violates the required disabled-state initial-condition transition and makes the documented invariant conditional on callers remembering a convention. Please make mutation go through the reset path (or avoid exposing a writable field).

jscott3201 (commented):
[Dataflow] Blocking state-transition finding: Alert Enrollment's public mutable detection flag permits a false→true bypass of the reset, exposing pre-disable event state without a new transition. REQUEST_CHANGES is not permitted because this is my own pull request.

jscott3201 (commented):
[Adversary] Clean review: no material defect found in the immutable head’s WPM rollback handling after adversarial inspection of invalid/nonzero Record_Count variants, repeated clears, later-write failures, rollback errors, token-backed custom objects, name-index restoration, and residual reconciliation.

jscott3201 (commented):
[Runtime] No actionable runtime/correctness defect found in `daab3a8053a3960ea66ceaa7975c167be8e5d123...389fc6b7cba94b0bf3fccf72c4b39846b33b1dc0`. I verified the branch was created at the stated merge-base and the local/remote/PR heads all resolve to the immutable requested SHA. The head's Linux workspace test, Clippy, rustfmt, file-size, and secret-scan checks completed successfully.

jscott3201 (commented):
[Final Review] READY

Immutable-head review completed at `389fc6b7cba94b0bf3fccf72c4b39846b33b1dc0` (merge-base `daab3a8053a3960ea66ceaa7975c167be8e5d123`). Independent runtime, adversarial, dataflow, and contract reviews found no actionable defects.

The `Record_Count` rollback path now captures readable state before moving log buffers into rollback tokens, rejects invalid/nonzero values without mutation, restores moved buffers after later WPM failure, and drops them only after successful WPM completion. The documented direct-assignment Alert compatibility boundary and partial-property fences remain unchanged and tracked by #264/#291.

PR #310, opened 2026-08-14
fix(server): reject malformed list edits atomically

### LINKED ISSUE #279: AddListElement/RemoveListElement silently apply a partial element list when an element fails to decode
Confirmed pre-existing (present before PR #278), from adversarial review of #278.

`crates/bacnet-server/src/handlers/list.rs` add arm (~95-103) and remove arm (~169-177): the element loop does `Err(_) => break`, so when one element of `listOfElements` fails `decode_application_value`, the loop silently terminates, the partial edit is committed, and the service ACKs success.

Concrete: AddListElement to MSI Alarm_Values with one valid Unsigned followed by a TLV-formattable-but-untyped element adds half the requested list and returns SimpleACK. Inconsistent with the same file's `framed_destination_list_edit` (PR #271), which hard-errors on undecodable edit bytes per Clause 15.13/15.14.

Fix direction: fail the whole operation with a determinate error (no partial commit), matching the framed handler's posture.

jscott3201:
## Summary
- reject undecodable AddListElement and RemoveListElement members instead of committing the decoded prefix
- use the existing `PROPERTY / INVALID_DATA_TYPE` mapping for generic and framed list edits
- verify malformed tails leave Alarm_Values unchanged for both services

## Validation
- `cargo test -p bacnet-server --locked handlers::list`

Closes #279

### DIFF
--- a/crates/bacnet-server/src/handlers/list.rs
+++ b/crates/bacnet-server/src/handlers/list.rs
@@ -1,6 +1,13 @@
 use super::*;
 use bacnet_objects::traits::BACnetObject;
 
+fn invalid_data_type() -> Error {
+    Error::Protocol {
+        class: ErrorClass::PROPERTY.to_raw() as u32,
+        code: ErrorCode::INVALID_DATA_TYPE.to_raw() as u32,
+    }
+}
+
 /// Apply an AddListElement/RemoveListElement edit to a property whose wire
 /// form is a framed `BACnetLIST of BACnetDestination` (NotificationClass
 /// `Recipient_List`): it reads back as [`PropertyValue::ApplicationData`]
@@ -20,10 +27,6 @@ fn framed_destination_list_edit(
     edit_bytes: &[u8],
     remove: bool,
 ) -> Result<(), Error> {
-    let invalid_data_type = || Error::Protocol {
-        class: ErrorClass::PROPERTY.to_raw() as u32,
-        code: ErrorCode::INVALID_DATA_TYPE.to_raw() as u32,
-    };
     let mut destinations = bacnet_encoding::constructed::decode_destination_list(current_bytes)
         .map_err(|_| invalid_data_type())?;
     let edits = bacnet_encoding::constructed::decode_destination_list(edit_bytes)
@@ -93,13 +96,10 @@ pub fn handle_add_list_element(db: &mut ObjectDatabase, service_data: &[u8]) -> 
     let mut offset = 0;
     let data = &request.list_of_elements;
     while offset < data.len() {
-        match decode_application_value(data, offset) {
-            Ok((val, new_offset)) => {
-                items.push(val);
-                offset = new_offset;
-            }
-            Err(_) => break,
-        }
+        let (val, new_offset) =
+            decode_application_value(data, offset).map_err(|_| invalid_data_type())?;
+        items.push(val);
+        offset = new_offset;
     }
 
     object
@@ -167,13 +167,10 @@ pub fn handle_remove_list_element(
     let mut offset = 0;
     let data = &request.list_of_elements;
     while offset < data.len() {
-        match decode_application_value(data, offset) {
-            Ok((val, new_offset)) => {
-                to_remove.push(val);
-                offset = new_offset;
-            }
-            Err(_) => break,
-        }
+        let (val, new_offset) =
+            decode_application_value(data, offset).map_err(|_| invalid_data_type())?;
+        to_remove.push(val);
+        offset = new_offset;
     }
 
     // Remove matching elements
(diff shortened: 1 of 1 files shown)


jscott3201 (commented):
[Dataflow] READY

Decision: no actionable PR-scope defects found in `98221e392b42882ac30d5ac92fbcb764dbeed07e...f238acc17cc17064b7924419f10e711ff27cd834`.

Data flow: `ListElementRequest` preserves the enclosed element bytes; both generic handlers now decode every member into a temporary vector before `write_property` (`crates/bacnet-server/src/handlers/list.rs:96-125`, `166-184`). A malformed member returns `PROPERTY/INVALID_DATA_TYPE` before the only write. The framed Recipient_List path strictly decodes both stored and request frames, merges/removes only after both decodes, then writes the re-framed result (`list.rs:30-46`); both service arms reach it (`list.rs:67-89`, `151-159`). The confirmed-request dispatcher converts this protocol error to an Error APDU with the same class/code (`crates/bacnet-server/src/server/requests/mod.rs:546-565`).

Findings: none.

Missing tests (non-blocking): add a malformed framed AddListElement preservation case parallel to the existing framed Remove test (`list.rs:456-502`), and an end-to-end confirmed-request assertion that the dispatcher emits `PROPERTY/INVALID_DATA_TYPE` rather than only checking the handler result.

jscott3201 (commented):
[Runtime] READY

## Decision
READY. No material runtime-correctness, atomicity, error-propagation, progress, or regression-test defect found in `98221e392b42882ac30d5ac92fbcb764dbeed07e...f238acc17cc17064b7924419f10e711ff27cd834`.

## Findings by severity
None. Explicitly, no shipping-blocking or material non-blocking findings survived review.

## Missing coverage (non-blocking)
- The generic success test uses one element for each operation (`crates/bacnet-server/src/handlers/list.rs:217-241`); there is no direct generic-handler success case for `[0x21, 2, 0x21, 3]` adding/removing both elements. The successful loop behavior is unchanged by this patch.
- The framed malformed-payload test covers RemoveListElement with an invalid first element (`crates/bacnet-server/src/handlers/list.rs:475-502`), but not AddListElement or a valid framed destination followed by malformed bytes. The shared helper nevertheless decodes the complete edit list before extending/retaining and before `write_property` (`crates/bacnet-server/src/handlers/list.rs:30-46`).

## Validation performed
- Reverified the PR base/head as `98221e392b42882ac30d5ac92fbcb764dbeed07e` / `f238acc17cc17064b7924419f10e711ff27cd834`; reviewed the authoritative one-file compare and surrounding decoder/object/dispatcher code.
- Traced concrete payload `[0x21, 0x02, 0xD1, 0x00]`. On the base implementation, Add against `[1]` decodes `2`, breaks on application tag 13, writes `[1,2]`, and returns success; Remove against `[2,3]` writes `[3]` and returns success. Therefore the new tests at `crates/bacnet-server/src/handlers/list.rs:243-291` fail on the base at `unwrap_err`, while the head errors with `PROPERTY / INVALID_DATA_TYPE` before `write_property` and preserves the stored list.
- Verified generic mutation order: Add fully decodes into a local list before the only write (`crates/bacnet-server/src/handlers/list.rs:96-125`); Remove fully decodes before `retain` and write (`crates/bacnet-server/src/handlers/list.rs:166-184`).
- Verified loop progress against `decode_application_value`: every successful branch advances past at least the decoded tag (`crates/bacnet-encoding/src/primitives/mod.rs:460-542`), so these loops cannot spin on a successful zero-progress decode.
- Head CI completed successfully for the Ubuntu test suite, Clippy, rustfmt, file-size, and secret-scan checks. Base failure was verified by source-level trace rather than by executing a locally patched base checkout.

jscott3201 (commented):
[Adversary] READY

No material defects found in the #279 scope.

Evidence: both generic handlers now decode every `listOfElements` member before reaching their only write (`crates/bacnet-server/src/handlers/list.rs:98-111`, `169-184`). A TLV-formattable but unsupported member (for example `21 02 D1 00`) fails `decode_application_value` (`crates/bacnet-encoding/src/primitives/mod.rs:534-538`) and is mapped to `PROPERTY/INVALID_DATA_TYPE` (`list.rs:4-9`), so the previous valid prefix cannot be committed. The added Add and Remove regression tests assert that exact error and preservation of the original list (`list.rs:244-290`). The confirmed-service dispatcher turns the handler error into an Error APDU rather than a SimpleACK (`crates/bacnet-server/src/server/requests/mod.rs:306-318`, `583-601`).

Unclaimed: malformed outer request framing is rejected by `ListElementRequest::decode` before either handler (`crates/bacnet-services/src/list_manipulation.rs:67-70`); it was already outside the partial-prefix behavior fixed by #279. I did not find a regression in the requested change.

Issue #309, opened 2026-08-14
Direct service decoders silently truncate over-wide Unsigned fields


jscott3201:
### Affected revision

Observed on `dev` after #308 at `98221e3`.

### Behavior

PR #308 fixed the generic application-tagged Enumerated decoder, but the follow-up narrowing audit found 137 direct `decode_unsigned(...)? as u8/u16/u32` conversions across 29 Rust files. These paths decode peer-controlled 1-8-octet Unsigned or Enumerated content to `u64`, then silently discard high bits when the destination field is narrower.

A concrete range-bypass is `WritePropertyRequest::decode` in `crates/bacnet-services/src/write_property.rs:103-119`: priority is narrowed with `as u8` before the required 1..=16 check. A wire value of 257 therefore becomes priority 1 and passes validation. Other representative cases include GetEnrollmentSummary priority and notification-class filters (`crates/bacnet-services/src/enrollment_summary.rs:175-220`) and ErrorPDU error class/code (`crates/bacnet-encoding/src/apdu/mod.rs:547-588`). RPM has the same ErrorPDU-style narrowing in `crates/bacnet-services/src/rpm.rs:296-305`.

The audit command was:

```text
rg -n 'decode_unsigned\\([^\\n]*\\)\\? as u(?:8|16|32)' crates --glob '*.rs'\n```\n\nThis is distinct from accepting a wider leading-zero representation whose numerical value fits the target. The defect is accepting a numerical value above the target type and wrapping it modulo 2^N.\n\n### Suggested direction\n\n- add checked typed decode helpers for `u8`, `u16`, and `u32`, preserving the current 1-8-octet unsigned parser;\n- replace unchecked peer-input narrowing in small crate/layer slices rather than one repository-wide PR;\n- start with fields whose post-cast validation can be bypassed, including WriteProperty priority;\n- add hostile values immediately above each target maximum and low-word-valid payloads;\n- require decode errors before state mutation or request dispatch.\n\n### Relevant code\n\n- `crates/bacnet-encoding/src/primitives/mod.rs`\n- `crates/bacnet-encoding/src/apdu/mod.rs`\n- `crates/bacnet-services/src/write_property.rs`\n- `crates/bacnet-services/src/enrollment_summary.rs`\n- `crates/bacnet-services/src/rpm.rs`\n\nFollow-up to #277 / #308.

jscott3201:
[...]

PR #311, opened 2026-08-14
fix(services): validate write priority before narrowing

jscott3201:
## Summary
- validate WriteProperty priority against 1..=16 as the full decoded Unsigned value
- prevent over-wide values such as 257 and 272 from aliasing valid priorities 1 and 16
- preserve receiver tolerance for a leading-zero representation of numeric priority 1

## Validation
- `cargo clean`
- `cargo test -p bacnet-services --locked write_property::tests::priority_validation`

Refs #309

### DIFF
--- a/crates/bacnet-services/src/write_property.rs
+++ b/crates/bacnet-services/src/write_property.rs
@@ -109,14 +109,16 @@ impl WritePropertyRequest {
                 if end > data.len() {
                     return Err(Error::decoding(pos, "WriteProperty truncated at priority"));
                 }
-                let prio = primitives::decode_unsigned(&data[pos..end])? as u8;
+                let prio = primitives::decode_unsigned(&data[pos..end])?;
                 if !(1..=16).contains(&prio) {
                     return Err(Error::decoding(
                         pos,
                         format!("WriteProperty priority {prio} out of range 1-16"),
                     ));
                 }
-                priority = Some(prio);
+                priority = Some(u8::try_from(prio).map_err(|_| {
+                    Error::decoding(pos, "WriteProperty priority conversion failed")
+                })?);
             }
         }
 
@@ -178,6 +180,44 @@ mod tests {
         req.encode(&mut buf);
         let decoded = WritePropertyRequest::decode(&buf).unwrap();
         assert_eq!(decoded.priority, Some(16));
+
+        for content in [
+            &[0x00][..],
+            &[0x11][..],
+            &[0x01, 0x01][..],
+            &[0x01, 0x10][..],
+        ] {
+            let mut buf = BytesMut::new();
+            WritePropertyRequest {
+                priority: None,
+                ..req.clone()
+            }
+            .encode(&mut buf);
+            tags::encode_tag(
+                &mut buf,
+                4,
+                TagClass::Context,
+                u32::try_from(content.len()).unwrap(),
+            );
+            buf.extend_from_slice(content);
+            assert!(
+                WritePropertyRequest::decode(&buf).is_err(),
+                "priority content {content:02X?} must be rejected"
+            );
+        }
+
+        let mut buf = BytesMut::new();
+        WritePropertyRequest {
+            priority: None,
+            ..req
+        }
+        .encode(&mut buf);
+        tags::encode_tag(&mut buf, 4, TagClass::Context, 2);
+        buf.extend_from_slice(&[0x00, 0x01]);
+        assert_eq!(
+            WritePropertyRequest::decode(&buf).unwrap().priority,
+            Some(1)
+        );
     }
 
     // -----------------------------------------------------------------------


jscott3201 (commented):
[Dataflow] READY

Decision: READY. No material defects found in the authoritative `7a521674513465419edde26f9746063fd7b10b7c...125fc74c47096f996c4defbf66ee6e2d597b943a` comparison.

Data flow: the context-tag-4 bytes now decode losslessly to `u64`, are range-checked before narrowing, and only then become `Option<u8>` (`crates/bacnet-services/src/write_property.rs:112-120`). `handle_write_property` performs this decode before value decode or `object.write_property` (`crates/bacnet-server/src/handlers/write_property.rs:493-535`), and the confirmed-service dispatcher calls that handler under the database write lock (`crates/bacnet-server/src/server/requests/mod.rs:110-121`). Therefore a rejected priority cannot reach the mutation boundary.

Verified values from the decoder semantics and tests:
- `0`, `17`, `257` (`01 01`), `272` (`01 10`), and `u64::MAX` fail the raw `u64` range check.
- Canonical `1` and `16` are retained; leading-zero numeric `1` (`00 01`) is retained.
- `decode_unsigned` accepts 1–8 big-endian bytes and produces the full `u64`, including `u64::MAX` (`crates/bacnet-encoding/src/primitives/mod.rs:62-74`); the range guard consequently rejects it without narrowing.

Findings by severity: none.

Missing tests (non-blocking): `priority_validation` covers 0, 17, 257, 272, 16, and leading-zero 1 (`crates/bacnet-services/src/write_property.rs:180-222`), but does not explicitly assert canonical `1`, `u64::MAX`, or handler/dispatcher-level no-mutation behavior for malformed priorities. The implementation’s ordering establishes those cases; adding explicit coverage would harden the boundary contract.

Explicit no-findings statement: no data-flow, ordering, mutation-boundary, compatibility, or cross-platform defect was found within the requested single-WriteProperty scope.

jscott3201 (commented):
[Luna] READY — no material findings in the authoritative `7a521674513465419edde26f9746063fd7b10b7c...125fc74c47096f996c4defbf66ee6e2d597b943a` compare.

## Decision
READY.

## Findings by severity
- None.

The validation order is correct: `crates/bacnet-services/src/write_property.rs:112-117` decodes the full Unsigned as `u64` and checks `1..=16` before narrowing. Thus encoded contents `01 01` (257) and `01 10` (272) are rejected instead of aliasing to priorities 1 and 16, while `00 01` remains accepted as priority 1. That tolerance follows the existing big-endian unsigned decoder at `crates/bacnet-encoding/src/primitives/mod.rs:65-77`, which preserves leading-zero representations.

The `u8::try_from` error path at `crates/bacnet-services/src/write_property.rs:119-121` is logically unreachable under the immediately preceding range check, but it is harmless defensive narrowing and is acceptable for maintainability; it does not weaken validation or create an observable bad path. The added tests at `crates/bacnet-services/src/write_property.rs:184-220` cover zero, 17, overflow aliases, and a leading-zero valid value.

## Assumptions / gaps
- This review is limited to the requested Clause 15.9 priority contract and the authoritative one-file diff; no independent interop run was performed.
- No finding requires changing or closing umbrella issue #309.

## Explicit no-findings statement
No blocker, high, medium, or low-severity defect was found in this change.

jscott3201 (commented):
[Adversary] READY

Decision: READY for `7a521674513465419edde26f9746063fd7b10b7c...125fc74c47096f996c4defbf66ee6e2d597b943a`.

Findings by severity: None.

Evidence: `crates/bacnet-services/src/write_property.rs:112-121` decodes the full priority as `u64`, range-checks it before `u8` conversion, and propagates failures. `crates/bacnet-encoding/src/primitives/mod.rs:65-78` rejects zero- and over-eight-octet Unsigned contents. `crates/bacnet-server/src/handlers/write_property.rs:485-531` decodes the request before obtaining a mutable object/write. The regression cases at `crates/bacnet-services/src/write_property.rs:184-220` use an otherwise valid request and cover 0, 17, 257, 272, plus leading-zero 1.

Unclaimed regressions: None in PR scope. No material PR-scope defects found.

jscott3201 (commented):
[Runtime] READY

## Decision
No material runtime-correctness or regression defect found in `7a521674513465419edde26f9746063fd7b10b7c...125fc74c47096f996c4defbf66ee6e2d597b943a`.

## Findings by severity
None. The decoder now keeps the peer-controlled Unsigned as `u64`, applies the required `1..=16` check to that full value, and only then performs checked narrowing (`crates/bacnet-services/src/write_property.rs:112-121`). Thus 257 and 272 can no longer alias priorities 1 and 16.

## Missing coverage
Non-blocking: the service test does not directly construct a zero-length priority, a complete priority longer than eight octets, or a priority whose declared length exceeds the remaining bytes. Current behavior is still safe: the bounds check rejects truncation at `crates/bacnet-services/src/write_property.rs:108-111`, while `decode_unsigned` rejects lengths outside 1-8 at `crates/bacnet-encoding/src/primitives/mod.rs:65-77`. Dedicated WriteProperty cases would protect those branches more directly.

## Validation performed
- Verified the PR still targets base `7a521674513465419edde26f9746063fd7b10b7c` and head `125fc74c47096f996c4defbf66ee6e2d597b943a`.
- Inspected the complete one-file compare plus the base/head implementations, tag decoder, and Unsigned codec.
- Confirmed the added vectors at `crates/bacnet-services/src/write_property.rs:184-220` reject 0, 17, 257, and 272 while accepting the leading-zero representation `[0x00, 0x01]` as numeric priority 1.
- Confirmed source-level base failure: at the base revision, the old narrowing at `crates/bacnet-services/src/write_property.rs:112` maps 257 to 1 and 272 to 16, so the new `is_err()` assertions for `[0x01, 0x01]` and `[0x01, 0x10]` fail on base. This differential was established from the exact base/head source; it was not locally re-run.
- Observed successful head CI for Test, Clippy, Rustfmt, file-size, and secret-scan checks.

Explicitly: no material findings.

PR #312, opened 2026-08-14
fix(transport): apply SC must-understand to destination options

### LINKED ISSUE #191: BACnet/SC Must-Understand handling is applied to Data Options, which AB.3.1.4 requires be forwarded unaltered, while received Destination Options are never checked
### Affected revision

### Behavior

The Must-Understand rule is enforced on the wrong header-option list, in both directions.

`crates/bacnet-transport/src/sc/data_attributes.rs:24-32` iterates `msg.data_options` and returns the first Must-Understand option that `is_understood_data_option` (`:105-107`, which recognizes only type 1, Secure Path) rejects. `reject_unsupported_must_understand_data_option` (`:45-75`) then drops the message and, for non-broadcast traffic, sends a BVLC-Result NAK COMMUNICATION/HEADER_NOT_UNDERSTOOD (`:60-66`) — the only HEADER_NOT_UNDERSTOOD emitter in the codebase. It runs at `crates/bacnet-transport/src/sc/mod.rs:415`, before the NPDU is delivered, and `continue`s the receive loop.

Clause AB.3.1.4 (p1395) assigns that NAK exclusively to destination options: "If a destination option is present whose 'Must Understand' flag is set (1) but the option is unknown or not supported by the BVLL entity of the destination node, then if the original message was a unicast BVLC message, a BVLC-Result NAK for the BVLC message shall be returned indicating an 'Error Class' of COMMUNICATION and an 'Error Code' of HEADER_NOT_UNDERSTOOD." The next paragraph handles data options oppositely: "For the handling of 'Data Options' see Clause AB.3.4. The hub function and the source and destination node's BVLL entity shall forward and not alter any of the data options." AB.3.4 (p1397) is explicit about what to do instead: "Data options present in the message shall be forwarded to the local network entity in the 'data_attributes' parameter of the DL_UNITDATA.indication primitive." A grep of the whole standard finds no Must-Understand processing rule for data options anywhere.

So the NPDU is dropped where the standard requires it be delivered with its options attached.

Conversely, received destination options are never inspected. A grep for `dest_options` outside `sc_frame.rs` and tests returns only `dest_options: Vec::new()` initializers on the send side, and the receive path at `crates/bacnet-transport/src/sc/mod.rs:457-462` consumes only `from_data_options(&msg)` and discards `msg.dest_options` — so the one case that does require the NAK is silently ignored. AB.2.3 (p1387) explains why the two lists differ: destination options "are used at the data link layer only", while data options "accompany the NPDU to the ultimate destination device".

`docs/conformance/bacnet-135-2020.json:684` presents the data-options NAK as conformance evidence and needs correcting. Rows `:614` and `:645` already record "destination-option Must Understand NAK behavior" as remaining work, so the two halves of this issue are already half-acknowledged in that file.

### Suggested direction

- move the Must-Understand check to `msg.dest_options`, keeping the existing unicast-NAK / broadcast-drop split, which is correct for that list;
- stop dropping or NAKing on unknown data options; forward every data option, understood or not, as a `DataAttribute` on `ReceivedNpdu`;
- keep Secure Path (type 1) recognition as a consumer-side concern rather than a gate;
- correct the `docs/conformance/bacnet-135-2020.json` row that cites the data-options NAK as evidence, and regenerate the derived conformance docs;
- test an unknown Must-Understand data option surviving to `data_attributes`, an unknown Must-Understand destination option producing the unicast NAK, and the same destination option being silently dropped on broadcast.

### Relevant code

- `crates/bacnet-transport/src/sc/data_attributes.rs`
- `crates/bacnet-transport/src/sc/mod.rs`
- `docs/conformance/bacnet-135-2020.json`

Pairs with #157, which covers the Secure Path type-1 marker constraints on the send path.

jscott3201:
## Summary
- apply unknown Must-Understand rejection to BACnet/SC Destination Options instead of Data Options
- preserve the complete offending Destination Option marker across hub VMAC transformation and in `COMMUNICATION / HEADER_NOT_UNDERSTOOD` Results
- route valid Encapsulated-NPDU rejection Results to the originating node without closing its established connection
- drop malformed peer Results and peer Results for other functions before relay
- deliver unknown Data Options, including options with bit 6 set, through `ReceivedNpdu.data_attributes`
- keep broadcast Destination Option rejection silent and regenerate the conformance evidence

Addendum 135-2020cf Every Segment semantics and the context-neutral `must_understand` field name remain explicitly unclaimed. Must-Understand processing for BVLC-SC functions other than Encapsulated-NPDU remains tracked as broader Annex AB work.

## Validation
- `cargo clean`
- `cargo test -p bacnet-transport --locked`
- `cargo test -p bacnet-benchmarks --test sc_websocket --locked`

Closes #191


jscott3201 (commented):
[Dataflow] NOT READY

- **High — the new unicast NAK cannot reach the originating node when traffic uses the in-tree hub.** `crates/bacnet-transport/src/sc/data_attributes.rs:58-70` sends the rejection on the recipient's WebSocket, and `:85-91` constructs that BVLC-Result without either VMAC. A real A→hub→B unicast therefore reaches B with A's Originating VMAC, but B's new rejection is then received by the hub as `ScFunction::Result`; `crates/bacnet-transport/src/sc_hub.rs:657-668` has no Result relay path and instead returns an `UNEXPECTED_DATA` NAK only to B. A never observes the required `COMMUNICATION / HEADER_NOT_UNDERSTOOD` result. The added test (`crates/bacnet-transport/src/sc/data_attribute_tests.rs:66-99`) uses a direct loopback socket, so it cannot expose this integration failure. Please make the response routable through the hub (and add an A→hub→B integration test) before claiming unicast NAK behavior.

Reviewed target: `8d8466e8ca170ba8447a7bd13da7754f19a4211c...76229055f9f402ccad7b4153b671447cc78f875e`.

jscott3201 (commented):
[Runtime] NOT READY

1. **[P1] The unicast NAK cannot reach the original sender through the supported hub topology.** `crates/bacnet-transport/src/sc/data_attributes.rs:58-69` writes the `BVLC-Result` back on the receiving node's hub WebSocket, while the builder at `crates/bacnet-transport/src/sc/data_attributes.rs:85-91` gives that Result no destination VMAC. The real hub relays only `EncapsulatedNpdu` (`crates/bacnet-transport/src/sc_hub.rs:479-655`); a Result falls into the unsupported-function arm and is NAKed back to the rejecting node (`crates/bacnet-transport/src/sc_hub.rs:657-668`). Concrete sequence: A unicasts an NPDU with an unsupported Must-Understand Destination Option through `ScHub` to B; B emits the new Result, the hub returns `UNEXPECTED_DATA` to B, B transitions to `Disconnected` on that NAK (`crates/bacnet-transport/src/sc/connection.rs:272-289`), and A never receives `HEADER_NOT_UNDERSTOOD`. The new test at `crates/bacnet-transport/src/sc/data_attribute_tests.rs:66-99` uses a raw loopback peer, so it observes B's outbound frame directly and misses the real routing/disconnection failure.

2. **[P2] Destination-option enforcement still applies only to Encapsulated-NPDU.** `crates/bacnet-transport/src/sc/data_attributes.rs:23-30` returns early for every other BVLC function, although AB.3.1.4's destination-node option processing is not limited to NPDUs. For example, a unicast `HeartbeatRequest` carrying an unknown Must-Understand Destination Option bypasses the check and receives a normal Heartbeat-ACK at `crates/bacnet-transport/src/sc/mod.rs:423-435`; a `DisconnectRequest` can likewise reach the state mutation at `crates/bacnet-transport/src/sc/connection.rs:253-269`. Result frames do need the no-Result-on-Result exception, but excluding every non-NPDU function leaves the required rejection incomplete. No added test covers another non-Result function.

3. **[P2] The offending wire marker is not preserved.** The decoder reads `has_data` and `more_follows` at `crates/bacnet-transport/src/sc_frame.rs:341-345` but drops both when constructing `ScOption` at `crates/bacnet-transport/src/sc_frame.rs:371-375`; `option_header_marker` then reconstructs only type, Must-Understand, and nonempty-data bits (`crates/bacnet-transport/src/sc/data_attributes.rs:33-41`). Thus an unsupported first option whose wire marker is `0xC2` is reported as `0x42`, and an option encoded with Header-Data set and a zero-length value also loses bit 5. The latter is additionally re-encoded without its original marker by `crates/bacnet-transport/src/sc_frame.rs:240-252`, so such a Data Option is not forwarded unaltered. The tests only use a single option and cannot encode the zero-length Header-Data form, so neither case is exercised.

GitHub's head checks pass, and the new tests would fail on the base behavior, but they do not cover these runtime paths.

jscott3201 (commented):
[Adversary] NOT READY

**P1 — the unicast NAK cannot reach the original sender through the supported hub topology.** `crates/bacnet-transport/src/sc_hub/relay.rs:35-39` turns A→B unicast into a frame with `originating_vmac=A` and no destination VMAC. When B rejects its Must-Understand destination option, `crates/bacnet-transport/src/sc/data_attributes.rs:85-100` creates a BVLC-Result with *both* VMAC fields absent. The hub only relays `EncapsulatedNpdu` (`crates/bacnet-transport/src/sc_hub.rs:479-654`); it treats that Result as an unsupported function and returns a second NAK only to B (`:657-668`). Thus A never receives `COMMUNICATION/HEADER_NOT_UNDERSTOOD`; B receives the hub's NAK and disconnects because any NAK sets its state to Disconnected (`crates/bacnet-transport/src/sc/connection.rs:272-293`). The new test injects the already-relayed frame directly into B and reads B's outbound socket (`crates/bacnet-transport/src/sc/data_attribute_tests.rs:73-94`), so it cannot detect this failure. The error Result must be addressed to the inbound originating VMAC and the hub must route that response before this behavior can be claimed.

**P2 — multi-option errors report the wrong Error Header Marker.** The parser observes `more_follows` but does not retain it (`crates/bacnet-transport/src/sc_frame.rs:341-377`); `option_header_marker()` then reconstructs only bits 6..0 (`crates/bacnet-transport/src/sc/data_attributes.rs:33-41`) and uses that value in the NAK (`:52-65`). For example, the first unsupported Must-Understand type-2 option in a two-option destination chain was received as marker `0xC2`, but the Result reports `0x42`. AB.2.4 requires the marker of the destination option that caused the failure. The sole-marker test at `crates/bacnet-transport/src/sc/data_attribute_tests.rs:68-94` misses this case.

jscott3201 (commented):
[Dataflow] NOT READY

- **HIGH — arbitrary peer Results can disconnect another established node.** `crates/bacnet-transport/src/sc_hub/relay.rs:77-152` accepts every structurally addressed `Result` from any registered client and relays it to the chosen destination without validating its payload or limiting it to the Encapsulated-NPDU rejection flow. The relay transformation makes it appear to A as a hub-delivered result. At `crates/bacnet-transport/src/sc/connection.rs:279-294`, a malformed Result, or a valid NAK whose `result_for` is anything other than `EncapsulatedNpdu`, transitions A to `Disconnected`. Concrete sequence: connected B sends `Result { originating_vmac: None, destination_vmac: Some(A), payload: [ConnectRequest, NAK, ...] }` (or an empty/malformed Result); the hub relays it with B as Originating VMAC; A disconnects. This newly opens a cross-client availability attack and is outside the Encapsulated-NPDU-only scope. Decode and constrain relayable peer Results to the expected Encapsulated-NPDU result flow (or correlate them), and add a live-hub test proving malformed/non-Encapsulated Results cannot affect A.

Coverage gap: the intended path is tested in separate pieces—recipient rejection with a loopback peer (`crates/bacnet-transport/src/sc/data_attribute_tests.rs:65-105`), raw WebSocket Result forwarding (`benchmarks/tests/sc_websocket.rs:182-224`), and A’s nonfatal handling with a loopback peer (`crates/bacnet-transport/src/sc/result_tests.rs:546-582`). None composes an A→hub→B destination-option rejection through real `ScTransport` endpoints and asserts the returned marker, B→hub→A addressing transform, no NPDU delivery, and A remaining connected.

Residual evidence note: the changed source ledger still identifies `f485021f...` and the unrelated C3 scope as the reviewed implementation (`docs/conformance/bacnet-135-2020.json:4-6`); the human ledger repeats that provenance (`docs/conformance/standard-135-2020-ledger.md:7-10`). The generated PICS anchors match the updated JSON rows (`docs/conformance/pics-draft.md:22-23`), but the source/human evidence cannot truthfully identify this PR’s implementation SHA until the provenance is refreshed.

jscott3201 (commented):
[Adversary] NOT READY

P1 — Any registered node can now disconnect any other connected node by spoofing a routed Result. For example, with A, B, and C connected, C can send a syntactically valid `Result` with no Originating VMAC, Destination VMAC=A, and a NAK payload whose `result_for` is `Result` (or `ConnectRequest`), despite A not having sent such a request. The new relay accepts any unicast Result and rewrites/sends it to A. A then treats every non-Encapsulated-NPDU NAK as fatal and stops its receive loop. An empty or otherwise malformed Result payload has the same outcome via the malformed-Result path. Before this PR the hub did not route Results to arbitrary peers. This makes one authenticated/compromised client able to tear down unrelated live SC transports and covers Result-on-Result as well.

Evidence: `crates/bacnet-transport/src/sc_hub.rs:479-489` routes every connected-client Result; `crates/bacnet-transport/src/sc_hub/relay.rs:85-103` validates only Originating/Destination VMAC shape and never validates the Result payload or correlates it to an Encapsulated-NPDU; `crates/bacnet-transport/src/sc/connection.rs:272-295` disconnects on a malformed Result and on every NAK whose `result_for != EncapsulatedNpdu`; `crates/bacnet-transport/src/sc/mod.rs:452-453` terminates the receive loop after that state transition. Restrict the new route to the supported Encapsulated-NPDU response form (and reject malformed Results) or add request/response correlation before forwarding.

P2 — The promised complete offending marker is not preserved through the real hub when the offending Destination Option has Header Data Flag set with a zero length. Send A→hub→B an Encapsulated-NPDU whose offending marker is `0x62` followed by `00 00` (type 2, Must Understand, Header Data present but empty). The hub decode represents it as `ScOption { data: Vec::new() }`, then its relay re-encodes it as `0x42`; B consequently returns a NAK containing `0x42`, so A never receives the original `0x62`. The existing empty-Header-Data test injects directly into the transport and therefore misses this hub path.

Evidence: `crates/bacnet-transport/src/sc_frame.rs:352-365` collapses both no-data and zero-length Header Data to an empty vector; `crates/bacnet-transport/src/sc_hub/relay.rs:101-103` re-encodes that lossy representation; `crates/bacnet-transport/src/sc_frame.rs:242-261` sets bit 5 only when the vector is nonempty; and `crates/bacnet-transport/src/sc/mod.rs:418-425` extracts the error marker from B's already-modified relayed wire bytes. Preserve Header Data presence in the option representation (or retain the ingress marker through relaying) and add an A→hub→B→hub→A empty-Header-Data integration test.

jscott3201 (commented):
[Runtime] NOT READY

### Actionable finding

- **P1 — The complete Error Header Marker is still lost on the real hub-routed path** (`crates/bacnet-transport/src/sc_hub.rs:525-528`, `crates/bacnet-transport/src/sc_frame.rs:169-177`, `crates/bacnet-transport/src/sc_frame.rs:239-262`, `crates/bacnet-transport/src/sc/data_attribute_tests.rs:126-130`). A valid peer can send a unicast Encapsulated-NPDU whose unsupported Destination Option marker is `0x62` (Must Understand + Header Data present + type 2) followed by a zero length. The hub decodes that option into `ScOption { data: Vec::new(), ... }`, which has no field retaining the Header Data flag, then re-encodes every relay at `sc_hub.rs:525-528`. `encode_sc_options` derives the flag from `!opt.data.is_empty()`, so the relayed marker becomes `0x42`. The destination's new raw-marker helper therefore sees only `0x42`, and the originating node receives a NAK with the wrong Error Header Marker. The same normalization also alters an empty-header Data Option despite the issue's forward-unaltered requirement. The new marker test injects modified wire bytes directly into the loopback transport and bypasses `ScHub`; the TLS test relays an already-constructed, options-free Result, so neither exercises this failure. Preserve the empty Header Data flag across hub decode/re-encode (or relay the option bytes without normalization) and add an end-to-end raw-wire test through the real TLS hub.

### Residual nonblocking gaps

- Must-Understand processing for other BVLC-SC functions, Every Segment semantics, and direct-connection Result routing remain explicitly deferred and are not treated as blockers here.
- The new Result relay awaits the target sink without the timeout used by broadcast relay; this retains the hub's existing unicast backpressure risk but is not specific enough to block this scoped correction.

### Validation / evidence

- Reverified PR base/head as `8d8466e8ca170ba8447a7bd13da7754f19a4211c...108c33f590ebf6218fec77be48bbacd072759b98` immediately before submission.
- Inspected the full diff and surrounding codec, transport receive loop, connection state machine, hub relay, loopback tests, and live TLS test.
- Verified that unicast Result addressing/relay and nonfatal Encapsulated-NPDU NAK handling are implemented and directly tested.
- GitHub checks at this head report Rustfmt, Clippy, Ubuntu tests, file-size, and no-secret checks successful; the defect above is a missing end-to-end wire-preservation case.

jscott3201 (commented):
[Dataflow] READY

No actionable findings.

Verified flow: raw-marker recovery preserves the first Must-Understand Destination Option marker, including More Options and empty Header Data (`crates/bacnet-transport/src/sc_frame.rs:385-425`); hub relay rewrites only VMAC fields and copies the validated option/payload suffix (`crates/bacnet-transport/src/sc_hub/relay.rs:64-93`); destination rejection emits source-addressed COMMUNICATION/HEADER_NOT_UNDERSTOOD NAKs and silently drops broadcasts (`crates/bacnet-transport/src/sc/data_attributes.rs:33-70`); all Data Options remain mapped into `ReceivedNpdu.data_attributes` (`crates/bacnet-transport/src/sc/data_attributes.rs:12-20`, `crates/bacnet-transport/src/sc/mod.rs:461-468`); Result relay accepts only syntactically valid Encapsulated-NPDU Results with a unicast destination and no originating VMAC, and filters other functions (`crates/bacnet-transport/src/sc_hub/relay.rs:121-211`); Encapsulated-NPDU NAKs leave the connection open (`crates/bacnet-transport/src/sc/connection.rs:272-297`).

Evidence anchors are consistent across the JSON ledger, human ledger, PICS, and architecture documentation (`docs/conformance/bacnet-135-2020.json:1119-1241`, `docs/conformance/standard-135-2020-ledger.md:135-137`, `docs/conformance/pics-draft.md:22-24`, `docs/architecture.md:86`). The TLS/WebSocket test exercises the exact `0x62` empty-Header-Data marker and routes the NAK to the source (`benchmarks/tests/sc_websocket.rs:186-244`); the malicious other-function Result filter is pinned (`benchmarks/tests/sc_websocket.rs:247-268`).

Residual explicit nonclaims: correlation of permitted Encapsulated-NPDU Results, direct-connection routing, other BVLC-SC function semantics, and Every Segment behavior (`docs/conformance/bacnet-135-2020.json:1204,1241`).

jscott3201 (commented):
[Adversary] READY

No material #191 findings. I verified `8d8466e8ca170ba8447a7bd13da7754f19a4211c...3f1554b0fd91013cc6bf8bc9c6a1b617d00109de`.

The hub drops malformed and non-Encapsulated-NPDU Results before routing (`crates/bacnet-transport/src/sc_hub/relay.rs:121-152`), rejects spoofed originating VMACs and injects the registered sender VMAC (`:138-158`), and leaves unknown/closed/oversize targets without changing target state (`:179-211`). Encapsulated-NPDU NAKs are nonfatal to the recipient (`crates/bacnet-transport/src/sc/connection.rs:272-295`). The relay copies the validated raw option/payload body while changing only VMAC framing (`crates/bacnet-transport/src/sc_hub/relay.rs:64-93`), and the destination derives the rejection marker from that raw wire representation (`crates/bacnet-transport/src/sc_frame.rs:385-426`; `crates/bacnet-transport/src/sc/mod.rs:418-426`). Broadcast Results are dropped and Result-on-Result responses are not generated.

Residual risk (not a #191 blocker): the pre-existing hub heartbeat path can remove a client when any shared WebSocket sink remains backpressured for its timeout (`crates/bacnet-transport/src/sc_hub/connection.rs:115-128`). The new Result relay itself only logs a timed-out/failed target send and does not remove or close that target (`crates/bacnet-transport/src/sc_hub/relay.rs:203-211`).

jscott3201 (commented):
[Runtime] NOT READY

### Actionable finding

- **[P1] Bound target-sink lock acquisition in the Result relay.** `crates/bacnet-transport/src/sc_hub/relay.rs:196` awaits `sink.lock()` before the five-second timeout begins at `:203-207`; `handle_client` directly awaits this function at `crates/bacnet-transport/src/sc_hub.rs:486-494`. Concrete failure: a connected target that stops reading can make one Result relay hold its sink for the full send timeout while other valid Encapsulated-NPDU Results queue on the mutex. Each waiter then gets another five-second send window, so with many registered sources the later source handlers can be stalled for minutes rather than five seconds. While queued, those handlers cannot process input or observe source replacement/closure (`close_requested` is only rechecked after the lock at `relay.rs:197`). An already-stuck unicast writer can make the wait unbounded (`sc_hub.rs:659-665`). This can prevent the required rejection Result from completing and pin otherwise healthy source connection tasks. The timeout needs to cover lock acquisition and send; add a contended/slow-target test, including source replacement while waiting.

### Verified evidence

- The live path is present: raw A sends the exact `0x62` marker, the hub preserves it, production `ScTransport` B emits the NAK, and A receives the relayed Result (`benchmarks/tests/sc_websocket.rs:185-243`).
- `encode_hub_relay_frame` computes the body after inbound VMACs and rebuilds unicast/broadcast control and VMAC fields while copying option/payload bytes (`crates/bacnet-transport/src/sc_hub/relay.rs:64-95`); the unicast and broadcast length deltas are respectively 0 and +6 bytes.
- Peer Results are syntactically decoded and restricted to Encapsulated-NPDU/unicast routing (`relay.rs:121-152`), and Encapsulated-NPDU NAKs remain nonfatal (`crates/bacnet-transport/src/sc/connection.rs:288-290`). Head CI test, rustfmt, and clippy checks succeeded.

### Residual nonblocking gaps

- Result/request correlation remains explicitly unimplemented, so any registered peer can still inject a valid Encapsulated-NPDU Result to another VMAC; the recipient now treats it as nonfatal. This is accurately recorded at `docs/conformance/bacnet-135-2020.json:1204`.
- Every Segment semantics, other BVLC-SC functions, and direct-connection routing remain explicit nonclaims.

jscott3201 (commented):
[Runtime] NOT READY

- **[P1] Preserve `CloseSource` when the combined relay timeout wins.** The new timeout correctly covers both `sink.lock()` and `send` (`crates/bacnet-transport/src/sc_hub/relay.rs:196-213`), and closure detected after acquiring the lock is propagated at `:198-202`. However, the timeout branch at `:214-218` always returns `Continue` without rechecking `close_requested`; a source replacement can also occur during the awaited send after the sole inner source check. Concrete sequence: old source B is waiting on a congested target sink, replacement B' registers and sets B's `closed`/`close_requested` flag (`crates/bacnet-transport/src/sc_hub.rs:327-345`), and the replacement close task calls `notify_waiters` while B is still inside `relay_result` (`:394-403`). If the relay times out, it returns `Continue`; that notification need not remain pending, so an old peer that does not complete the WebSocket close handshake can leave B's handler waiting for input indefinitely instead of taking `CloseSource`. Recheck `close_requested` after every awaited relay outcome (including timeout/send completion), and cover replacement during lock wait/send with a test.

The original unbounded lock-wait finding itself is fixed: lock acquisition and send now share one five-second timeout, and target closure still maps to `Continue`.

jscott3201 (commented):
[Runtime] READY

No actionable issues remain in the timeout/replacement delta. `crates/bacnet-transport/src/sc_hub/relay.rs:196-213` bounds target-sink lock acquisition and send within the same five-second timeout. Lines 214-225 now recheck `close_requested` after both successful completion and timeout, preserving `CloseSource` when replacement occurs during either await; target closure continues to return `Continue` without closing the healthy source.

PR #313, opened 2026-08-14
fix(services): validate property value priority before narrowing

jscott3201:
## Summary
- validate `BACnetPropertyValue` priority against `1..=16` as the full decoded Unsigned value
- prevent over-wide values such as `257` and `272` from aliasing valid priorities `1` and `16`
- preserve receiver tolerance for a leading-zero representation of numeric priority `1`

This shared decoder is used by WritePropertyMultiple, CreateObject initial values, and COV notification values.

## Validation

Refs #309

### DIFF
--- a/crates/bacnet-services/src/common.rs
+++ b/crates/bacnet-services/src/common.rs
@@ -171,14 +171,16 @@ impl BACnetPropertyValue {
                         "BACnetPropertyValue truncated at priority",
                     ));
                 }
-                let prio = primitives::decode_unsigned(&data[new_pos..end])? as u8;
+                let prio = primitives::decode_unsigned(&data[new_pos..end])?;
                 if !(1..=16).contains(&prio) {
                     return Err(Error::decoding(
                         new_pos,
                         format!("BACnetPropertyValue priority {prio} out of range 1-16"),
                     ));
                 }
-                priority = Some(prio);
+                priority = Some(u8::try_from(prio).map_err(|_| {
+                    Error::decoding(new_pos, "BACnetPropertyValue priority conversion failed")
+                })?);
                 return Ok((
                     Self {
                         property_identifier: PropertyIdentifier::from_raw(prop_id),
@@ -265,18 +267,29 @@ mod tests {
             property_identifier: PropertyIdentifier::PRESENT_VALUE,
             property_array_index: None,
             value: vec![0x10], // app boolean true
-            priority: Some(8),
+            priority: None,
         };
-        let mut buf = BytesMut::new();
-        pv.encode(&mut buf);
+        let mut base = BytesMut::new();
+        pv.encode(&mut base);
 
-        // Manually corrupt priority to 0
-        let data = buf.to_vec();
-        let mut corrupted = data.clone();
-        // Priority is the last encoded byte — find and change it
-        let last = corrupted.len() - 1;
-        corrupted[last] = 0; // set priority value to 0
-        assert!(BACnetPropertyValue::decode(&corrupted, 0).is_err());
+        for priority in [0, 17, 257, 272, u64::MAX] {
+            let mut encoded = base.clone();
+            primitives::encode_ctx_unsigned(&mut encoded, 3, priority);
+            let error = BACnetPropertyValue::decode(&encoded, 0).unwrap_err();
+            assert!(
+                error
+                    .to_string()
+                    .contains(&format!("priority {priority} out of range 1-16")),
+                "unexpected error for priority {priority}: {error}"
+            );
+        }
+
+        let mut leading_zero = base;
+        // Context tag 3 with a two-octet leading-zero encoding of numeric 1.
+        leading_zero.extend_from_slice(&[0x3A, 0x00, 0x01]);
+        let (decoded, consumed) = BACnetPropertyValue::decode(&leading_zero, 0).unwrap();
+        assert_eq!(decoded.priority, Some(1));
+        assert_eq!(consumed, leading_zero.len());
     }
 
     // -----------------------------------------------------------------------


jscott3201 (commented):
[Dataflow] READY

Reviewed exactly `3847e8684aaa47f9f6ab875ecace6a0f48738d41...fa5171117104c9bce8bf2107c6cb50450ecf3c43`; the PR base/head match the review packet. No material findings.

**Evidence**
- `crates/bacnet-services/src/common.rs:162-192` decodes the complete context-[3] Unsigned as `u64`, rejects values outside `1..=16`, and only then performs the lossless `u8` narrowing. Thus wire values `257` and `272` cannot alias priorities 1 and 16.
- Optional priority remains `None` on the no-[3] path (`common.rs:162-164,196-203`). Every accepted value is within `u8` range and preserves its numeric value (`common.rs:174-183`); the valid leading-zero representation of 1 is specifically covered at `common.rs:287-292`.
- The direct decoders all delegate to this one boundary: WPM (`wpm.rs:81-88`), CreateObject initial values (`object_mgmt.rs:119-125`), and COV notification values (`cov.rs:364-371`). No downstream narrowing or alternate decode path was found in those consumers, so common-level coverage is sufficient for this one-file behavioral change.
- Rejection occurs before server mutation: `handle_write_property_multiple_inner` first calls `WritePropertyMultipleRequest::decode` (`crates/bacnet-server/src/handlers/write_property.rs:226-232`); property/value validation follows (`:243-273`); only then does the commit loop begin (`:275-286`). An invalid priority therefore returns from the shared request decode before `db.get_mut` or `object.write_property`.
- The changed unit test rejects 0, 17, 257, 272, and `u64::MAX` (`common.rs:265-285`), while existing round trips cover omitted priority (`:237-247`) and an ordinary valid priority (`:251-261`).

**Non-blocking coverage note**
There is no new wire-to-server WPM regression that asserts a 257/272 priority leaves an already-populated object unchanged. The ordering above establishes that invariant, and no WPM handler code changed; this is not a merge blocker for the one-file slice.

**Commands / results**
- `github_pull_request_read(get, #313)`: base and head exactly match the packet.
- `github_get_commit(full_patch, fa517111...)`: one changed file, `crates/bacnet-services/src/common.rs`.
- Read shared decoder plus direct consumers and WPM handler at the reviewed head (paths/lines above).
- No local test command was run: this read-only review environment provides no command-execution tool. CI at review time had Rustfmt, secret scan, and file-size checks passing; test and Clippy runs were still in progress.

jscott3201 (commented):
[Adversary] READY

No shipping-impacting defect found in `3847e8684aaa47f9f6ab875ecace6a0f48738d41...fa5171117104c9bce8bf2107c6cb50450ecf3c43`.

Evidence:
- `crates/bacnet-services/src/common.rs:174-183` decodes the complete priority as `u64`, rejects any decoded value outside `1..=16`, and only then narrows it to `u8`. Thus wire `257` (`[3] 0x01 0x01`), `272` (`[3] 0x01 0x10`), and `u64::MAX` cannot alias priorities 1 or 16.
- `crates/bacnet-encoding/src/primitives/mod.rs:65-77` rejects zero-length and >8-octet unsigned encodings before decoding. `crates/bacnet-services/src/common.rs:167-173` rejects a declared priority length that exceeds the remaining buffer; `crates/bacnet-encoding/src/tags.rs:217-253` rejects truncated extended lengths and enforces the tag-length sanity limit. No panic/allocation path was found for those malformed priority lengths.
- `crates/bacnet-services/src/common.rs:275-292` covers 0, 17, 257, 272, `u64::MAX`, and the intentionally tolerated non-minimal `[0x3A, 0x00, 0x01]` form. The latter decodes numerically to 1, then passes the same range check; it is not a narrowing alias.
- The shared decoder is used by the WPM, CreateObject, and COV list loops (`crates/bacnet-services/src/wpm.rs:73-89`, `crates/bacnet-services/src/object_mgmt.rs:103-126`, `crates/bacnet-services/src/cov.rs:350-372`), each propagating its decode errors.

Residual risk: as explicitly intended, any 1-to-8-octet leading-zero encoding whose numeric value is 1..=16 remains accepted. Nine-plus-octet forms are rejected by `decode_unsigned`.

Reproducer command: `cargo test -p bacnet-services --locked bacnet_property_value_priority_validation`. The head commit's GitHub `Test (ubuntu-latest)` check completed successfully; static trace of the targeted test matches the outcomes above.

jscott3201 (commented):
[Runtime] READY

Reviewed exact target `3847e8684aaa47f9f6ab875ecace6a0f48738d41...fa5171117104c9bce8bf2107c6cb50450ecf3c43`; base and head were reverified immediately before submission.

No material correctness, runtime, concurrency, or resource-use defects found.

Evidence:
- `crates/bacnet-services/src/common.rs:167-183` bounds-checks the priority content, decodes the complete `u64`, validates numeric `1..=16`, and only then narrows to `u8`. Thus hostile values `257` and `272` cannot alias priorities `1` and `16`.
- `crates/bacnet-encoding/src/primitives/mod.rs:65-77` accepts 1-8 content octets and preserves numeric meaning for leading-zero encodings; malformed empty or over-eight-octet values fail rather than narrow.
- `crates/bacnet-services/src/common.rs:275-292` exercises both range boundaries, the actual low-byte alias values (`257`, `272`), `u64::MAX`, a valid leading-zero encoding, and exact consumed length.
- The shared callers propagate decode failure before appending the value: `crates/bacnet-services/src/wpm.rs:86-88`, `crates/bacnet-services/src/object_mgmt.rs:123-125`, and `crates/bacnet-services/src/cov.rs:369-371`.

Validation: GitHub CI at the reviewed head reports success for Ubuntu tests, Clippy, Rustfmt, file-size, and secret-scan checks. No local commands were run because this review environment exposed no command-execution tool. No material test-coverage gap remains; caller-level hostile-wire tests would be defense in depth, while the shared decoder test exercises the real failure mode.

PR #314, opened 2026-08-14
fix(encoding): validate ErrorPDU enums before narrowing

jscott3201:
## Summary
- decode ErrorPDU Error Class and Error Code as full-width Unsigned values before converting to `u16`
- reject over-wide values such as `65,536` and `65,537` instead of aliasing standard errors
- preserve the valid `u16::MAX` proprietary boundary and wider leading-zero representations

This removes the final two direct `decode_unsigned(...)? as u8/u16/u32` sites from `bacnet-encoding`.

## Validation

Refs #309

### DIFF
--- a/crates/bacnet-encoding/src/apdu/mod.rs
+++ b/crates/bacnet-encoding/src/apdu/mod.rs
@@ -569,7 +569,13 @@ fn decode_error(data: Bytes) -> Result<ErrorPdu, Error> {
             "ErrorPDU truncated at error class",
         ));
     }
-    let error_class_raw = primitives::decode_unsigned(&data[tag_end..class_end])? as u16;
+    let error_class_raw = primitives::decode_unsigned(&data[tag_end..class_end])?;
+    let error_class_raw = u16::try_from(error_class_raw).map_err(|_| {
+        Error::decoding(
+            tag_end,
+            format!("ErrorPDU error class {error_class_raw} exceeds u16"),
+        )
+    })?;
     offset = class_end;
 
     let (tag, tag_end) = tags::decode_tag(&data, offset)?;
@@ -585,7 +591,13 @@ fn decode_error(data: Bytes) -> Result<ErrorPdu, Error> {
     if code_end > data.len() {
         return Err(Error::decoding(tag_end, "ErrorPDU truncated at error code"));
     }
-    let error_code_raw = primitives::decode_unsigned(&data[tag_end..code_end])? as u16;
+    let error_code_raw = primitives::decode_unsigned(&data[tag_end..code_end])?;
+    let error_code_raw = u16::try_from(error_code_raw).map_err(|_| {
+        Error::decoding(
+            tag_end,
+            format!("ErrorPDU error code {error_code_raw} exceeds u16"),
+        )
+    })?;
     offset = code_end;
 
     let error_data = if offset < data.len() {
--- a/crates/bacnet-encoding/src/apdu/tests.rs
+++ b/crates/bacnet-encoding/src/apdu/tests.rs
@@ -306,6 +306,49 @@ fn error_with_trailing_data_round_trip() {
     assert_eq!(apdu, decoded);
 }
 
+#[test]
+fn error_enumerated_values_must_fit_u16() {
+    let encode_error = |error_class, error_code| {
+        let mut buf = BytesMut::with_capacity(16);
+        buf.put_u8(0x50);
+        buf.put_u8(1);
+        buf.put_u8(ConfirmedServiceChoice::READ_PROPERTY.to_raw());
+        primitives::encode_app_enumerated(&mut buf, error_class);
+        primitives::encode_app_enumerated(&mut buf, error_code);
+        buf.freeze()
+    };
+
+    for (error_class, error_code, field, value) in
+        [(65_536, 0, "class", 65_536), (0, 65_537, "code", 65_537)]
+    {
+        let error = decode_apdu(encode_error(error_class, error_code)).unwrap_err();
+        assert!(
+            error
+                .to_string()
+                .contains(&format!("ErrorPDU error {field} {value}")),
+            "unexpected error for error {field} {value}: {error}"
+        );
+    }
+
+    let decoded = decode_apdu(encode_error(u16::MAX as u32, u16::MAX as u32)).unwrap();
+    let Apdu::Error(decoded) = decoded else {
+        panic!("expected ErrorPDU");
+    };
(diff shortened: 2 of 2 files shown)


jscott3201 (commented):
[Dataflow] READY

No material defects found in `e62f40c...3e174ce`.

Evidence:
- `decode_error` now decodes each Enumerated as `u64` and performs a fallible `u16` conversion before constructing `ErrorPdu` (`crates/bacnet-encoding/src/apdu/mod.rs:571-579,593-601,610-616`). Thus class/code values above 65535 fail APDU decoding rather than aliasing after narrowing.
- The client only calls `dispatch_apdu` on `Ok(decoded)` (`crates/bacnet-client/src/client/lifecycle.rs:127-146`); its Error arm completes the TSM only after that gate (`crates/bacnet-client/src/client/dispatch.rs:38-49`). The server has the same decode-before-dispatch gate (`crates/bacnet-server/src/server/lifecycle.rs:79-84`) and only records Error results in its dispatch arm (`crates/bacnet-server/src/server/dispatch.rs:141-158`). Malformed ErrorPDUs therefore cannot advance either transaction state machine.
- Error enums are `u16` newtypes (`crates/bacnet-types/src/enums/protocol.rs:79-95`), and `from_raw` preserves every in-range raw value (`crates/bacnet-types/src/enums/mod.rs:36-46`), including unknown/proprietary values. Trailing service-specific bytes remain sliced after the two fields (`crates/bacnet-encoding/src/apdu/mod.rs:603-607`).
- APDU regression coverage rejects an over-wide class and code independently, retains the `u16::MAX` proprietary boundary, and accepts non-minimal leading-zero forms (`crates/bacnet-encoding/src/apdu/tests.rs:310-351`); the existing trailing-data round trip covers preservation (`:295-307`). This is sufficient coverage for the decoder change.

Non-blocking optional coverage: a distinct unknown in-range value (for example `0x1234`) could be asserted separately from the `u16::MAX` boundary, and a client lifecycle/TSM test could assert an invalid wire ErrorPDU leaves a live transaction pending. The current decode gates provide the required behavior.

jscott3201 (commented):
[Adversary] READY

Reviewed exactly `e62f40cbc6f0f559ebd34a5c92c1ea990773ede1...3e174ceca5b3139d39eecb5677a4f9e03940003e`; the PR still reports those base/head SHAs. No material defect found.

Evidence:
- `crates/bacnet-encoding/src/apdu/mod.rs:572-579` and `:594-600` decode each ENUMERATED as `u64` and reject `u16::try_from` failures before `ErrorClass::from_raw` / `ErrorCode::from_raw` at `:609-614`. Thus 65,536→0, 65,537→1, every `u32`-exceeding value through `u64::MAX`, and their leading-zero representations cannot alias into a dispatched ErrorPDU.
- `crates/bacnet-encoding/src/primitives/mod.rs:65-77` rejects zero-length and >8-octet integer contents. `crates/bacnet-encoding/src/apdu/mod.rs:563-571,588-593` bounds-check tag-derived slices before decoding; `crates/bacnet-encoding/src/tags.rs:217-253` rejects truncated/excessive tag lengths. No panic or large allocation path was found for malformed tags.
- Non-minimal but representable 1–8-octet forms remain valid by design; the three-octet regression vector is exercised at `crates/bacnet-encoding/src/apdu/tests.rs:340-349`.
- Decode errors occur before dispatch: `crates/bacnet-client/src/client/lifecycle.rs:127-147` and `crates/bacnet-server/src/server/lifecycle.rs:79-80,333-363` dispatch only the `Ok(decoded)` branch.

Adversarial vectors traced: class 65536 (`50 01 0c 93 01 00 00 91 00`), code 65537 (`50 01 0c 91 00 93 01 00 01`), class `u64::MAX` (`50 01 0c 95 08 ff ff ff ff ff ff ff ff 91 00`), zero length (`50 01 0c 90 91 00`), 9-octet content, and leading-zero values. All reach an error before `Apdu::Error` construction except the representable leading-zero form, which correctly constructs its original value.

Validation: source-level trace completed. Local commands were not executable in this read-only review environment. GitHub Rustfmt and Clippy checks are successful; the Test check was still in progress when reviewed. Suggested repro: `cargo test -p bacnet-encoding --lib apdu::tests::error_enumerated_values_must_fit_u16 --locked`.

jscott3201 (commented):
[Runtime] READY

Reviewed exactly `e62f40cbc6f0f559ebd34a5c92c1ea990773ede1...3e174ceca5b3139d39eecb5677a4f9e03940003e`. I found no material runtime defect.

Evidence:
- `crates/bacnet-encoding/src/apdu/mod.rs:563-579` and `:588-601` compute and bounds-check each field end, decode the complete unsigned value, and only then perform checked `u16::try_from`; `65_536`/`65_537` therefore cannot alias lower standard values.
- Conversion failures use `tag_end`, the first content octet, as the decode offset (`apdu/mod.rs:573-578`, `:595-600`). `tags::decode_tag` advances past compact or extended headers before returning that offset and caps declared lengths (`crates/bacnet-encoding/src/tags.rs:163-264`).
- `ErrorClass` and `ErrorCode` are intentionally `u16` newtypes (`crates/bacnet-types/src/enums/protocol.rs:79-95`), so `u16::MAX` remains representable. `decode_unsigned` accumulates 1-8 big-endian octets without requiring minimal width (`crates/bacnet-encoding/src/primitives/mod.rs:65-77`), preserving accepted leading-zero encodings.
- Trailing service-specific bytes still begin at the full error-code `code_end` and are sliced unchanged (`apdu/mod.rs:601-607`). The existing trailing-data round trip covers this path (`crates/bacnet-encoding/src/apdu/tests.rs:294-306`).
- The regression test constructs real ErrorPDUs through the production Enumerated encoder, exercises both original wraparound cases, accepts both `u16::MAX` fields, and directly exercises three-octet leading-zero wire encodings (`apdu/tests.rs:309-350`). These cases would fail against the base casts.

Missing coverage (non-blocking): the new test does not assert the exact `Error::Decoding` offsets/variant or exercise 4-8-octet leading-zero/over-wide values. The inspected tag/unsigned paths make those cases follow the same checked conversion, so I do not consider this merge-blocking.

Head CI results: `cargo fmt --all --check`, `cargo clippy --workspace --exclude rusty-bacnet --all-targets --locked`, `cargo test --workspace --exclude rusty-bacnet --locked --features bacnet-types/serde,bacnet-transport/ipv6,bacnet-transport/sc-tls,bacnet-transport/serial,bacnet-transport/ethernet`, and both repository guard scripts completed successfully.

PR #315, opened 2026-08-14
fix(services): validate RPM error enums before narrowing

jscott3201:
## Summary
- reject ReadPropertyMultiple error-class and error-code values above `u16::MAX`
- require both values to use application-tagged Enumerated encoding
- preserve valid leading-zero encodings and the `u16::MAX` boundary
- verify the decoder consumes the closing property-result tag

Part of #309.

## Validation
- `cargo test -p bacnet-services --locked rpm::tests::error_pair_values_must_be_enumerated_and_fit_u16`
- `git diff --check`

### DIFF
--- a/crates/bacnet-services/src/rpm.rs
+++ b/crates/bacnet-services/src/rpm.rs
@@ -289,6 +289,12 @@ impl ReadPropertyMultipleACK {
 fn decode_error_pair(data: &[u8], offset: usize) -> Result<(ErrorClass, ErrorCode, usize), Error> {
     // error-class: app-tagged enumerated
     let (tag, pos) = tags::decode_tag(data, offset)?;
+    if tag.class != tags::TagClass::Application || tag.number != tags::app_tag::ENUMERATED {
+        return Err(Error::decoding(
+            offset,
+            "RPM error class: expected application-tagged enumerated",
+        ));
+    }
     let end = pos + tag.length as usize;
     if end > data.len() {
         return Err(Error::decoding(pos, "RPM error truncated at error-class"));
@@ -305,6 +311,12 @@ fn decode_error_pair(data: &[u8], offset: usize) -> Result<(ErrorClass, ErrorCod
 
     // error-code: app-tagged enumerated
     let (tag, pos) = tags::decode_tag(data, offset)?;
+    if tag.class != tags::TagClass::Application || tag.number != tags::app_tag::ENUMERATED {
+        return Err(Error::decoding(
+            offset,
+            "RPM error code: expected application-tagged enumerated",
+        ));
+    }
     let end = pos + tag.length as usize;
     if end > data.len() {
         return Err(Error::decoding(pos, "RPM error truncated at error-code"));
@@ -426,7 +438,7 @@ mod tests {
     }
 
     #[test]
-    fn error_pair_values_must_fit_u16() {
+    fn error_pair_values_must_be_enumerated_and_fit_u16() {
         let encode_pair = |error_class, error_code| {
             let mut buf = BytesMut::new();
             primitives::encode_app_enumerated(&mut buf, error_class);
@@ -460,6 +472,18 @@ mod tests {
         assert_eq!(error_class, ErrorClass::PROPERTY);
         assert_eq!(error_code, ErrorCode::UNKNOWN_PROPERTY);
         assert_eq!(consumed, leading_zero.len());
+
+        for (tag_offset, field) in [(0, "class"), (2, "code")] {
+            let mut encoded = encode_pair(2, 32);
+            encoded[tag_offset] = 0x21; // Application Unsigned with one content octet.
+            let error = decode_error_pair(&encoded, 0).unwrap_err();
+            assert!(
+                error.to_string().contains(&format!(
+                    "RPM error {field}: expected application-tagged enumerated"
+                )),
+                "unexpected error for error {field} tag: {error}"
+            );
+        }
     }
 
     // -----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


jscott3201 (commented):
[Dataflow] READY

No material findings. `decode_error_pair` preserves the decoded `u64` until the checked `u16::try_from` succeeds for both ErrorClass and ErrorCode (`crates/bacnet-services/src/rpm.rs:296-316`); both indexed and non-indexed property-access-error paths call that helper and advance to its returned post-closing-tag offset (`:237-246`, `:260-269`, `:319-325`). The boundary and leading-zero cases are covered at `:428-463`.

jscott3201 (commented):
[Runtime] READY

No material findings.

Evidence:
- Both RPM error-result branches call `decode_error_pair`: the optional array-index path at `crates/bacnet-services/src/rpm.rs:237-246`, and the no-array-index path at `crates/bacnet-services/src/rpm.rs:260-269`.
- `decode_error_pair` decodes to `u64` and applies checked `u16::try_from` before constructing either newtype at `crates/bacnet-services/src/rpm.rs:296-316`; values above 65,535 therefore return `Error::Decoding` at the absolute content offset rather than wrapping. The underlying decoder accepts 1–8 big-endian octets (`crates/bacnet-encoding/src/primitives/mod.rs:65-77`), while `ErrorClass` and `ErrorCode` are `u16` wrappers (`crates/bacnet-types/src/enums/protocol.rs:79-95`).
- The helper validates closing tag 5 and returns `tag_end`, consuming it, at `crates/bacnet-services/src/rpm.rs:319-325`.
- `error_pair_values_must_fit_u16` covers an oversized class, an oversized code, the exact `u16::MAX` accepted boundary, leading-zero encodings, and closing-tag consumption at `crates/bacnet-services/src/rpm.rs:429-462`.

Residual test risk: the new test invokes the helper directly rather than sending oversized values through `ReadPropertyMultipleACK::decode`, and it does not assert the `Error::Decoding` variant/offset or exercise the optional array-index branch. The production call sites are direct and unchanged, so this is non-blocking coverage debt rather than a runtime defect.

Validation: no local commands were run because this environment exposed no command-execution tool. GitHub's `Test (ubuntu-latest)`, `Clippy`, and `Rustfmt` checks completed successfully for head `094249fb99a403513a5ae7d5ee7c34a34cc3a64c`.

jscott3201 (commented):
[Adversary] NOT READY

P2 — `decode_error_pair` still accepts non-Enumerated (and context-tagged) inner values as a valid property-access-error. At `crates/bacnet-services/src/rpm.rs:291-316`, both headers are decoded only to obtain a length; neither is checked for Application class plus `app_tag::ENUMERATED`. Thus the well-formed but type-invalid RPM-ACK payload `0c 00 00 00 00 1e 29 55 5e 21 02 21 20 5f 1f` is accepted: the two `0x21` Application Unsigned tags are interpreted as ErrorClass::PROPERTY (2) and ErrorCode::UNKNOWN_PROPERTY (32), rather than being rejected. Valid Enumerated headers would be `0x91` (`tags.rs:15-19,22-36`; `rpm.rs:155-158`). This is silent type aliasing of malformed network input.

The flaw predates this PR (the changed range-narrowing lines do not add the missing header checks), but it remains in the function this PR modifies and its new tests (`rpm.rs:429-463`) exercise only correctly tagged Enumerated inputs. Please validate both tag class/number before decoding their contents and add malformed-tag coverage.

jscott3201 (commented):
[Adversary] READY

Reviewed exactly `16752d72c9710da5e79cac8ba7992357c4e0f97b...bb082b0663500e50f6ed392d3418ef8c5d9c0c25`. No material findings. `crates/bacnet-services/src/rpm.rs:291-321` now rejects both non-Application and non-Enumerated inner error tags before using their lengths, then preserves the checked `u16` narrowing and closing-tag consumption. The tests at `rpm.rs:438-487` cover out-of-range values, the u16 boundary, leading-zero values, and wrong Application tag type for both fields.

jscott3201 (commented):
[Dataflow] READY

No material findings in `16752d72c9710da5e79cac8ba7992357c4e0f97b...bb082b0663500e50f6ed392d3418ef8c5d9c0c25`. `decode_error_pair` rejects a non-application/Enumerated tag before consuming payload, preserves the raw decoded value until checked `u16` narrowing, and returns the offset after the verified closing tag (`crates/bacnet-services/src/rpm.rs:291-337`). Both indexed and non-indexed ACK paths use this helper (`:237-246`, `:260-269`); boundary, leading-zero, and wrong-tag cases are covered (`:441-486`).

jscott3201 (commented):
[Runtime] READY

No material findings on `16752d72c9710da5e79cac8ba7992357c4e0f97b...bb082b0663500e50f6ed392d3418ef8c5d9c0c25`.

Evidence:
- `decode_error_pair` now requires both the application class and Enumerated tag number before reading either payload (`crates/bacnet-services/src/rpm.rs:291-297` and `:313-319`). These checks correspond to `TagClass::Application` and application tag 9 in `crates/bacnet-encoding/src/tags.rs:12-19` and `:22-34`.
- Checked narrowing still occurs only after `decode_unsigned` returns its `u64`, at `crates/bacnet-services/src/rpm.rs:302-328`; `ErrorClass` and `ErrorCode` are `u16` wrappers (`crates/bacnet-types/src/enums/protocol.rs:79-95`).
- Both ACK branches still route error results through this helper (`crates/bacnet-services/src/rpm.rs:237-246` and `:260-269`), and closing tag 5 is consumed via returned `tag_end` (`:331-337`).
- The updated regression rejects an application Unsigned tag for both class and code while retaining oversized-value, exact-boundary, leading-zero, and closing-tag-consumption cases (`crates/bacnet-services/src/rpm.rs:441-487`).

Residual coverage gaps (non-blocking): the wrong-tag regression exercises tag-number mismatch but not a context-class mismatch; it invokes the helper directly rather than full `ReadPropertyMultipleACK::decode`, does not cover the optional array-index branch, and does not assert exact `Error::Decoding` offsets.

Validation: no local commands were run because no command-execution tool was available. GitHub's `Test (ubuntu-latest)`, `Clippy`, and `Rustfmt` checks passed for head `bb082b0663500e50f6ed392d3418ef8c5d9c0c25`.

PR #316, opened 2026-08-14
fix(services): validate WriteGroup unsigned widths

jscott3201:
## Summary
- reject WriteGroup group numbers above `u32::MAX`
- validate write priority as the full decoded value before converting to `u8`
- reject override priorities above `u8::MAX`
- preserve exact maxima and wider leading-zero encodings

Part of #309.

## Validation
- `cargo test -p bacnet-services --locked write_group`
- `git diff --check`

### DIFF
--- a/crates/bacnet-services/src/write_group.rs
+++ b/crates/bacnet-services/src/write_group.rs
@@ -79,7 +79,13 @@ impl WriteGroupRequest {
         if end > data.len() {
             return Err(Error::decoding(pos, "WriteGroup truncated at group-number"));
         }
-        let group_number = primitives::decode_unsigned(&data[pos..end])? as u32;
+        let group_number_raw = primitives::decode_unsigned(&data[pos..end])?;
+        let group_number = u32::try_from(group_number_raw).map_err(|_| {
+            Error::decoding(
+                pos,
+                format!("WriteGroup group number {group_number_raw} exceeds u32"),
+            )
+        })?;
         if group_number == 0 {
             return Err(Error::Encoding(
                 "WriteGroup group number 0 is reserved".into(),
@@ -96,13 +102,19 @@ impl WriteGroupRequest {
                 "WriteGroup truncated at write-priority",
             ));
         }
-        let write_priority = primitives::decode_unsigned(&data[pos..end])? as u8;
-        if !(1..=16).contains(&write_priority) {
+        let write_priority_raw = primitives::decode_unsigned(&data[pos..end])?;
+        if !(1..=16).contains(&write_priority_raw) {
             return Err(Error::decoding(
                 pos,
-                format!("WriteGroup write-priority {write_priority} out of range 1-16"),
+                format!("WriteGroup write-priority {write_priority_raw} out of range 1-16"),
             ));
         }
+        let write_priority = u8::try_from(write_priority_raw).map_err(|_| {
+            Error::decoding(
+                pos,
+                format!("WriteGroup write-priority {write_priority_raw} exceeds u8"),
+            )
+        })?;
         offset = end;
 
         // [2] changeList — opening tag 2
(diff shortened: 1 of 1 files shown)


jscott3201 (commented):
[Dataflow] READY

No material findings.

Verified `be6cdaf82f8692dbb3c1bb8a3559e65993b04ec4...b7361d0eb6e548297cac895ed82787d37269a066`:
- `crates/bacnet-services/src/write_group.rs:79-86` rejects a decoded group number that cannot fit the `u32` field before `WriteGroupRequest` is constructed.
- `crates/bacnet-services/src/write_group.rs:102-117` validates the full decoded `u64` write-priority value before narrowing; values 1–16 reach the typed `u8` only after validation.
- `crates/bacnet-services/src/write_group.rs:156-168` rejects an unrepresentable `override_priority` before assigning the `GroupChannelValue` field. The optional-field offset advances only after a valid conversion, while the error path returns, so later change-list entries cannot be parsed or dispatched after an invalid value.
- `crates/bacnet-services/src/write_group.rs:283-345` covers each overflow case, `u32::MAX`/`u8::MAX`, and non-minimal leading-zero encodings.
- `crates/bacnet-server/src/handlers/write_group.rs:handle_write_group` propagates decode errors, and `crates/bacnet-server/src/server/requests/unconfirmed.rs:BACnetServer::handle_unconfirmed_request` logs and returns on that error, so invalid requests do not reach the successful WriteGroup dispatch arm.
- Encode and public typed fields remain `u32`/`u8` (`crates/bacnet-services/src/write_group.rs:19-37,43-67`); no API surface changed.

Residual coverage: the focused service test does not exercise the server's invalid-unconfirmed-request branch end-to-end. The code path is direct and unchanged; this is non-blocking.

Commands/evidence: GitHub PR metadata, exact head-commit patch, target-version source retrieval, and check-run status were inspected; no local commands or mutations were performed.

jscott3201 (commented):
[Runtime] READY

No material findings in `be6cdaf82f8692dbb3c1bb8a3559e65993b04ec4...b7361d0eb6e548297cac895ed82787d37269a066`.

- `crates/bacnet-services/src/write_group.rs:82-88` decodes `groupNumber` to `u64` and uses checked `u32::try_from`, so values above `u32::MAX` cannot wrap into a valid group.
- `crates/bacnet-services/src/write_group.rs:105-117` applies the protocol range `1..=16` to the full decoded write-priority before narrowing; `257` is therefore rejected rather than wrapping to `1`.
- `crates/bacnet-services/src/write_group.rs:157-167` checks override-priority before storing it. `priority_pos = new_off - content.len()` is the content start because `decode_optional_context` returns `content = data[new_pos..end]` and `new_off = end` (`crates/bacnet-encoding/src/tags.rs:397-416`); the subtraction is input-independent and cannot underflow under that API contract. The offset advances only after successful parsing.
- `crates/bacnet-services/src/write_group.rs:303-345` exercises the three former wraparound cases, exact accepted maxima, and wider leading-zero encodings through the real decoder. The valid leading-zero override followed by the value and closing tags also exercises optional-field offset synchronization.
- `decode_unsigned` accepts only 1-8 octets and accumulates into `u64` without overflowing (`crates/bacnet-encoding/src/primitives/mod.rs:65-77`). No new concurrency, unbounded allocation, or reachable panic path was found.

Residual gap: the new test does not assert the numeric override-priority error offset directly, though the offset derivation is verified from the tag API above. GitHub CI's Ubuntu test and Clippy checks passed on the reviewed head.

jscott3201 (commented):
[Adversary] READY

No material defect found in the authoritative `be6cdaf82f8692dbb3c1bb8a3559e65993b04ec4...b7361d0eb6e548297cac895ed82787d37269a066` diff.

Verified behavior:
- `crates/bacnet-services/src/write_group.rs:82-88` rejects decoded group numbers that cannot fit `u32`, rather than truncating them.
- `crates/bacnet-services/src/write_group.rs:105-117` checks the full decoded write-priority against `1..=16` before narrowing.
- `crates/bacnet-services/src/write_group.rs:159-166` rejects override priorities above `u8::MAX`; the offset is the beginning of the optional value content.
- `crates/bacnet-services/src/write_group.rs:287-345` covers above-width values, exact maxima, and wider leading-zero encodings.

Residual hostile-input gap: none identified within this change's scope.

Validation: inspected the exact PR diff and surrounding tag/unsigned decoder paths; GitHub CI reports Test (ubuntu-latest), Clippy, and Rustfmt successful. No local commands were run.

PR #317, opened 2026-08-14
fix(services): validate Who-Is and I-Am widths

jscott3201:
## Summary
- reject Who-Is low and high limits above `u32::MAX`
- reject I-Am maximum APDU lengths above `u32::MAX`
- reject I-Am segmentation and vendor values above their `u8` and `u16` representations
- require the four I-Am fields to use their specified application tags
- preserve exact maxima and wider leading-zero encodings

Part of #309.

## Validation
- `cargo test -p bacnet-services --locked who_is`
- `git diff --check`

### DIFF
--- a/crates/bacnet-services/src/who_is.rs
+++ b/crates/bacnet-services/src/who_is.rs
@@ -132,8 +132,11 @@ impl IAmRequest {
         let mut offset = 0;
 
         let (tag, pos) = tags::decode_tag(data, offset)?;
+        // Application L/V/T values 6 and 7 are reserved, but decode_tag treats
+        // them as extended lengths because they mark context opening/closing tags.
         if tag.class != tags::TagClass::Application
             || tag.number != tags::app_tag::OBJECT_IDENTIFIER
+            || data[offset] & 0x07 > 5
         {
             return Err(Error::decoding(
                 offset,
@@ -148,7 +151,10 @@ impl IAmRequest {
         offset = end;
 
         let (tag, pos) = tags::decode_tag(data, offset)?;
-        if tag.class != tags::TagClass::Application || tag.number != tags::app_tag::UNSIGNED {
+        if tag.class != tags::TagClass::Application
+            || tag.number != tags::app_tag::UNSIGNED
+            || data[offset] & 0x07 > 5
+        {
             return Err(Error::decoding(
                 offset,
                 "IAm max APDU length: expected application-tagged unsigned",
@@ -168,7 +174,10 @@ impl IAmRequest {
         offset = end;
 
         let (tag, pos) = tags::decode_tag(data, offset)?;
-        if tag.class != tags::TagClass::Application || tag.number != tags::app_tag::ENUMERATED {
+        if tag.class != tags::TagClass::Application
+            || tag.number != tags::app_tag::ENUMERATED
+            || data[offset] & 0x07 > 5
+        {
             return Err(Error::decoding(
                 offset,
                 "IAm segmentation: expected application-tagged enumerated",
@@ -185,7 +194,10 @@ impl IAmRequest {
         offset = end;
 
         let (tag, pos) = tags::decode_tag(data, offset)?;
-        if tag.class != tags::TagClass::Application || tag.number != tags::app_tag::UNSIGNED {
+        if tag.class != tags::TagClass::Application
+            || tag.number != tags::app_tag::UNSIGNED
+            || data[offset] & 0x07 > 5
+        {
             return Err(Error::decoding(
                 offset,
                 "IAm vendor ID: expected application-tagged unsigned",
@@ -444,6 +456,17 @@ mod tests {
                 "unexpected error for {field} tag: {error}"
             );
         }
+
+        let valid = encode_request(1476, 0, 999);
+        let mut reserved_lvt = BytesMut::from(&[0xc6, 4][..]);
+        reserved_lvt.extend_from_slice(&valid[1..]);
+        let error = IAmRequest::decode(&reserved_lvt).unwrap_err();
+        assert!(
+            error
+                .to_string()
+                .contains("IAm object identifier: expected application-tagged"),
+            "unexpected error for reserved application L/V/T: {error}"
+        );
     }
 
     #[test]
--- a/crates/bacnet-services/src/who_is.rs
+++ b/crates/bacnet-services/src/who_is.rs
@@ -61,7 +61,10 @@ impl WhoIsRequest {
             if end > data.len() {
                 return Err(Error::decoding(pos, "WhoIs truncated at low-limit"));
             }
-            low_limit = Some(primitives::decode_unsigned(&data[pos..end])? as u32);
+            let low_limit_raw = primitives::decode_unsigned(&data[pos..end])?;
+            low_limit = Some(u32::try_from(low_limit_raw).map_err(|_| {
+                Error::decoding(pos, format!("WhoIs low-limit {low_limit_raw} exceeds u32"))
+            })?);
             offset = end;
         }
 
@@ -73,7 +76,13 @@ impl WhoIsRequest {
                 if end > data.len() {
                     return Err(Error::decoding(pos, "WhoIs truncated at high-limit"));
                 }
-                high_limit = Some(primitives::decode_unsigned(&data[pos..end])? as u32);
+                let high_limit_raw = primitives::decode_unsigned(&data[pos..end])?;
+                high_limit = Some(u32::try_from(high_limit_raw).map_err(|_| {
+                    Error::decoding(
+                        pos,
+                        format!("WhoIs high-limit {high_limit_raw} exceeds u32"),
+                    )
+                })?);
             }
         }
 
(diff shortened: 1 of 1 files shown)


jscott3201 (commented):
[Dataflow] READY

No material findings.

Verified data flow:
- `decode_unsigned` returns the full 1–8-octet `u64` (`crates/bacnet-encoding/src/primitives/mod.rs:65-78`); the changed `u32`/`u8`/`u16` checked conversions reject values before constructing either public request (`crates/bacnet-services/src/who_is.rs:64-67`, `79-85`, `147-175`).
- A failed Who-Is decode returns from the server before range evaluation or I-Am emission (`crates/bacnet-server/src/server/requests/unconfirmed.rs:38-44`, `55-77`). A failed I-Am decode takes the error arm and cannot upsert/publish a discovered device (`crates/bacnet-client/src/client/dispatch.rs:125-161`).
- Boundary and leading-zero coverage rejects each first out-of-range value and preserves `u32::MAX`, `u8::MAX`, and `u16::MAX` (`crates/bacnet-services/src/who_is.rs:287-363`). The public field types and encode paths remain unchanged (`crates/bacnet-services/src/who_is.rs:19-46`, `116-129`).
- Compatibility: `TryFrom` is available under the workspace MSRV 1.93 (`Cargo.toml:42-45`); all conversions are fixed-width, so no target-width dependency was introduced.

Residual risk: the new vectors are codec-unit tests rather than malformed-packet server/client integration tests; consumer control flow was inspected as above. No shell command runner was exposed in this review environment, so no cargo commands were run.

Inspection operations: verified PR base/head metadata against `c5e0ae0edaa4a688659e7260389fd9a6d36eba82...808dfa3e063072a2a9ebd124132e046f8ab7acac`; read the exact diff and the service, primitive, server-dispatch, client-dispatch, integration-test, enum, and workspace-config sources at the target SHA.

jscott3201 (commented):
[Adversary] NOT READY

### P2 — Validate each required I-Am application tag before using its contents
`crates/bacnet-services/src/who_is.rs:134-175` obtains each tag only to calculate the content slice; it never verifies the tag class or tag number, despite the contract at `who_is.rs:112-114`. Consequently, a malformed I-Am body such as `C4 02 00 04 D2 92 05 C4 21 00 32 03 E7` is accepted: after a valid Device object identifier, it supplies Enumerated (not Unsigned) for `max-apdu-length`, Unsigned (not Enumerated) for `segmentation-supported`, and Signed (not Unsigned) for `vendor-id`. The raw contents still decode to `1476`, `0`, and `999`, so `IAmRequest::decode` returns success. The client caller then upserts that malformed announcement into discovery state (`crates/bacnet-client/src/client/dispatch.rs:124-156`).

This width-validation change should reject incorrect application tag identities as well as oversized contents; require Application tags `OBJECT_IDENTIFIER`, `UNSIGNED`, `ENUMERATED`, and `UNSIGNED` in that order, with a regression test for tag aliases.

jscott3201 (commented):
[Runtime] READY

No material runtime defect found in `c5e0ae0edaa4a688659e7260389fd9a6d36eba82...808dfa3e063072a2a9ebd124132e046f8ab7acac`.

- The checked conversions at `crates/bacnet-services/src/who_is.rs:64-85` and `crates/bacnet-services/src/who_is.rs:147-175` reject peer-controlled numerical overflow before either Who-Is response dispatch (`crates/bacnet-server/src/server/requests/unconfirmed.rs:37-46`) or I-Am device-table mutation (`crates/bacnet-client/src/client/dispatch.rs:124-160`). I found no panic, resource-growth, concurrency, or partial-state-update path introduced by the change.
- The hostile low-word-valid values and wider leading-zero representations are exercised at `crates/bacnet-services/src/who_is.rs:287-363`; existing equal/range and optional-bound behavior remains ordered after successful narrowing.
- GitHub's test, Clippy, and Rustfmt checks succeeded for the reviewed head.

Non-blocking pre-existing follow-up: `crates/bacnet-client/src/client/requests.rs:64-68` still narrows a discovered device's accepted `u32` maximum APDU length with `as u16` before applying the transport cap. For example, an I-Am value of `65_536` wraps to zero, and a later non-empty confirmed request reaches the zero-segment-size error at `crates/bacnet-encoding/src/segmentation.rs:25-43`. This predates and is outside the five decoder conversions in this focused PR, so it does not change the READY decision.

jscott3201 (commented):
[Dataflow] NOT READY

**P1 — application tags using the context-only opening/closing L/V/T forms still pass the new guards and update discovery.** The tag model documents L/V/T 6/7 as *context-specific* opening/closing tags (`crates/bacnet-encoding/src/tags.rs:49-52`), but `decode_tag` only recognizes those forms when `class == Context` (`crates/bacnet-encoding/src/tags.rs:190-215`). For an application tag with L/V/T 6 or 7 it instead takes the extended-length path (`crates/bacnet-encoding/src/tags.rs:217-264`) and returns a normal `Application` tag with the expected number.

Concrete reproducer: encode an otherwise valid I-Am, replace the first canonical Object-Identifier header `0xC4` with `0xC6`, and insert `0x04` immediately after it. This is an application Object-Identifier with the context-only L/V/T=6 form, followed by an extended length of four. It decodes as `Application/Object-Identifier/length=4`, so the new guard at `crates/bacnet-services/src/who_is.rs:134-142` accepts it; the remaining valid fields complete the decode. The client then reaches the `Ok(i_am)` arm and upserts/publishes the forged discovery record (`crates/bacnet-client/src/client/dispatch.rs:123-160`). The same construction applies to each application field.

The added tag tests only change tag numbers through canonical `encode_tag` output (`crates/bacnet-services/src/who_is.rs:390-446`), so they do not cover L/V/T=6/7 application encodings. This leaves malformed I-Am data able to reach discovery despite the new tag guards.

Verified target: `c5e0ae0edaa4a688659e7260389fd9a6d36eba82...a16c0d02570fe34e380110600693d241dbad8e6b`. No commands were run; review used read-only source/diff inspection.

jscott3201 (commented):
[Runtime] READY

No material runtime findings in `c5e0ae0edaa4a688659e7260389fd9a6d36eba82...a16c0d02570fe34e380110600693d241dbad8e6b`.

- All four guards validate both `TagClass::Application` and the required application tag number immediately after `decode_tag` and before content slicing or value decoding (`crates/bacnet-services/src/who_is.rs:134-147`, `150-168`, `170-185`, and `187-201`). A mismatched tag therefore cannot have its payload reinterpreted as the expected I-Am field.
- Rejection still occurs before the only runtime consumer mutates the discovery table: `IAmRequest::decode` must return `Ok` before `crates/bacnet-client/src/client/dispatch.rs:124-149` constructs and upserts a device.
- The table-driven cases at `crates/bacnet-services/src/who_is.rs:390-446` corrupt each field independently while keeping all preceding fields valid, so each guard is actually reached. The canonical round trip at `crates/bacnet-services/src/who_is.rs:237-248` and width/leading-zero cases at `crates/bacnet-services/src/who_is.rs:346-388` continue to exercise accepted encodings.
- The guards add only fixed comparisons and static-string error construction; I found no new panic, allocation-growth, concurrency, or partial-state path.

GitHub Test (ubuntu-latest), Clippy, Rustfmt, file-size, and no-secret checks all succeeded for head `a16c0d02570fe34e380110600693d241dbad8e6b`.

jscott3201 (commented):
[Dataflow] READY

No material findings. The reserved application L/V/T finding is resolved at all four I-Am fields.

Evidence:
- Each field validates application class, its expected tag number, and rejects L/V/T 6/7 (`data[offset] & 0x07 > 5`) before consuming the field: Object-Identifier `crates/bacnet-services/src/who_is.rs:134-150`; max-APDU `153-174`; segmentation `176-194`; vendor ID `196-213`. `decode_tag` has already established that each `offset` addresses a tag header, so these raw-header reads are bounded.
- Thus the former `0xC6, 0x04` Object-Identifier construction errors before `ObjectIdentifier::decode`; the same predicate rejects both 6 and 7 for the remaining three fields. L/V/T 5 remains allowed, preserving extended-length/leading-zero values.
- The regression test constructs that former bypass and asserts decode failure (`crates/bacnet-services/src/who_is.rs:460-469`); existing boundary coverage preserves the u32/u8/u16 maxima.
- A decode error takes the client error arm and cannot upsert or emit a discovery event; construction/upsert is exclusively in `Ok(i_am)` (`crates/bacnet-client/src/client/dispatch.rs:122-161`). Who-Is checked-narrowing errors still return before server filtering/I-Am emission (`crates/bacnet-server/src/server/requests/unconfirmed.rs:37-77`).

Residual risk: the new reserved-L/V/T regression vector directly covers Object-Identifier/LVT=6; LVT=7 and fields 2–4 are covered by the identical per-field predicate on inspection, rather than individual vectors. No shell commands were run in this read-only review environment.

Verified target: `c5e0ae0edaa4a688659e7260389fd9a6d36eba82...f869ca87d37ff58415a44153337e04d9f9c4d5c1`.

jscott3201 (commented):
[Runtime] READY

No material runtime findings in `c5e0ae0edaa4a688659e7260389fd9a6d36eba82...f869ca87d37ff58415a44153337e04d9f9c4d5c1`.

- Each I-Am field now rejects application-tagged headers whose raw L/V/T is reserved 6 or 7, in addition to checking class and tag number (`crates/bacnet-services/src/who_is.rs:134-150`, `153-174`, `176-194`, and `196-213`). Valid extended-length encodings still use raw L/V/T 5, so the wider leading-zero values at `crates/bacnet-services/src/who_is.rs:382-400` remain accepted.
- The raw-header indexing is safe: `decode_tag` first proves `offset < data.len()` at `crates/bacnet-encoding/src/tags.rs:163-171`; every `data[offset]` check occurs only after that call succeeds. Truncated headers return before indexing, and no new panic path is introduced.
- The regression at `crates/bacnet-services/src/who_is.rs:460-469` constructs the concrete `0xC6, 0x04` object-identifier header that `decode_tag` would otherwise interpret through its extended-length path (`crates/bacnet-encoding/src/tags.rs:217-227`) and verifies that I-Am decoding rejects it.
- Width rejection, tag-number rejection, canonical round trips, and mutation ordering remain intact. The added checks are fixed-cost comparisons with no concurrency or resource-growth impact.

GitHub Test (ubuntu-latest), Clippy, Rustfmt, file-size, and no-secret checks all succeeded for head `f869ca87d37ff58415a44153337e04d9f9c4d5c1`.

PR #318, opened 2026-08-14
fix(client): preserve over-wide discovered APDU limits

jscott3201:
## Summary
- convert discovered `u32` maximum APDU lengths to `u16` without modulo wrapping
- cap values above `u16::MAX` before applying the transport limit
- cover the `65,536 -> 0` alias with a loopback confirmed-request regression

## Why
I-Am decoding now preserves the full `u32` maximum APDU length. The client request path still used `as u16`, so a peer advertising 65,536 caused a non-empty request to enter segmentation with a zero-byte segment limit.

## Testing
- `git diff --check`

Related to #309.

### DIFF
--- a/crates/bacnet-client/src/client/requests.rs
+++ b/crates/bacnet-client/src/client/requests.rs
@@ -62,7 +62,7 @@ impl<T: TransportPort + 'static> BACnetClient<T> {
                     let dt = self.device_table.lock().await;
                     let device = dt.get_by_mac(mac);
                     let max_apdu = device
-                        .map(|d| d.max_apdu_length as u16)
+                        .map(|d| u16::try_from(d.max_apdu_length).unwrap_or(u16::MAX))
                         .unwrap_or(self.config.max_apdu_length);
                     let max_seg = device.and_then(|d| d.max_segments_accepted);
                     (max_apdu.min(target_transport_max_apdu), max_seg)
--- a/crates/bacnet-client/src/client/tests.rs
+++ b/crates/bacnet-client/src/client/tests.rs
@@ -302,6 +302,64 @@ async fn confirmed_request_simple_ack() {
 }
 
 #[tokio::test]
+async fn confirmed_request_does_not_wrap_discovered_max_apdu_length() {
+    let client_mac = vec![0x01];
+    let remote_mac = vec![0x02];
+    let (client_transport, remote_transport) =
+        LoopbackTransport::pair(client_mac, remote_mac.clone());
+    let mut remote_network = NetworkLayer::new(remote_transport);
+    let mut remote_rx = remote_network.start().await.unwrap();
+    let mut client = BACnetClient::generic_builder()
+        .transport(client_transport)
+        .build()
+        .await
+        .unwrap();
+
+    client.device_table.lock().await.upsert(DiscoveredDevice {
+        object_identifier: ObjectIdentifier::new(ObjectType::DEVICE, 2002).unwrap(),
+        mac_address: MacAddr::from_slice(&remote_mac),
+        max_apdu_length: u32::from(u16::MAX) + 1,
+        segmentation_supported: Segmentation::NONE,
+        max_segments_accepted: None,
+        vendor_id: 42,
+        last_seen: Instant::now(),
+        source_network: None,
+        source_address: None,
+    });
+
+    let responder = tokio::spawn(async move {
+        let received = timeout(Duration::from_secs(1), remote_rx.recv())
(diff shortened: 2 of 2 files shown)


jscott3201 (commented):
[Runtime] READY

No material runtime, concurrency, or regression-test defect found in `bb7230c3bc3db188dbe93b9ca0a448a2ffc5f84c...78fce06b81dd0292b3009224bd540b8d3f39b468`.

Evidence:
- `crates/bacnet-client/src/client/requests.rs:61-69` converts every discovered value through `u16::try_from`: values through `u16::MAX` remain exact, over-wide `u32` values saturate to `u16::MAX`, and the result is then bounded by the live transport APDU budget. The device-table guard is confined to this block and is dropped before either request path awaits network/segmentation work (`requests.rs:70-79`).
- Routed requests are unchanged and do not perform this conversion: `requests.rs:82-94` uses the configured maximum and routed transport budget, while `requests.rs:228-232` subtracts the routed NPDU overhead. Thus the modulo-wrap failure was local-only; the patch adds no routed lock or task behavior.
- The regression is discriminating: `client/tests.rs:318-328` installs exactly `65,536`, and `client/tests.rs:330-358` requires an unsegmented request and successful ACK. On the base implementation, `65,536 as u16` is zero; `requests.rs:70-79` then enters segmentation, where `client/segmentation.rs:180-181` computes a zero payload and `bacnet-encoding/src/segmentation.rs:40-43` rejects the non-empty request before anything is sent. The test therefore exercises the original failure rather than merely checking a conversion helper.
- The successful unsegmented path registers its TSM transaction only after sizing (`requests.rs:98-110`) and uses `TsmGuard`; normal response/send-error/timeout cleanup is at `requests.rs:157-185`. The test awaits the responder and explicitly stops both network/client sides (`client/tests.rs:330-359`).
- The head's GitHub checks report successful `Test (ubuntu-latest)`, Clippy, and Rustfmt runs.

Residual test gaps (non-blocking): no explicit boundary cases for `u16::MAX` or `u32::MAX`, and no routed analogue. The routed branch does not consume discovered max-APDU values, so a routed test would not exercise this conversion.

Commands run: none; review was read-only via exact-SHA GitHub retrieval and local file inspection.

jscott3201 (commented):
[Dataflow] READY

No material diff findings.

Verified data flow: `IAmRequest::decode` retains the wire unsigned value after checked conversion to `u32` (`crates/bacnet-services/src/who_is.rs:167-174`); dispatch stores that same value in `DiscoveredDevice` (`crates/bacnet-client/src/client/dispatch.rs:137-147`), whose field is `u32` (`crates/bacnet-client/src/discovery.rs:12-22`). The changed request path converts it with `u16::try_from(...).unwrap_or(u16::MAX)` and then applies the transport budget (`crates/bacnet-client/src/client/requests.rs:61-79`). Segmentation only consumes this already-narrowed, transport-capped `u16` (`crates/bacnet-client/src/client/segmentation.rs:169-182`).

The saturation order is correct. `TransportPort::max_apdu_length` is `u16` (`crates/bacnet-transport/src/port.rs:128-132`), so for every peer value `p: u32` and transport cap `t: u16`, `min(saturating_u16(p), t)` equals `min(p, t)`: values through 65,535 preserve their value; 65,536 through `u32::MAX` become 65,535 before the minimum and thus resolve to `t`. The cap is additionally reduced for routed NPDU header bytes with `saturating_sub` (`requests.rs:228-233`). Built-in MS/TP reports 480 (`crates/bacnet-transport/src/mstp/port.rs:428-430`), while BACnet/SC exposes its negotiated `u16` cap (`crates/bacnet-transport/src/sc/mod.rs:774-776`).

The regression test exercises the prior 65,536 -> 0 alias and observes a non-segmented confirmed request (`crates/bacnet-client/src/client/tests.rs:305-359`). The previous zero would reach `split_payload` via segmentation and reject a non-empty payload at a zero segment size (`crates/bacnet-encoding/src/segmentation.rs:36-55`).

Residual risk (non-blocking): the new test seeds the discovery table directly (`tests.rs:318-328`), rather than delivering an encoded I-Am through decode/dispatch. The production decode/store path is checked above; adding an end-to-end I-Am fixture would guard that integration boundary. No remaining peer-advertised-I-Am `u32` -> `u16` consumer was found in the reviewed client flow. The other repository casts are server-local-config paths, whose configuration is bounded to transport and checked before startup (`crates/bacnet-server/src/server/lifecycle.rs:28-36`), so they are not this peer-data path.

PR #320, opened 2026-08-14
fix(services): validate You-Are vendor ID

jscott3201:
## Summary
- reject You-Are vendor IDs whose decoded value exceeds `u16::MAX`
- require context tags 0, 1, and 2 for the mandatory fields
- cover exact maximum, wider leading-zero, overflow-alias, and wrong-tag encodings

## Why
`decode_unsigned` returns `u64`. The prior `as u16` conversion let peer values such as 65,536 and 65,537 alias to vendor IDs 0 and 1.

## Testing
- `cargo test -p bacnet-services --locked who_am_i`
- `git diff --check`

Related to #309.

### DIFF
--- a/crates/bacnet-services/src/who_am_i.rs
+++ b/crates/bacnet-services/src/who_am_i.rs
@@ -70,15 +70,22 @@ impl YouAreRequest {
 
         // [0] vendorID
         let (tag, pos) = tags::decode_tag(data, offset)?;
+        if !tag.is_context(0) {
+            return Err(Error::decoding(offset, "YouAre expected context tag 0"));
+        }
         let end = pos + tag.length as usize;
         if end > data.len() {
             return Err(Error::decoding(pos, "YouAre truncated at vendor-id"));
         }
-        let vendor_id = primitives::decode_unsigned(&data[pos..end])? as u16;
+        let vendor_id = u16::try_from(primitives::decode_unsigned(&data[pos..end])?)
+            .map_err(|_| Error::decoding(pos, "YouAre vendor-id exceeds u16"))?;
         offset = end;
 
         // [1] modelName
         let (tag, pos) = tags::decode_tag(data, offset)?;
+        if !tag.is_context(1) {
+            return Err(Error::decoding(offset, "YouAre expected context tag 1"));
+        }
         let end = pos + tag.length as usize;
         if end > data.len() {
             return Err(Error::decoding(pos, "YouAre truncated at model-name"));
@@ -88,6 +95,9 @@ impl YouAreRequest {
 
         // [2] serialNumber
         let (tag, pos) = tags::decode_tag(data, offset)?;
+        if !tag.is_context(2) {
+            return Err(Error::decoding(offset, "YouAre expected context tag 2"));
+        }
         let end = pos + tag.length as usize;
         if end > data.len() {
             return Err(Error::decoding(pos, "YouAre truncated at serial-number"));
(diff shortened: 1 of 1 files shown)


jscott3201 (commented):
[Dataflow] READY

No material findings.

Evidence reviewed:
- `crates/bacnet-services/src/who_am_i.rs:73-83` validates the vendor field as primitive context tag 0 before consuming its content and narrows the decoded `u64` with `u16::try_from`. This rejects values above 65,535 while accepting representable non-minimal encodings such as a leading-zero three-octet value.
- `crates/bacnet-services/src/who_am_i.rs:86-107` enforces primitive context tags 1 and 2 at each mandatory-field offset before passing content to the CharacterString decoder. `Tag::is_context` also excludes application and opening/closing tag forms.
- `crates/bacnet-services/src/who_am_i.rs:109-126` keeps the optional [3]/[4] state transitions unchanged: [4] remains accepted when [3] is absent, and offsets advance only for a matching optional field.
- `crates/bacnet-encoding/src/tags.rs:58-63, 423-448` confirms those checks validate class/number/primitive form and that optional decoding preserves the offset on a non-match.
- `crates/bacnet-encoding/src/primitives/mod.rs:67-79` confirms the raw unsigned decoder accepts 1–8 octets, with the public `u16` boundary now enforced by the service decoder rather than by a wire-width shortcut.

Residual risk: tests were inspected but not executed in this read-only review environment. Pre-existing behavior that ignores unrecognized trailing data after optional fields is outside this change.

jscott3201 (commented):
[Adversary] READY — Ready to merge — reviewed against `fffba233c3f41888223ada0a200f1e913f50e699..3f823d8ef05eae8e20daf1e532872f23d3f1f0dd`.

No material defects found. `crates/bacnet-services/src/who_am_i.rs:72-105` now requires primitive context tags 0/1/2 and range-checks the decoded vendor ID; `crates/bacnet-encoding/src/tags.rs:65-70` rejects application/opening/closing forms; `crates/bacnet-encoding/src/primitives/mod.rs:66-77` rejects empty or >8-byte unsigned payloads.

Coverage note (non-blocking): `crates/bacnet-services/src/who_am_i.rs:204-229` covers limits, aliases, and wrong context numbers, but does not directly exercise application-tag or opening/closing-tag mandatory fields, or empty/9-byte vendor payloads. Those paths are covered mechanically by the cited shared predicates/codecs.

jscott3201 (commented):
[Runtime] READY

No material PR-introduced runtime or test defect found for `fffba233...3f823d8`.

Verified:
- Mandatory fields now reject wrong classes, numbers, and opening/closing forms at `crates/bacnet-services/src/who_am_i.rs:72-75`, `:85-88`, and `:97-100`; `Tag::is_context` enforces all four conditions at `crates/bacnet-encoding/src/tags.rs:64-70`.
- Vendor values are decoded before a checked `u16::try_from` at `crates/bacnet-services/src/who_am_i.rs:76-81`. `decode_unsigned` handles 1–8 big-endian octets at `crates/bacnet-encoding/src/primitives/mod.rs:65-77`, so 65535 and wider leading-zero encodings remain accepted while 65536, 65537, and `u64::MAX` are rejected.
- The focused tests exercise the exact maximum, a three-octet leading-zero maximum, the former wraparound aliases, and each wrong mandatory context number at `crates/bacnet-services/src/who_am_i.rs:203-230`. Existing empty/truncated cases remain at `:232-249`.

Baseline, non-blocking for this patch: the decoder still tolerates well-formed trailing/unexpected elements after the mandatory fields because unmatched optional tags leave the offset unchanged and the final offset is discarded (`crates/bacnet-services/src/who_am_i.rs:108-127`). This behavior is present at the base SHA and was not introduced here.

Verification: GitHub head checks passed for Test (ubuntu-latest), Clippy, Rustfmt, file-size, and no-secret scan. No local command runner was available in this review session, so the PR-described cargo commands were not independently rerun.

PR #321, opened 2026-08-14
fix(services): validate Who-Has limit widths

jscott3201:
## Summary
- check Who-Has low and high limits against `u32` before conversion
- reject overflow aliases before server response filtering
- preserve exact `u32::MAX` values encoded with wider leading-zero content

## Why
`decode_unsigned` returns `u64`. The previous `as u32` conversions let peer values above `u32::MAX` wrap and change whether a server considered itself inside the requested device-instance range.

## Testing
- `cargo test -p bacnet-services --locked who_has`
- `git diff --check`

Related to #309.

### DIFF
--- a/crates/bacnet-services/src/who_has.rs
+++ b/crates/bacnet-services/src/who_has.rs
@@ -56,7 +56,13 @@ impl WhoHasRequest {
         let mut low_limit = None;
         let (opt, new_offset) = tags::decode_optional_context(data, offset, 0)?;
         if let Some(content) = opt {
-            low_limit = Some(primitives::decode_unsigned(content)? as u32);
+            let low_limit_raw = primitives::decode_unsigned(content)?;
+            low_limit = Some(u32::try_from(low_limit_raw).map_err(|_| {
+                Error::decoding(
+                    offset,
+                    format!("WhoHas low-limit {low_limit_raw} exceeds u32"),
+                )
+            })?);
             offset = new_offset;
         }
 
@@ -64,7 +70,13 @@ impl WhoHasRequest {
         let mut high_limit = None;
         let (opt, new_offset) = tags::decode_optional_context(data, offset, 1)?;
         if let Some(content) = opt {
-            high_limit = Some(primitives::decode_unsigned(content)? as u32);
+            let high_limit_raw = primitives::decode_unsigned(content)?;
+            high_limit = Some(u32::try_from(high_limit_raw).map_err(|_| {
+                Error::decoding(
+                    offset,
+                    format!("WhoHas high-limit {high_limit_raw} exceeds u32"),
+                )
+            })?);
             offset = new_offset;
         }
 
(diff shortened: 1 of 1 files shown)


jscott3201 (commented):
[Dataflow] READY

No material findings.

Evidence: `crates/bacnet-services/src/who_has.rs:56-81` decodes each peer limit to `u64` and rejects failed `u32::try_from` conversions before the decoded request is returned. `crates/bacnet-server/src/handlers/object_mgmt.rs:6-41` performs that decode before range filtering, object lookup, or I-Have construction; `crates/bacnet-server/src/server/requests/unconfirmed.rs:96-137` broadcasts only on `Ok(Some(i_have))` and logs the decode error otherwise. Thus an overflowing low or high value cannot reach response dispatch. Valid `u32` values—including the deliberately non-minimal leading-zero encoding of `u32::MAX`—retain their numeric value (`who_has.rs:194-231`).

Client compatibility is preserved: the Rust and Python client surfaces accept `Option<u32>` limits (`crates/bacnet-client/src/client/discovery.rs:77-95`; `crates/rusty-bacnet/src/client/client_methods/cov_discovery.rs:83-132`), and `WhoHasRequest::encode` widens only for the unsigned wire codec (`who_has.rs:32-40`). I-Have encoding is unchanged (`who_has.rs:115-120`).

Residual risk / pre-existing behavior: range filtering is applied only when both limits are present (`object_mgmt.rs:14-18`); this PR does not alter that behavior. Review was static; tests were not executed here.

jscott3201 (commented):
[Adversary] READY

Reviewed the authoritative range `188797c6febb85c122b3e94a6ace118c89dbe686...224d0bdc7882cb397cda49f7f971e52bf470eddc`. No material defect found in this change.

Verified behavior:
- `crates/bacnet-services/src/who_has.rs:59-64` and `:73-79` reject every decoded `Unsigned` value above `u32::MAX` before it reaches the device-instance range filter. This prevents the prior low-word aliasing caused by `as u32`.
- The same decoder still accepts representable wide, leading-zero encodings (`who_has.rs:221-230`), while `decode_unsigned` rejects empty or >8-octet payloads and tag decoding rejects truncated lengths (`crates/bacnet-encoding/src/primitives/mod.rs:51-60`, `crates/bacnet-encoding/src/tags.rs:402-424`).
- Both object CHOICE paths remain downstream of successful request decoding (`crates/bacnet-server/src/handlers/object_mgmt.rs:7-37`), so malformed overflow limits cannot silently become an I-Have response.

Coverage gap (non-blocking): `who_has_limits_must_fit_u32` tests the low limit at `u64::MAX` but not the high limit independently, and does not directly exercise malformed/zero-length limit payloads or both `WhoHasObject` alternatives with an overflowing limit (`crates/bacnet-services/src/who_has.rs:195-230`).

Pre-existing observation, not a diff finding: limits in `4_194_304..=u32::MAX` are still accepted although an object identifier's device instance is only 22 bits (`crates/bacnet-types/src/primitives.rs:42-44`). The base SHA had the same behavior; this PR neither introduces nor broadens it.

CI reports Rustfmt, Clippy, file-size, no-secret, and Ubuntu tests successful.

jscott3201 (commented):
[Runtime] READY

No material runtime or test defect survived review of `188797c6febb85c122b3e94a6ace118c89dbe686...224d0bdc7882cb397cda49f7f971e52bf470eddc`.

Verified evidence:
- `crates/bacnet-services/src/who_has.rs:57-66` and `:71-80` decode each optional Unsigned into `u64` and use `u32::try_from` before advancing to or parsing later fields. Values above `u32::MAX` therefore fail rather than aliasing into the server's range filter.
- `crates/bacnet-encoding/src/tags.rs:397-417` leaves the offset unchanged when an optional tag is absent, advances to the content end only on a matching tag, and rejects truncated declared content. Thus low-only, high-only, both-present, and both-absent layouts preserve the correct CHOICE offset.
- `crates/bacnet-encoding/src/primitives/mod.rs:65-77` numerically decodes 1-8 octets into `u64`; a 5-byte `00 ff ff ff ff` payload remains exactly `u32::MAX`, while a numerically larger payload is rejected by the new conversion.
- `crates/bacnet-services/src/who_has.rs:208-231` exercises low overflow at `u32::MAX + 1`, independent high overflow, `u64::MAX`, and exact `u32::MAX` encoded with a leading zero for both fields. Existing tests cover the `[2]` identifier path at `:169-182`, the `[3]` name path at `:184-195`, and representative malformed/truncated CHOICE inputs at `:251-293`.
- `crates/bacnet-server/src/handlers/object_mgmt.rs:11-18` propagates decode failure before consulting either limit, so hostile over-wide values cannot reach response filtering.

Baseline behavior, not introduced by this diff: `WhoHasRequest::decode` returns after the first CHOICE value without validating trailing bytes (`crates/bacnet-services/src/who_has.rs:84-106`), and the server applies range filtering only when both optional limits are present (`crates/bacnet-server/src/handlers/object_mgmt.rs:13-18`).

Verification: GitHub reports `Test (ubuntu-latest)`, Clippy, Rustfmt, file-size, and no-secret checks successful at the reviewed head. I could not run local shell commands in this review runtime; residual coverage gaps are a direct high-only-at-offset-zero assertion and explicit malformed 0-/9-byte limit tests, though the inspected helper paths reject those encodings.

PR #324, opened 2026-08-14
fix(services): validate private transfer widths

jscott3201:
## Summary
- check request and acknowledgment vendor IDs against their existing `u32` API fields
- check request and acknowledgment service numbers against `u32`
- require mandatory context tags 0 and 1 in both decoders
- cover overflow aliases, exact maxima, wider leading-zero encodings, and wrong tags

## Why
Both Private Transfer decoders converted peer `u64` values with `as`, allowing over-wide vendor IDs and service numbers to alias representable values before confirmed responses reached Python callers.

This preserves the existing public `u32` vendor-ID contract; changing that API to a narrower BACnet-specific type is outside this decoder-width slice.

## Testing
- `cargo test -p bacnet-services --locked private_transfer`
- `git diff --check`

Related to #309.

### DIFF
--- a/crates/bacnet-services/src/private_transfer.rs
+++ b/crates/bacnet-services/src/private_transfer.rs
@@ -39,6 +39,12 @@ impl PrivateTransferRequest {
 
         // [0] vendorID
         let (tag, pos) = tags::decode_tag(data, offset)?;
+        if !tag.is_context(0) {
+            return Err(Error::decoding(
+                offset,
+                "PrivateTransfer expected context tag 0 for vendorID",
+            ));
+        }
         let end = pos + tag.length as usize;
         if end > data.len() {
             return Err(Error::decoding(
@@ -46,11 +52,23 @@ impl PrivateTransferRequest {
                 "PrivateTransfer truncated at vendorID",
             ));
         }
-        let vendor_id = primitives::decode_unsigned(&data[pos..end])? as u32;
+        let vendor_id_raw = primitives::decode_unsigned(&data[pos..end])?;
+        let vendor_id = u32::try_from(vendor_id_raw).map_err(|_| {
+            Error::decoding(
+                pos,
+                format!("PrivateTransfer vendorID {vendor_id_raw} exceeds u32"),
+            )
+        })?;
         offset = end;
 
         // [1] serviceNumber
         let (tag, pos) = tags::decode_tag(data, offset)?;
+        if !tag.is_context(1) {
+            return Err(Error::decoding(
+                offset,
+                "PrivateTransfer expected context tag 1 for serviceNumber",
+            ));
+        }
         let end = pos + tag.length as usize;
         if end > data.len() {
             return Err(Error::decoding(
@@ -58,7 +76,13 @@ impl PrivateTransferRequest {
                 "PrivateTransfer truncated at serviceNumber",
             ));
         }
-        let service_number = primitives::decode_unsigned(&data[pos..end])? as u32;
+        let service_number_raw = primitives::decode_unsigned(&data[pos..end])?;
+        let service_number = u32::try_from(service_number_raw).map_err(|_| {
+            Error::decoding(
+                pos,
+                format!("PrivateTransfer serviceNumber {service_number_raw} exceeds u32"),
+            )
+        })?;
         offset = end;
 
         // [2] serviceParameters (optional, opening/closing)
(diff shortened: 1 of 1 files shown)


jscott3201 (commented):
[Dataflow] READY

Findings: no material defect in `f2f9520ae21074e1336359134a17da5927a74a49...b3623b14d4396b0fd5cb39b0e01e6bf3e22f5083`.

Evidence:
- Both decoders validate mandatory context tags before decoding and perform checked `u64`→`u32` conversion, so a peer value above `u32::MAX` or a wrong mandatory tag fails before a request/ack struct is returned: `crates/bacnet-services/src/private_transfer.rs:41-86,139-184`.
- The confirmed Python path decodes the peer ACK before constructing the result dict; rejected fields therefore cannot reach Python callers: `crates/rusty-bacnet/src/client/client_methods/file_list_private_text_life.rs:211-232`.
- Public Python inputs are `u32` and both confirmed/unconfirmed paths build the request then use its shared encoder: `crates/rusty-bacnet/src/client/client_methods/file_list_private_text_life.rs:187-210,239-269`. The encoder widens each `u32` to `u64`, and the new tests cover `u32::MAX`, leading-zero five-octet encodings, and over-wide aliases for each field and decoder: `crates/bacnet-services/src/private_transfer.rs:24-34,275-318`.

Residual risks (pre-existing, outside this diff): optional/trailing data is still only consumed when it begins with opening context tag 2; other trailing tags are ignored in both decoders (`crates/bacnet-services/src/private_transfer.rs:88-104,186-202`). Server dispatch for either private-transfer service is absent: confirmed unsupported services reject via the fallback (`crates/bacnet-server/src/server/requests/mod.rs:14-37,368-377`), while the unconfirmed dispatch list/chain excludes it (`crates/bacnet-server/src/server/requests/unconfirmed.rs:12-19,37-182`).

jscott3201 (commented):
[Adversary] READY

Reviewed exactly `f2f9520ae21074e1336359134a17da5927a74a49...b3623b14d4396b0fd5cb39b0e01e6bf3e22f5083`.

No material defects found. The decoders now reject non-primitive/non-context mandatory tags at `crates/bacnet-services/src/private_transfer.rs:43-48, 67-72, 141-146, 165-170`, and reject decoded unsigned values above the existing `u32` API boundary at `:55-62, 79-86, 153-160, 177-184`. The added tests cover both fields at `u32::MAX`, leading-zero five-octet encodings, `u32::MAX + 1`, `u64::MAX`, and representative wrong mandatory tags (`:274-324`).

I could not independently execute the test suite in this environment.

jscott3201 (commented):
[Runtime] READY

No material defect survived review of `f2f9520ae21074e1336359134a17da5927a74a49...b3623b14d4396b0fd5cb39b0e01e6bf3e22f5083`.

Verified:
- Both request fields validate primitive context tags before reading content (`crates/bacnet-services/src/private_transfer.rs:41-47`, `:65-71`), and ACK mirrors those checks (`:139-145`, `:163-169`). `Tag::is_context` also excludes opening/closing forms (`crates/bacnet-encoding/src/tags.rs:64-70`).
- Request and ACK slice only the declared, in-buffer content and use checked `u64 -> u32` conversions (`crates/bacnet-services/src/private_transfer.rs:48-85`, `:146-183`). `decode_unsigned` accepts 1-8 octets and computes the numeric value, so wider leading-zero encodings remain accepted when the value fits (`crates/bacnet-encoding/src/primitives/mod.rs:65-77`).
- Tests exercise `u32::MAX`, five-octet leading-zero maxima, `u32::MAX + 1`, `u64::MAX`, both field positions, both decoders, wrong context numbers, and an application-tagged vendor field (`crates/bacnet-services/src/private_transfer.rs:275-318`). The successful Linux CI test runs `cargo test --workspace --exclude rusty-bacnet --locked` with the configured feature set (`.github/workflows/ci.yml:126-153`).

Baseline issues, not introduced by this diff:
- Optional request/ACK bodies still use the TLV-walking `extract_context_value` (`crates/bacnet-services/src/private_transfer.rs:90-98`, `:188-196`). That helper parses enclosed bytes as tags (`crates/bacnet-encoding/src/tags.rs:282-328`), despite the raw helper documenting vendor-defined parameter bodies as potentially non-TLV (`:337-354`). For example, an encoded opaque body `[0xDE, 0xAD]` can fail to decode. The same calls exist in base file SHA `9e0a707217069a27fb0f8b2c70fa1a8dcb74d316`.
- Both decoders also continue to ignore a trailing element that is not opening tag 2 (`crates/bacnet-services/src/private_transfer.rs:90-98`, `:188-196`), and primitive unsigned length errors retain `decode_unsigned`'s relative offset 0 (`crates/bacnet-encoding/src/primitives/mod.rs:65-71`). Both behaviors predate this PR.

Residual gap: I did not independently run local commands because this review environment exposed no shell executor; I verified the successful GitHub test check and inspected the exact head sources and patch.

PR #325, opened 2026-08-14
fix(services): validate atomic file counts

jscott3201:
## Summary
- check AtomicReadFile requested octet and record counts against `u32`
- check AtomicWriteFile record counts and AtomicReadFileAck returned counts against `u32`
- require application Unsigned tags and reject reserved application L/V/T forms for these fields
- preserve the existing `MAX_DECODED_ITEMS` cap for record payload decoders
- cover overflow aliases, exact accepted limits, wider leading-zero encodings, wrong tags, and reserved forms

## Why
The file-service decoders converted peer `u64` counts with `as`. Over-wide record counts could alias below `MAX_DECODED_ITEMS` before record iteration or server dispatch, while over-wide read counts could alias before read sizing.

The boundary tests live in a separate module so the existing `file.rs` remains within the repository source-size cap.

## Testing
- `cargo test -p bacnet-services --locked file::`
- `git diff --check`

Related to #309.

### DIFF
--- a/crates/bacnet-services/src/file.rs
+++ b/crates/bacnet-services/src/file.rs
@@ -21,6 +21,31 @@ fn checked_slice<'a>(
     Ok((&content[p..end], end))
 }
 
+fn checked_unsigned_u32(
+    content: &[u8],
+    offset: usize,
+    context: &str,
+) -> Result<(u32, usize), Error> {
+    let (tag, pos) = tags::decode_tag(content, offset)?;
+    if tag.class != tags::TagClass::Application
+        || tag.number != tags::app_tag::UNSIGNED
+        || content[offset] & 0x07 > 5
+    {
+        return Err(Error::decoding(
+            offset,
+            format!("{context} expected application Unsigned"),
+        ));
+    }
+    let end = pos + tag.length as usize;
+    if end > content.len() {
+        return Err(Error::decoding(pos, format!("{context} truncated")));
+    }
+    let raw = primitives::decode_unsigned(&content[pos..end])?;
+    let value = u32::try_from(raw)
+        .map_err(|_| Error::decoding(pos, format!("{context} {raw} exceeds u32")))?;
+    Ok((value, end))
+}
+
 // ---------------------------------------------------------------------------
 // AtomicReadFile-Request
 // ---------------------------------------------------------------------------
@@ -112,12 +137,11 @@ impl AtomicReadFileRequest {
             let (slice, inner) =
                 checked_slice(content, 0, "AtomicReadFile stream file-start-position")?;
             let file_start_position = primitives::decode_signed(slice)?;
-            let (slice, _) = checked_slice(
+            let (requested_octet_count, _) = checked_unsigned_u32(
                 content,
                 inner,
                 "AtomicReadFile stream requested-octet-count",
             )?;
-            let requested_octet_count = primitives::decode_unsigned(slice)? as u32;
             FileAccessMethod::Stream {
                 file_start_position,
                 requested_octet_count,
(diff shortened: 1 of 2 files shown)


jscott3201 (commented):
[Dataflow] READY

Reviewed exactly `d1cc8b4fb67fc9f6c23f23b79a94483738babc7e...6ec77c8cc66d68459efc5204832cfe9bbeb23d06`. No change-introduced material defect found.

Evidence:
- `crates/bacnet-services/src/file.rs:24-47` validates application Unsigned form (including rejecting reserved L/V/T 6/7), bounds the tag payload before decoding it, and performs `u64 -> u32` fallible conversion before yielding a count.
- The AtomicReadFile request paths use that helper before constructing either access variant (`file.rs:137-161`); server object lookup follows request decode (`crates/bacnet-server/src/handlers/file.rs:119-124`). Thus over-wide/wrong-tag peer counts cannot reach object access or response sizing.
- AtomicWriteFile and AtomicReadFileAck validate their record counts before their respective `MAX_DECODED_ITEMS` tests and record-data loops (`file.rs:228-253`, `file.rs:340-369`), so rejected values do not enter record allocation/iteration or caller-visible ACK values.
- Rust and Python client entry points are already bounded to `u32` (`crates/bacnet-client/src/client/file_list.rs:82-123`; `crates/rusty-bacnet/src/client/client_methods/file_list_private_text_life.rs:17-104`) and return raw response bytes; they do not bypass the service decoder for server-originated requests.

Residual risks (pre-existing; not introduced by this change):
- Record count/data cardinality is still not enforced: both record decoders stop when their content ends and can return a count larger than `file_record_data.len()` (`crates/bacnet-services/src/file.rs:238-253`, `file.rs:350-369`).
- Server file handlers still use unchecked `usize` arithmetic/casts while sizing stream and record operations (`crates/bacnet-server/src/handlers/file.rs:166-167`, `:191-206`, `:279-282`). On 32-bit targets, valid maximum `u32` counts combined with large starts/data can wrap; these paths predate this PR.

jscott3201 (commented):
[Runtime] READY

Reviewed the exact range `d1cc8b4fb67fc9f6c23f23b79a94483738babc7e...40be0b3df96f4d727b78649f351be96508149bbe`. I found no material runtime or test defect in the changed count-decoding behavior.

Evidence:
- `crates/bacnet-services/src/file.rs:24-46` decodes from the supplied inner offset, requires application Unsigned, rejects raw application L/V/T 6/7, bounds the declared contents, decodes all 1-8 value octets, and performs a checked `u64`→`u32` conversion.
- All four affected fields use that helper: read stream/request record at `file.rs:140-158`, write record at `file.rs:231-245`, and read-ACK record at `file.rs:343-363`. The two payload-bearing paths perform checked narrowing before the existing `MAX_DECODED_ITEMS` guard.
- `crates/bacnet-services/src/file_width_tests.rs:7-15,47-78` constructs and accepts a genuine five-octet leading-zero Unsigned representation of `u32::MAX`; `file_width_tests.rs:38-44,80-111` appends and decodes exactly 10,000 actual record elements; `file_width_tests.rs:113-165` exercises the overflow aliases, wrong application/context tags, and reserved L/V/T forms.
- Exact-head GitHub checks passed: full-feature Ubuntu workspace tests, Clippy, Rustfmt, and the 700-LOC file-size check.

Baseline behavior, not attributed to this PR: AtomicReadFile ignores content after the decoded count (`file.rs:140-165`), while the write and read-ACK record loops accept fewer or extra record elements than the declared count (`file.rs:238-251`, `file.rs:353-369`). Those framing/count-consistency behaviors are unchanged by this range.

Residual verification gap: I could not execute local shell commands in this review environment, so command evidence is from the successful exact-head CI run. There is no dedicated `MAX_DECODED_ITEMS + 1` or six-to-eight-octet leading-zero fixture, though the direct cap and shared 1-8-octet decoder paths are visible in the cited code.

PR #326, opened 2026-08-14
fix(services): validate read property widths

jscott3201:
## Summary
- check ReadProperty request and ACK property identifiers against their existing `u32` wrapper
- check optional request and ACK property array indexes against `u32`
- require mandatory context tags 0 and 1 in both decoders
- cover overflow aliases, exact maxima, wider leading-zero encodings, and wrong tags

## Why
Both ReadProperty decoders converted peer `u64` identifiers and array indexes with `as`. Over-wide request values could alias a different property or array element before database lookup; over-wide ACK values could similarly reach Rust and Python callers as valid-looking metadata.

This preserves the shipped `u32` `PropertyIdentifier` contract. Enforcing its documented 22-bit BACnet domain would be a separate compatibility change.

## Testing
- `cargo test -p bacnet-services --locked read_property`
- `git diff --check`

Related to #309.

### DIFF
--- a/crates/bacnet-services/src/read_property.rs
+++ b/crates/bacnet-services/src/read_property.rs
@@ -7,6 +7,29 @@ use bacnet_types::error::Error;
 use bacnet_types::primitives::ObjectIdentifier;
 use bytes::BytesMut;
 
+fn decode_context_u32(
+    data: &[u8],
+    offset: usize,
+    tag_number: u8,
+    context: &str,
+) -> Result<(u32, usize), Error> {
+    let (tag, pos) = tags::decode_tag(data, offset)?;
+    if !tag.is_context(tag_number) {
+        return Err(Error::decoding(
+            offset,
+            format!("{context} expected context tag {tag_number}"),
+        ));
+    }
+    let end = pos + tag.length as usize;
+    if end > data.len() {
+        return Err(Error::decoding(pos, format!("{context} truncated")));
+    }
+    let raw = primitives::decode_unsigned(&data[pos..end])?;
+    let value = u32::try_from(raw)
+        .map_err(|_| Error::decoding(pos, format!("{context} {raw} exceeds u32")))?;
+    Ok((value, end))
+}
+
 // ---------------------------------------------------------------------------
 // ReadPropertyRequest
 // ---------------------------------------------------------------------------
@@ -41,6 +64,12 @@ impl ReadPropertyRequest {
 
         // [0] object-identifier
         let (tag, pos) = tags::decode_tag(data, offset)?;
+        if !tag.is_context(0) {
+            return Err(Error::decoding(
+                offset,
+                "ReadProperty request expected context tag 0 for object-id",
+            ));
+        }
         let end = pos + tag.length as usize;
         if end > data.len() {
             return Err(Error::decoding(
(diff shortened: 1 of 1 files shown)


jscott3201 (commented):
[Adversary] READY

No material defect was introduced in `043b511e8430dd0c2f99c1b9206aaa844299526b...e75da277d8f710126955341abce39b3cccfb11ed`. The new helper rejects empty/over-8-byte values through `decode_unsigned`, rejects >`u32::MAX` before conversion, and `is_context` excludes application, opening, and closing tags (`crates/bacnet-services/src/read_property.rs:10-30`; `crates/bacnet-encoding/src/tags.rs:64-80`).

Baseline issue (pre-existing; not counted against this diff's readiness):
- **P2 — service decoders silently accept bytes outside their grammar.** A valid request followed by an unexpected context tag, e.g. `0C 00 00 00 01 19 55 49 01`, is returned as a valid request because the optional-index peek only consumes context tag 2 and the decoder returns without requiring EOF (`crates/bacnet-services/src/read_property.rs:91-104`). A valid ACK followed by arbitrary data, e.g. `0C 00 00 00 01 19 55 3E 00 3F 21 01`, is also returned as valid: both paths discard the offset returned by `extract_context_value` (`crates/bacnet-services/src/read_property.rs:184-190,199-206`). These are full peer service bodies at the server/client boundaries (`crates/bacnet-server/src/handlers/read_property.rs:7-13`; `crates/bacnet-client/src/client/property.rs:22-27`), so malformed wire data is silently ignored. The same behavior is present in the specified base; it should be remediated separately.

Coverage gaps: tests cover one five-byte leading-zero representation only, not 6–8-byte leading-zero representations; they do not explicitly cover opening/closing mandatory tags, application tags for ACK fields, malformed property/index lengths, low-word overflow aliases, or ACK/request trailing data (`crates/bacnet-services/src/read_property.rs:305-345`).

Evidence: compared the PR's verified base/head via `github_pull_request_read(get/get_diff)` and statically traced `decode_tag`, `decode_unsigned`, `extract_context_value`, and the client/server call sites. No test command was available in this read-only review environment.

jscott3201 (commented):
[Runtime] NOT READY

- **[P2] A malformed present `[2]` field is silently treated as an absent array index** — `crates/bacnet-services/src/read_property.rs:91-98`. After the two mandatory fields, any remaining bytes can only be the optional primitive context tag 2, but this branch returns `Ok` whenever the next decoded tag is not that exact primitive form. For example, `[0x0c, 0, 0, 0, 1, 0x19, 0x55, 0x2e]` (valid object-id and property-id followed by an unterminated opening tag 2) decodes successfully with `property_array_index: None`: `decode_tag` accepts `0x2e`, while `Tag::is_context` deliberately excludes opening/closing tags (`crates/bacnet-encoding/src/tags.rs:64-70`). The server immediately dispatches this successful parse (`crates/bacnet-server/src/handlers/read_property.rs:12,35`), so malformed peer input becomes a valid no-index read instead of a decoding error. A valid primitive `[2]` followed by trailing garbage is likewise accepted because the returned end offset is discarded. The added tests at `crates/bacnet-services/src/read_property.rs:330-346` cover wrong mandatory tags but not malformed optional tag 2 or full request consumption.

The checked `u32` conversion, leading-zero handling, mandatory tags 0/1, and ACK opening-tag-3 control flow otherwise look sound. The 22-bit domain is intentionally not treated as part of this finding.

jscott3201 (commented):
[Runtime] READY

No material runtime or test defect found in `043b511e8430dd0c2f99c1b9206aaa844299526b...39a97ff5288eb7e5c48e21d1267a0621924650e6`.

The prior P2 is resolved. `ReadPropertyRequest::decode` now rejects every non-primitive-context-2 remainder and requires a decoded optional index to end exactly at the payload boundary (`crates/bacnet-services/src/read_property.rs:91-107`). Both ACK paths require extraction of opening tag 3 to consume the entire payload (`crates/bacnet-services/src/read_property.rs:187-197` and `206-215`). The regression test exercises the malformed opening-tag-2 input, indexed-request trailing data, and ACK trailing data with and without an index (`crates/bacnet-services/src/read_property.rs:364-387`).

The checked `u32` conversion and mandatory tag checks remain intact, and the updated head's test, Clippy, Rustfmt, file-size, and secret-scan checks are successful.

jscott3201 (commented):
[Adversary] READY

No material defect found in `043b511e8430dd0c2f99c1b9206aaa844299526b...39a97ff5288eb7e5c48e21d1267a0621924650e6`.

Verified behavior:
- The request decoder now treats any bytes after property-id as the optional field and rejects anything other than a primitive context tag 2; a present index must consume the rest of the service body (`crates/bacnet-services/src/read_property.rs:89-108`). Thus a closing/opening/application/wrong context tag cannot be mistaken for an absent index, and index-plus-tail is rejected.
- Both ACK paths—without and with optional index—require `extract_context_value` to consume exactly the body after the required opening tag 3 (`crates/bacnet-services/src/read_property.rs:180-223`).
- Valid constructed property values remain accepted: `extract_context_value` increments depth for an inner opening tag, decrements it for its close, and returns after the outer closing tag (`crates/bacnet-encoding/src/tags.rs:273-334`); the returned raw inner bytes are preserved by the ACK paths (`crates/bacnet-services/src/read_property.rs:194-203,212-222`).

Remaining baseline gap (pre-existing; outside this PR): `extract_context_value` tracks only nesting depth, not a tag-number stack. For example, an ACK whose property value is `3E 2E 21 01 4F 3F` (outer [3], opened [2], application unsigned 1, mismatched closing [4], outer close [3]) is accepted as raw property bytes: the inner close merely decrements depth and only the final close is checked (`crates/bacnet-encoding/src/tags.rs:285-311`).

Coverage gap: add a positive ReadPropertyACK test with a nested constructed property-value and negative mismatched-inner-closing test; existing follow-up tests cover request opening-[2], index-plus-tail, and ACK tails but not those cases (`crates/bacnet-services/src/read_property.rs:364-387`).

Evidence: verified the PR base/head with `github_pull_request_read(get)` and inspected the exact PR diff and referenced codec sources. No test command was available in this read-only environment.

Issue #327, opened 2026-08-14
Validate nested context closing tags during extraction


jscott3201:
## Problem

`bacnet_encoding::tags::extract_context_value` tracks nesting depth but does not retain the opening tag numbers for nested constructed values. A nested opening tag can therefore be closed by a different tag number without a decode error, provided the final closing tag matches the outer tag.

For example, inside an outer `[3]` value, nested opening `[2]` followed by closing `[4]` is accepted when a final closing `[3]` exists.

## Evidence

- `crates/bacnet-encoding/src/tags.rs:282-311` increments/decrements a scalar `depth`.
- The closing tag number is checked only when `depth` reaches zero, against the outer `tag_number`.
- This was observed while reviewing ReadProperty ACK property-value extraction in #326.

## Expected behavior

Track nested opening tag numbers and reject any mismatched closing tag at its nesting level. Preserve current nesting-depth and bounds limits.

## Suggested coverage

- matching nested opening/closing tags succeed
- mismatched nested closing tag fails
- mismatched outer closing tag still fails
- multiple and deeply nested constructed values retain current depth-limit behavior

jscott3201:
Resolved by #353, merged into `dev` as `c924898683212ea88f0256dd87e28bec37190af1`. Nested context extraction now validates matching closing tags, bounds, and depth, with hostile codec and service regressions.

PR #328, opened 2026-08-14
fix(services): validate write property widths

jscott3201:
## Summary
- check WriteProperty property identifiers and optional array indexes against their existing `u32` fields
- require mandatory context tags 0 and 1
- require any post-value field to be priority tag 4 and reject trailing fields
- cover overflow aliases, exact maxima, wider leading-zero encodings, wrong tags, malformed priorities, and trailing data

## Why
The decoder converted peer `u64` identifiers and indexes with `as`, allowing over-wide values to alias a different property or array element before the server performed object lookup, value decoding, name-index work, or database mutation. It also ignored unknown or trailing fields after the property value.

## Testing
- `cargo test -p bacnet-services --locked write_property`
- `git diff --check`

Related to #309.

### DIFF
--- a/crates/bacnet-services/src/write_property.rs
+++ b/crates/bacnet-services/src/write_property.rs
@@ -7,6 +7,29 @@ use bacnet_types::error::Error;
 use bacnet_types::primitives::ObjectIdentifier;
 use bytes::BytesMut;
 
+fn decode_context_u32(
+    data: &[u8],
+    offset: usize,
+    tag_number: u8,
+    context: &str,
+) -> Result<(u32, usize), Error> {
+    let (tag, pos) = tags::decode_tag(data, offset)?;
+    if !tag.is_context(tag_number) {
+        return Err(Error::decoding(
+            offset,
+            format!("{context} expected context tag {tag_number}"),
+        ));
+    }
+    let end = pos + tag.length as usize;
+    if end > data.len() {
+        return Err(Error::decoding(pos, format!("{context} truncated")));
+    }
+    let raw = primitives::decode_unsigned(&data[pos..end])?;
+    let value = u32::try_from(raw)
+        .map_err(|_| Error::decoding(pos, format!("{context} {raw} exceeds u32")))?;
+    Ok((value, end))
+}
+
 // ---------------------------------------------------------------------------
 // WritePropertyRequest
 // ---------------------------------------------------------------------------
@@ -53,6 +76,12 @@ impl WritePropertyRequest {
 
         // [0] object-identifier
         let (tag, pos) = tags::decode_tag(data, offset)?;
+        if !tag.is_context(0) {
+            return Err(Error::decoding(
+                offset,
+                "WriteProperty expected context tag 0 for object-id",
+            ));
+        }
         let end = pos + tag.length as usize;
         if end > data.len() {
             return Err(Error::decoding(pos, "WriteProperty truncated at object-id"));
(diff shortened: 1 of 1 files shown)


jscott3201 (commented):
[Dataflow] READY

No material defects found in `e5ac704ef159a539a89c2c3f3293a951645566ea...fa42a028be2c450ac2b5329b869495aac7d1e7ba`.

Evidence:
- The request decoder validates context tags and losslessly bounds peer property IDs and array indexes to `u32` before it returns a request (`crates/bacnet-services/src/write_property.rs:10-30, 77-116`), so aliases cannot reach object lookup, array gating, value decode, name-index handling, or mutation.
- The handler orders object existence, array gating, property-value decoding, Object_Name uniqueness, mutation, and name-index refresh without a mutation before a fallible validation stage (`crates/bacnet-server/src/handlers/write_property.rs:489-538`).
- Only a successful handler result records the OID; SimpleACK is sent before event/COV notification dispatch (`crates/bacnet-server/src/server/requests/mod.rs:110-121, 484-548`). A malformed request therefore produces an Error APDU and no notification side effects.
- Rust client routes construct the same request type and use its encoder, whose public ID/index fields are `u32` and are emitted as unsigned context values (`crates/bacnet-client/src/client/property.rs:223-253, 285-320`; `crates/bacnet-services/src/write_property.rs:60-71`).

Residual risks:
- The documented 22-bit BACnet PropertyIdentifier-domain question remains outside this change (#327) and is not a finding here.
- I did not execute the test suite in this read-only review.

jscott3201 (commented):
[Adversary] READY

No material defect found in `e5ac704ef159a539a89c2c3f3293a951645566ea...fa42a028be2c450ac2b5329b869495aac7d1e7ba`.

Verified statically: `write_property.rs:10-30` rejects empty/over-8-byte unsigned payloads through `decode_unsigned` and rejects values above `u32::MAX`; the mandatory field and optional-index paths enforce the expected context forms at `:79-103`; property-value framing remains balanced through `:107-116`; and the post-value path accepts only context tag 4, validates 1..=16, and rejects all trailing bytes at `:120-145`.

Coverage gaps (non-blocking): `write_property.rs:229-324` does not directly exercise zero-length property/index payloads, every wire width 1..=8 for both fields, application/opening/closing forms for the optional index, balanced nested value tags or mismatched nested close tags, 5..=8-byte zero-prefixed priorities, or `u64::MAX` priority. The current logic covers these cases by the same tag and unsigned/range checks, but explicit vectors would lock that behavior down.

Scope separation / shared #327 (not charged to this one-file diff): WPM still narrows property IDs and indexes with `as u32` in `crates/bacnet-services/src/common.rs:48,63,131,146`; `wpm.rs:86` consumes that decoder. A WPM `[0]` property ID payload `01 00 00 00 55` aliases low word `0x55`. This pre-existing sibling path needs its own fix.

Evidence: PR target SHAs verified via GitHub metadata; diff and the cited source were inspected. No shell/test command was available or run.

jscott3201 (commented):
[Runtime] READY

No material runtime or test defect survives review of `e5ac704ef159a539a89c2c3f3293a951645566ea...fa42a028be2c450ac2b5329b869495aac7d1e7ba`.

Evidence:
- `crates/bacnet-services/src/write_property.rs:16-30` verifies the expected primitive context tag, bounds the content, decodes without narrowing, and uses `u32::try_from` before constructing either public `u32` field. The overflow error is anchored at the field content offset.
- `crates/bacnet-services/src/write_property.rs:78-114` requires context tags 0 and 1, conditionally consumes only primitive tag 2, and then requires the opening tag 3 framing.
- `crates/bacnet-services/src/write_property.rs:120-145` accepts only primitive context tag 4 after the value, validates the decoded `u64` before its `u8` conversion, preserves valid 1..=16 encodings (including a leading-zero representation), and rejects bytes after priority at the first trailing offset.
- `crates/bacnet-services/src/write_property.rs:230-324` exercises exact `u32::MAX` values with five-octet leading-zero encodings, the `u32::MAX + 1` and `u64::MAX` aliases for both property and index, wrong mandatory tags/classes, malformed/wrong priority, trailing fields, priority boundaries, and two-octet leading-zero priority compatibility.
- The real server path consumes the full extracted property payload before mutation at `crates/bacnet-server/src/handlers/write_property.rs:455-475`, so these decoder checks guard the lookup/write boundary rather than only a test helper.
- Head-SHA CI reports successful Ubuntu tests, Clippy, and Rustfmt.

The shared nested context-tag matching behavior is intentionally excluded here and remains tracked by #327.

PR #329, opened 2026-08-14
fix(services): validate shared property widths

jscott3201:
## Summary
- check shared BACnetPropertyReference and BACnetPropertyValue property identifiers and optional array indexes against their existing `u32` fields
- require mandatory property identifier context tag 0
- preserve parent-owned sequence framing and optional priority behavior across RPM, WPM, COV, audit, and object-management consumers
- cover overflow aliases, `u64::MAX`, exact maxima in wider leading-zero forms, and wrong mandatory tags

## Why
These shared decoders converted peer `u64` values with `as`, allowing over-wide values to alias different properties or array elements. BACnetPropertyValue reaches WritePropertyMultiple validation and mutation, while PropertyReference drives RPM reads and subscription/audit references.

## Testing
- `cargo test -p bacnet-services --locked common::tests`
- focused RPM, WPM, COV, object-management, and audit tests
- `git diff --check`

Related to #309.

### DIFF
--- a/crates/bacnet-services/src/common.rs
+++ b/crates/bacnet-services/src/common.rs
@@ -9,6 +9,29 @@ use bytes::{BufMut, BytesMut};
 /// Safety limit for decoded sequences to prevent unbounded allocations.
 pub const MAX_DECODED_ITEMS: usize = 10_000;
 
+fn decode_context_u32(
+    data: &[u8],
+    offset: usize,
+    expected_tag: u8,
+    field: &str,
+) -> Result<(u32, usize), Error> {
+    let (tag, pos) = tags::decode_tag(data, offset)?;
+    if !tag.is_context(expected_tag) {
+        return Err(Error::decoding(
+            offset,
+            format!("{field} expected context tag {expected_tag}"),
+        ));
+    }
+    let end = pos + tag.length as usize;
+    if end > data.len() {
+        return Err(Error::decoding(pos, format!("{field} truncated")));
+    }
+    let value = primitives::decode_unsigned(&data[pos..end])?;
+    let value =
+        u32::try_from(value).map_err(|_| Error::decoding(pos, format!("{field} exceeds u32")))?;
+    Ok((value, end))
+}
+
 // ---------------------------------------------------------------------------
 // PropertyReference
 // ---------------------------------------------------------------------------
@@ -37,30 +60,17 @@ impl PropertyReference {
 
     pub fn decode(data: &[u8], offset: usize) -> Result<(Self, usize), Error> {
         // [0] propertyIdentifier
-        let (tag, pos) = tags::decode_tag(data, offset)?;
-        let end = pos + tag.length as usize;
-        if end > data.len() {
-            return Err(Error::decoding(
-                pos,
-                "PropertyReference truncated at property-id",
-            ));
-        }
-        let prop_id = primitives::decode_unsigned(&data[pos..end])? as u32;
-        let mut offset = end;
+        let (prop_id, mut offset) =
+            decode_context_u32(data, offset, 0, "PropertyReference property-id")?;
 
         // [1] propertyArrayIndex (optional)
         let mut array_index = None;
         if offset < data.len() {
-            let (tag, new_pos) = tags::decode_tag(data, offset)?;
+            let (tag, _) = tags::decode_tag(data, offset)?;
             if tag.is_context(1) {
-                let end = new_pos + tag.length as usize;
-                if end > data.len() {
-                    return Err(Error::decoding(
-                        new_pos,
-                        "PropertyReference truncated at array-index",
-                    ));
-                }
-                array_index = Some(primitives::decode_unsigned(&data[new_pos..end])? as u32);
+                let (value, end) =
+                    decode_context_u32(data, offset, 1, "PropertyReference array-index")?;
+                array_index = Some(value);
                 offset = end;
             }
         }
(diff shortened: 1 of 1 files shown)


jscott3201 (commented):
[Dataflow] READY

No PR-blocking findings in `6079bb71a9a7736b0e737f1ab2ccf67d9f6d173b...44de125c96b5f5c4a83265cab16cc9cd189703db`.

Evidence: `decode_context_u32` verifies the primitive context tag, bounds the payload, decodes it, and `try_from`s to `u32` before either property type is constructed (`crates/bacnet-services/src/common.rs:12-30`). Both `[0]` identifiers and optional `[1]` indexes take that path (`common.rs:63-74`, `common.rs:133-144`). Thus a `u32::MAX + 1` or `u64::MAX` fails while the wider leading-zero representation of `u32::MAX` remains valid (`common.rs:309-360`). The optional-index peeks use `is_context`, so parent closing tags are not consumed; value `[2]` and priority `[3]` framing remains owned by `BACnetPropertyValue` (`common.rs:67-76`, `common.rs:146-194`).

Data-flow check: RPM decodes references before handler lookups (`rpm.rs:89`; `bacnet-server/src/handlers/read_property.rs:71-80`); WPM parses all values before validation or commits (`wpm.rs:86`; `handlers/write_property.rs:231-278`); CreateObject parses initial values before the handler adds/mutates an object (`object_mgmt.rs:123`; `handlers/object_mgmt.rs:55-55,133-164`); COV notification delivery decodes before enqueueing (`cov.rs:369`; `bacnet-client/src/client/dispatch.rs:87-100,166-179`); COV-multiple subscriptions decode the shared reference before validation/subscription insertion (`cov_multiple.rs:219`; `handlers/cov.rs:243-243,320-381`); and audit target properties use the same decoder (`audit.rs:257`).

Residual baseline risks (not introduced by this one-file PR): the independently implemented SubscribeCOVProperty reference still narrows both fields with `as u32` (`crates/bacnet-services/src/cov.rs:215-225`) before `handle_subscribe_cov_property` validates and inserts/cancels a subscription (`crates/bacnet-server/src/handlers/cov.rs:134-143,154-201`). A wire property ID of `2^32 + 85` aliases property 85 there. The RPM ACK and COVNotificationMultiple codecs likewise retain pre-existing direct narrowings (`rpm.rs:212-224`; `cov_multiple.rs:461-470`); the former is returned directly by client RPM reads (`bacnet-client/src/client/property.rs:103-125`). These are separate cleanup targets, not regressions in this compare.

jscott3201 (commented):
[Adversary] READY

Reviewed exactly `6079bb71a9a7736b0e737f1ab2ccf67d9f6d173b...44de125c96b5f5c4a83265cab16cc9cd189703db`; the PR base/head still match the dispatch. No material hostile-wire regression found.

Verified by inspection:
- `crates/bacnet-services/src/common.rs:12-35` rejects non-context/non-`[0|1]` fields, empty or >8-octet unsigned payloads (via `decode_unsigned`), and all decoded values above `u32::MAX`; it retains leading-zero representations whose numeric value fits.
- `crates/bacnet-services/src/common.rs:64-88` and `:134-205` preserve optional-index peeking and return offsets: an absent `[1]` leaves the next property/parent closer unconsumed; a present primitive context `[1]` is consumed; `BACnetPropertyValue` still owns only value `[2]` and optional priority `[3]`.
- Application, opening, and closing tag forms fail `is_context`; value `[2]` still requires an opening tag and `extract_context_value` consumes its matching closing tag. Outer constructed-tag matching remains the separate #327 concern, not a regression introduced here.

Non-blocking coverage gap: `common.rs` adds direct overflow/max/wrong-`[0]` tests, but does not explicitly table-test 0-8 byte fields (notably 6-8 byte leading-zero in-range values), application/opening/closing variants, or a shared-decoder call embedded between adjacent properties and its enclosing closer. The implementation itself covers these cases through the shared helper and existing caller framing.

Commands run: none. Evidence was SHA-pinned GitHub PR metadata/diff plus source inspection of `common.rs`, `tags.rs`, `primitives/mod.rs`, and RPM/WPM/COV/object-management/COV-multiple/audit callers; no test execution is claimed.

jscott3201 (commented):
[Runtime] READY

No material runtime or test defect survived review of `6079bb71a9a7736b0e737f1ab2ccf67d9f6d173b...44de125c96b5f5c4a83265cab16cc9cd189703db`.

Evidence:
- `crates/bacnet-services/src/common.rs:18-32` requires the expected primitive context tag, checks the payload boundary, decodes the complete unsigned value, and uses checked `u32::try_from`; this rejects overflow aliases while accepting numerically valid wider encodings with leading zeroes.
- `crates/bacnet-services/src/common.rs:63-75` and `crates/bacnet-services/src/common.rs:133-146` preserve absolute offsets and leave the offset unchanged when optional array tag 1 is absent. `crates/bacnet-services/src/common.rs:148-200` leaves value framing and optional priority handling intact.
- `crates/bacnet-services/src/common.rs:279-306` retains priority range and leading-zero compatibility. `crates/bacnet-services/src/common.rs:310-360` covers `u32::MAX` in a five-octet leading-zero form, property/index overflow at `u32::MAX + 1` and `u64::MAX`, and wrong mandatory tag 0; `crates/bacnet-services/src/common.rs:367-440` retains malformed/truncation coverage.
- Parent loops begin each child at its field tag and retain the returned offset: RPM `crates/bacnet-services/src/rpm.rs:83-91`, WPM `crates/bacnet-services/src/wpm.rs:81-88`, COV `crates/bacnet-services/src/cov.rs:364-371`, COV-multiple `crates/bacnet-services/src/cov_multiple.rs:206-222`, audit `crates/bacnet-services/src/audit.rs:251-260`, and CreateObject `crates/bacnet-services/src/object_mgmt.rs:119-125`. Thus an absent optional field leaves the next property or parent closing tag for the parent, while present fields advance exactly through their payload.

Verification:
- Re-read the PR before submission; base/head still matched the review packet.
- GitHub checks observed successful: `Test (ubuntu-latest)`, `Clippy`, `Rustfmt`, file-size cap, and no-secret scan.
- Local shell commands were not run because this reviewer had no command-execution tool.

Residual gaps:
- The new malformed-value cases are unit-level rather than duplicated through every parent decoder; unchanged parent round-trip tests plus the shared call paths above provide indirect coverage.
- Nested parent closing-tag-number validation remains excluded as issue #327, as requested.

PR #330, opened 2026-08-14
fix(services): validate RPM result widths

jscott3201:
## Summary
- check ReadPropertyMultiple ACK property identifiers and optional array indexes against their existing `u32` fields
- require mandatory object context tag 0 on RPM requests and ACKs
- cover overflow aliases, `u64::MAX`, exact maxima in wider leading-zero forms, and wrong object tags

## Why
The ACK decoder converted peer `u64` property identifiers and array indexes with `as`, allowing malformed responses to alias a different property or array element before being returned from the client API. Both RPM top-level decoders also accepted an object identifier under the wrong context tag.

## Testing
- `cargo test -p bacnet-services --locked rpm`
- `git diff --check`

Related to #309.

### DIFF
--- a/crates/bacnet-services/src/rpm.rs
+++ b/crates/bacnet-services/src/rpm.rs
@@ -54,6 +54,12 @@ impl ReadPropertyMultipleRequest {
 
             // [0] object-identifier
             let (tag, pos) = tags::decode_tag(data, offset)?;
+            if !tag.is_context(0) {
+                return Err(Error::decoding(
+                    offset,
+                    "RPM request expected context tag 0",
+                ));
+            }
             let end = pos + tag.length as usize;
             if end > data.len() {
                 return Err(Error::decoding(pos, "RPM request truncated at object-id"));
@@ -173,6 +179,9 @@ impl ReadPropertyMultipleACK {
 
             // [0] object-identifier
             let (tag, pos) = tags::decode_tag(data, offset)?;
+            if !tag.is_context(0) {
+                return Err(Error::decoding(offset, "RPM ACK expected context tag 0"));
+            }
             let end = pos + tag.length as usize;
             if end > data.len() {
                 return Err(Error::decoding(pos, "RPM ACK truncated at object-id"));
@@ -209,7 +218,9 @@ impl ReadPropertyMultipleACK {
                 if end > data.len() {
                     return Err(Error::decoding(tag_end, "RPM ACK truncated at property-id"));
                 }
-                let prop_raw = primitives::decode_unsigned(&data[tag_end..end])? as u32;
+                let prop_raw = primitives::decode_unsigned(&data[tag_end..end])?;
+                let prop_raw = u32::try_from(prop_raw)
+                    .map_err(|_| Error::decoding(tag_end, "RPM ACK property-id exceeds u32"))?;
                 let property_identifier = PropertyIdentifier::from_raw(prop_raw);
                 offset = end;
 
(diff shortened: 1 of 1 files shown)


jscott3201 (commented):
[Dataflow] READY

Findings: none.

Evidence:
- `crates/bacnet-services/src/rpm.rs:221-238` decodes each ACK property ID/index before conversion and rejects values above `u32::MAX`; thus no narrowing alias can reach `ReadResultElement` or the client API. The exact-max, leading-zero, `u32::MAX + 1`, and `u64::MAX` paths are covered at `rpm.rs:520-536`.
- Both mandatory object fields now require primitive context tag 0 (`rpm.rs:55-68`, `rpm.rs:180-190`), matching the request and ACK producers at `rpm.rs:28-36` and `rpm.rs:141-169`. The server decodes requests before processing (`crates/bacnet-server/src/handlers/read_property.rs:66-75`) and emits the same ACK model at `read_property.rs:173-183`.
- Both direct and routed client paths return `ReadPropertyMultipleACK::decode` directly (`crates/bacnet-client/src/client/property.rs:103-125`, `136-155`); Python then exposes the decoded property ID and index unchanged (`crates/rusty-bacnet/src/types/rpm_wpm.rs:37-45`). No post-decode conversion can reintroduce aliasing.
- Result sequence framing remains bounded and advances only through matching element/list boundaries (`rpm.rs:199-210`, `240-263`); server iteration preserves its request-spec/property traversal order (`read_property.rs:74-80`, `173-176`).

Residual risk (baseline, not introduced here): `PropertyIdentifier` remains a `u32` while its documented BACnet wire field is 22-bit (`crates/bacnet-types/src/enums/property_id.rs:12-16`). This PR intentionally preserves the existing `u32` compatibility contract, so IDs in `4_194_304..=u32::MAX` remain accepted as non-aliased but nonconformant values.

jscott3201 (commented):
[Adversary] READY

Reviewed exactly `8b1a1527a8da94182ae05da7faf7ea65dc070779...0b172e23b6cfcdc70c1bb4ac219b275be24bef06`; base and head were re-verified immediately before submission.

No material defect found in the unsigned-narrowing or object-tag changes. `rpm.rs:221-223` and `rpm.rs:235-238` reject every decoded numeric value above `u32::MAX` before constructing the returned property ID/index, eliminating low-word aliases. Zero-byte values are rejected by `decode_unsigned`; 1–8-byte values, leading-zero encodings, omitted indexes, and both success/error paths reach the same checked conversions. `rpm.rs:57-62` and `rpm.rs:182-184` require primitive context tag 0, rejecting application, wrong-context, opening, and closing object tags. Decoder state is reset per element/object (`rpm.rs:175-297`), so duplicate/trailing elements fail at their expected tags rather than carrying state across boundaries.

Non-blocking coverage gap: `rpm.rs:519-550` tests one five-byte leading-zero maximum, two eight-byte overflow cases, and one wrong context tag per decoder. It does not enumerate zero through eight byte widths, a concrete low-word alias such as `00 00 00 01 00 00 00 55`, application/open/close object tags, error-result-with-index, duplicate/trailing fields, or multi-object ACK boundaries. This is a regression-test breadth gap, not a demonstrated implementation defect.

#327 is excluded: the nested context-closing validation problem is pre-existing and outside this diff.

Evidence commands/queries: `github_pull_request_read(get|get_diff|get_files|get_commits)`, `github_get_commit(detail=full_patch)`, and `github_get_file_contents` for `rpm.rs`, `tags.rs`, `primitives/mod.rs`, `common.rs`, and `primitives.rs`. No shell test command was run in this read-only review environment.

jscott3201 (commented):
[Runtime] READY

No material runtime defects found in `8b1a1527a8da94182ae05da7faf7ea65dc070779...0b172e23b6cfcdc70c1bb4ac219b275be24bef06`.

- `crates/bacnet-services/src/rpm.rs:54-66` and `crates/bacnet-services/src/rpm.rs:180-190` reject non-primitive/non-context-0 object fields before decoding their payloads.
- `crates/bacnet-services/src/rpm.rs:213-238` replaces lossy property-ID and array-index casts with checked `u64`→`u32` conversions, so out-of-range peer values cannot alias valid fields.
- `crates/bacnet-services/src/rpm.rs:359-373` constructs complete ACK frames, and `crates/bacnet-services/src/rpm.rs:520-550` exercises exact `u32::MAX` with a wider leading-zero encoding, both overflow sites at `u32::MAX + 1` and `u64::MAX`, and wrong object tags for request and ACK decoding.
- The surrounding result loop and success/error choice paths retain their offsets and allocation bounds (`crates/bacnet-services/src/rpm.rs:175-293`); the change adds no concurrency or resource-lifetime behavior.

GitHub's Test, Clippy, Rustfmt, file-size, and no-secret checks passed. I did not execute local test commands in this review environment.

PR #331, opened 2026-08-14
fix(services): validate list element widths

jscott3201:
## Summary
- check AddListElement/RemoveListElement property identifiers and optional array indexes against their existing `u32` fields
- require mandatory object/property context tags and opening tag 3 for the element list
- reject bytes after the constructed list field
- cover overflow aliases, `u64::MAX`, exact maxima in wider leading-zero forms, wrong tags, and trailing data

## Why
The shared decoder converted peer `u64` identifiers and indexes with `as`, allowing malformed requests to redirect a list operation to another property or array element before the server performed lookup, read, or mutation. It also accepted wrong mandatory tags and ignored trailing fields.

## Testing
- `cargo test -p bacnet-services --locked list_manipulation`
- `cargo test -p bacnet-server --locked handlers::list`
- `git diff --check`

Related to #309.

### DIFF
--- a/crates/bacnet-services/src/list_manipulation.rs
+++ b/crates/bacnet-services/src/list_manipulation.rs
@@ -39,6 +39,12 @@ impl ListElementRequest {
 
         // [0] objectIdentifier
         let (tag, pos) = tags::decode_tag(data, offset)?;
+        if !tag.is_context(0) {
+            return Err(Error::decoding(
+                offset,
+                "ListElement request expected context tag 0",
+            ));
+        }
         let end = pos + tag.length as usize;
         if end > data.len() {
             return Err(Error::buffer_too_short(end, data.len()));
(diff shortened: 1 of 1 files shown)


jscott3201 (commented):
[Dataflow] READY

Findings: none.

Evidence:
- Client construction continues to use the existing `u32` API and canonical context encoding (`crates/bacnet-client/src/client/file_list.rs:143-160, 174-191`; `crates/bacnet-services/src/list_manipulation.rs:22-35`).
- The shared request decoder validates tags 0/1, performs fallible `u64`→`u32` conversions for the property and optional array index, requires opening tag 3, and rejects any bytes after its matching close before returning a request (`crates/bacnet-services/src/list_manipulation.rs:40-103`). Thus an oversized value, wrong mandatory tag, malformed constructed list, or trailing field cannot reach object lookup.
- Both dispatch arms pass the complete service payload to their handlers (`crates/bacnet-server/src/server/requests/mod.rs:306-318`), and each handler decodes it before `db.get_mut` (`crates/bacnet-server/src/handlers/list.rs:52-64, 133-152`).
- For ordinary lists, all element frames are decoded before the sole write; for framed destination lists, both stored and incoming frames are decoded and the replacement is constructed before write-back (`crates/bacnet-server/src/handlers/list.rs:26-43, 96-124, 168-187`). The malformed-tail and framed-payload tests cover the no-partial-commit outcome (`crates/bacnet-server/src/handlers/list.rs:241-294, 470-503`).
- The new decoder tests exercise widened leading-zero maxima, overflow values, wrong tags, and trailing data (`crates/bacnet-services/src/list_manipulation.rs:165-210`).

Residual risk / baseline: the established `PropertyIdentifier` representation remains `u32`, so values within `u32` but outside BACnet's narrower property-identifier domain are still decoded and deferred to object/property validation. This PR intentionally preserves that API/compatibility policy; it is not introduced by this change.

jscott3201 (commented):
[Adversary] READY

No material defect found in the verified target `d65220d3f016b8f27e845816bddf869590c30801...a83fb551537cd45f8b4d57fe9f5d1677563853d4`.

Verified behavior:
- Required field ownership is now enforced for `[0]`, `[1]`, and constructed `[3]`; bytes after the `[3]` close are rejected (`crates/bacnet-services/src/list_manipulation.rs:42-47`, `:57-62`, `:88-101`).
- Both peer-controlled numeric fields are decoded as `u64` and checked before entering the existing `u32` model, so e.g. `[00 00 00 01 00 00 00 00]` and `u64::MAX` now fail instead of narrowing (`crates/bacnet-services/src/list_manipulation.rs:67-70`, `:78-82`; `crates/bacnet-encoding/src/primitives/mod.rs:65-76`).
- Tests cover an in-range leading-zero maximum, overflow aliases, `u64::MAX`, wrong mandatory tags, and a trailing field (`crates/bacnet-services/src/list_manipulation.rs:165-218`).

Residual coverage gaps (non-blocking): the new unit tests do not enumerate zero-length contents or every 1–8-octet accepted/rejected representation for both fields. The shared unsigned decoder rejects zero-length and accepts only 1–8 octets (`crates/bacnet-encoding/src/primitives/mod.rs:65-76`), so this is a coverage gap rather than a demonstrated defect.

Scope exclusion: nested constructed closing-tag mismatches inside `[3]` remain the separate pre-existing #327 behavior in `tags::extract_context_value` (`crates/bacnet-encoding/src/tags.rs:282-311`); it is not introduced by this one-file change.

jscott3201 (commented):
[Runtime] READY

No material runtime defect found in `d65220d3f016b8f27e845816bddf869590c30801...a83fb551537cd45f8b4d57fe9f5d1677563853d4`.

Evidence checked:
- Property identifiers and optional array indexes are decoded as `u64` and converted with fallible `u32::try_from`, so `u32::MAX + 1` and `u64::MAX` fail instead of aliasing a lower target (`crates/bacnet-services/src/list_manipulation.rs:55-83`).
- Mandatory context/opening tags and complete request consumption are enforced before a request is returned (`crates/bacnet-services/src/list_manipulation.rs:40-100`).
- Both mutation paths decode the full request before database lookup, property reads, or writes (`crates/bacnet-server/src/handlers/list.rs:52-66`, `crates/bacnet-server/src/handlers/list.rs:133-150`).
- Constructed-list scanning is linear in the supplied payload and bounded to 32 nesting levels (`crates/bacnet-encoding/src/tags.rs:273-329`, `crates/bacnet-encoding/src/tags.rs:153-158`).
- Added tests exercise the exact maximum in a wider form, both overflow boundaries for both fields, wrong mandatory tags, and trailing bytes (`crates/bacnet-services/src/list_manipulation.rs:165-210`).

Verification: the exact head SHA has successful GitHub `Test (ubuntu-latest)`, `Clippy`, and `Rustfmt` checks. The test job runs the workspace suite with the full Linux feature set (`.github/workflows/ci.yml:126-153`).

PR #333, opened 2026-08-14
fix(services): validate text message widths

jscott3201:
## Summary
- check TextMessage numeric class and priority values against their existing `u32` fields
- require source-device, priority, and message context tags
- validate the optional message-class CHOICE is present, bounded, recognized, and singular
- reject trailing top-level data
- cover overflow aliases, `u64::MAX`, exact maxima in wider leading-zero forms, malformed CHOICE payloads, wrong tags, and trailing bytes

## Why
The decoder narrowed peer values with `as` in both the numeric message-class alternative and message priority, allowing over-wide inputs to alias other values. Malformed inner lengths could also panic during slicing, and wrong or trailing fields were accepted by both confirmed and unconfirmed TextMessage handlers.

This preserves the shipped `u32` enum wrappers rather than introducing narrower protocol-domain enforcement.

## Testing
- `cargo test -p bacnet-services --locked text_message`
- `git diff --check`

Related to #309.

### DIFF
--- a/crates/bacnet-services/src/text_message.rs
+++ b/crates/bacnet-services/src/text_message.rs
@@ -62,6 +62,12 @@ impl TextMessageRequest {
 
         // [0] textMessageSourceDevice
         let (tag, pos) = tags::decode_tag(data, offset)?;
+        if !tag.is_context(0) {
+            return Err(Error::decoding(
+                offset,
+                "TextMessage expected context tag 0",
+            ));
+        }
         let end = pos + tag.length as usize;
         if end > data.len() {
             return Err(Error::decoding(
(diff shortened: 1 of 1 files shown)


jscott3201 (commented):
[Dataflow] READY

No material PR-introduced defects found.

Evidence:
- The decoder now owns each required field tag, bounds the inner message-class value before slicing, consumes exactly one recognized CHOICE alternative, and rejects extra class or top-level bytes (`crates/bacnet-services/src/text_message.rs:65-78, 85-127, 133-167`). Thus over-wide numeric class/priority values return decoding errors rather than narrowing aliases (`:101-110, :146-149`), and malformed inner lengths cannot reach the prior unchecked slice.
- Producer encoding remains the accepted shape: `[1]` wraps context `[0]` unsigned or `[1]` character string; priority is context `[2]`; message is context `[3]` (`crates/bacnet-services/src/text_message.rs:37-56, 81-127, 131-167`).
- Both service paths reach the same handler (`crates/bacnet-server/src/handlers/alarm_event/text_message.rs:11-15`). Confirmed decode success yields SimpleACK and decode failure yields an Error APDU (`crates/bacnet-server/src/server/requests/mod.rs:336-340`); unconfirmed decode failure is logged and returns without a response (`crates/bacnet-server/src/server/requests/unconfirmed.rs:163-176`).

Baseline / residual risk: this change deliberately retains the existing `u32` enum-wrapper domain, so non-aliased priority values outside the named normal/urgent constants remain representable. That is pre-existing behavior, not a narrowing or panic path introduced here. Review was source-based; I did not execute the test suite.

jscott3201 (commented):
[Adversary] READY

Reviewed exactly `d2cf6ee0d3aa1cf46fc86af90d86ff75f48ae8b1...9de1a0dffd27bbc020ac12c3da6ba3e9927a29f5` (PR head/base rechecked immediately before submission). No material hostile-wire-input defect found.

Verified behavior: the shared confirmed/unconfirmed decoder rejects non-owned source/priority/message tags and top-level trailing bytes (`crates/bacnet-services/src/text_message.rs:64-79,132-167`); class contents must be nonempty, a single recognized primitive member, and wholly consumed (`:84-127`); and the `u64` decoder's 1–8-byte result is fallibly narrowed to `u32` for both numeric fields (`:101-110,146-149`; `crates/bacnet-encoding/src/primitives/mod.rs:65-77`). Truncated or mismatched constructed data returns an error from `extract_context_value` rather than being sliced (`crates/bacnet-encoding/src/tags.rs:273-334`). Invalid/unknown character encodings similarly propagate an error from `decode_character_string` (`crates/bacnet-encoding/src/primitives/mod.rs:187-230`).

Non-blocking coverage gaps: `text_message_values_must_fit_u32` covers a 5-byte leading-zero maximum and 8-byte overflow values, but not the requested 0- and each 1–8-byte class/priority matrix (`text_message.rs:255-276`). `text_message_rejects_malformed_message_class` has empty/truncated/multiple/unknown cases but no malformed character-string or mismatched-closing-tag case (`:279-286`); the sole text-class test is a valid round trip (`:227-238`). These omissions do not reveal an acceptance path in the inspected decoder.

Checks/evidence: GitHub `Test (ubuntu-latest)` and `Clippy` are successful for this head. Local test commands were not executed in this read-only review environment. Scope is PR #333 only; #327 was not assessed.

jscott3201 (commented):
[Runtime] READY

No material runtime defect found in `d2cf6ee0d3aa1cf46fc86af90d86ff75f48ae8b1...9de1a0dffd27bbc020ac12c3da6ba3e9927a29f5`.

Evidence:
- `crates/bacnet-services/src/text_message.rs:64-79` validates the required source tag and bounds before slicing.
- `crates/bacnet-services/src/text_message.rs:84-127` uses the decoded opening-tag endpoint, bounds the inner value, rejects empty/unknown/non-singular CHOICE content, and uses checked `u64`→`u32` conversion.
- `crates/bacnet-services/src/text_message.rs:132-167` validates both remaining required tags, checks priority width, bounds payload access, and rejects trailing top-level data.
- `crates/bacnet-services/src/text_message.rs:255-310` exercises maximum/overflow encodings, malformed CHOICE shapes, wrong required tags, and trailing bytes. The shared decoder is called by both confirmed and unconfirmed dispatch paths.

The current head’s test, Clippy, and formatting checks passed.

PR #335, opened 2026-08-14
fix(services): harden enrollment summary decoding

jscott3201:
## Summary
- replace all 11 unchecked GetEnrollmentSummary request/ACK narrowing conversions with checked field-width conversions
- validate mandatory request tags, nested recipient/process and priority-filter members, priority ordering, and complete request consumption
- validate all five ACK application datatypes and reject reserved application L/V/T forms
- preserve public `u32` enum wrappers, unknown in-range values, and wider leading-zero encodings
- add grouped boundary, alias, `u64::MAX`, nested framing, wrong-tag, and trailing-data regressions

## Why
This service family accepted peer values above each destination width and silently aliased them into request filters or client-visible ACK entries. Its nested filters also defaulted or skipped malformed members, and ACK entries were interpreted by position without validating application datatypes.

The handler filtering behavior is unchanged; this PR is limited to request and response decode boundaries.

## Testing
- `cargo test -p bacnet-services --locked enrollment_summary`
- `git diff --check`

Related to #309.

### DIFF
--- a/crates/bacnet-services/src/enrollment_summary.rs
+++ b/crates/bacnet-services/src/enrollment_summary.rs
@@ -9,6 +9,10 @@ use bytes::BytesMut;
 
 use crate::common::MAX_DECODED_ITEMS;
 
+fn is_application_tag(tag: &tags::Tag, header: u8, number: u8) -> bool {
+    tag.class == tags::TagClass::Application && tag.number == number && header & 0x07 <= 5
+}
+
 // ---------------------------------------------------------------------------
 // GetEnrollmentSummaryRequest
 // ---------------------------------------------------------------------------
@@ -92,6 +96,12 @@ impl GetEnrollmentSummaryRequest {
 
         // [0] acknowledgmentFilter
         let (tag, pos) = tags::decode_tag(data, offset)?;
+        if !tag.is_context(0) {
+            return Err(Error::decoding(
+                offset,
+                "EnrollmentSummary expected context tag 0",
+            ));
+        }
         let end = pos + tag.length as usize;
         if end > data.len() {
             return Err(Error::decoding(
@@ -99,7 +109,10 @@ impl GetEnrollmentSummaryRequest {
                 "EnrollmentSummary truncated at acknowledgmentFilter",
             ));
         }
-        let acknowledgment_filter = primitives::decode_unsigned(&data[pos..end])? as u32;
+        let acknowledgment_filter = primitives::decode_unsigned(&data[pos..end])?;
+        let acknowledgment_filter = u32::try_from(acknowledgment_filter).map_err(|_| {
+            Error::decoding(pos, "EnrollmentSummary acknowledgmentFilter exceeds u32")
+        })?;
         offset = end;
 
         // [1] enrollmentFilter (optional, constructed)
@@ -107,39 +120,68 @@ impl GetEnrollmentSummaryRequest {
         if offset < data.len() {
             let (tag, tag_end) = tags::decode_tag(data, offset)?;
             if tag.is_opening_tag(1) {
-                // Parse BACnetRecipientProcess
                 let mut device = None;
-                let mut process_id = 0u32;
                 let (content, new_offset) = tags::extract_context_value(data, tag_end, 1)?;
                 let mut inner_offset = 0;
-                while inner_offset < content.len() {
+
+                if inner_offset < content.len() {
                     let (inner_tag, inner_pos) = tags::decode_tag(content, inner_offset)?;
                     if inner_tag.is_opening_tag(0) {
-                        // recipient CHOICE — parse device [0]
                         let (recipient_content, recipient_end) =
                             tags::extract_context_value(content, inner_pos, 0)?;
-                        if !recipient_content.is_empty() {
-                            let (dev_tag, dev_pos) = tags::decode_tag(recipient_content, 0)?;
-                            if dev_tag.is_context(0) {
-                                let dev_end = dev_pos + dev_tag.length as usize;
-                                if dev_end <= recipient_content.len() {
-                                    device = Some(ObjectIdentifier::decode(
-                                        &recipient_content[dev_pos..dev_end],
-                                    )?);
-                                }
-                            }
+                        let (dev_tag, dev_pos) = tags::decode_tag(recipient_content, 0)?;
+                        if !dev_tag.is_context(0) {
+                            return Err(Error::decoding(
+                                tag_end + inner_pos,
+                                "EnrollmentSummary expected recipient device tag 0",
+                            ));
                         }
+                        let dev_end = dev_pos + dev_tag.length as usize;
+                        if dev_end > recipient_content.len() {
+                            return Err(Error::decoding(
+                                tag_end + dev_pos,
+                                "EnrollmentSummary truncated at recipient device",
+                            ));
+                        }
+                        if dev_end != recipient_content.len() {
+                            return Err(Error::decoding(
+                                tag_end + dev_end,
+                                "EnrollmentSummary recipient has trailing data",
+                            ));
+                        }
+                        device = Some(ObjectIdentifier::decode(
+                            &recipient_content[dev_pos..dev_end],
+                        )?);
                         inner_offset = recipient_end;
-                    } else if inner_tag.is_context(1) {
-                        let inner_end = inner_pos + inner_tag.length as usize;
-                        if inner_end <= content.len() {
-                            process_id =
-                                primitives::decode_unsigned(&content[inner_pos..inner_end])? as u32;
-                        }
-                        inner_offset = inner_end;
-                    } else {
-                        inner_offset = inner_pos + inner_tag.length as usize;
                     }
+                }
+
+                let (process_tag, process_pos) = tags::decode_tag(content, inner_offset)?;
+                if !process_tag.is_context(1) {
+                    return Err(Error::decoding(
+                        tag_end + inner_offset,
+                        "EnrollmentSummary expected processIdentifier tag 1",
+                    ));
+                }
+                let process_end = process_pos + process_tag.length as usize;
+                if process_end > content.len() {
+                    return Err(Error::decoding(
+                        tag_end + process_pos,
+                        "EnrollmentSummary truncated at processIdentifier",
+                    ));
+                }
+                let process_id = primitives::decode_unsigned(&content[process_pos..process_end])?;
+                let process_id = u32::try_from(process_id).map_err(|_| {
+                    Error::decoding(
+                        tag_end + process_pos,
+                        "EnrollmentSummary processIdentifier exceeds u32",
+                    )
+                })?;
+                if process_end != content.len() {
+                    return Err(Error::decoding(
+                        tag_end + process_end,
+                        "EnrollmentSummary enrollmentFilter has trailing data",
+                    ));
                 }
                 enrollment_filter = Some(RecipientProcess {
                     device,
@@ -153,9 +195,10 @@ impl GetEnrollmentSummaryRequest {
         let mut event_state_filter = None;
         let (opt_data, new_offset) = tags::decode_optional_context(data, offset, 2)?;
         if let Some(content) = opt_data {
-            event_state_filter = Some(EventState::from_raw(
-                primitives::decode_unsigned(content)? as u32
-            ));
+            let value = primitives::decode_unsigned(content)?;
+            event_state_filter = Some(EventState::from_raw(u32::try_from(value).map_err(
+                |_| Error::decoding(offset, "EnrollmentSummary eventStateFilter exceeds u32"),
+            )?));
             offset = new_offset;
         }
 
@@ -163,9 +206,10 @@ impl GetEnrollmentSummaryRequest {
         let mut event_type_filter = None;
         let (opt_data, new_offset) = tags::decode_optional_context(data, offset, 3)?;
         if let Some(content) = opt_data {
-            event_type_filter = Some(EventType::from_raw(
-                primitives::decode_unsigned(content)? as u32
-            ));
+            let value = primitives::decode_unsigned(content)?;
+            event_type_filter = Some(EventType::from_raw(u32::try_from(value).map_err(|_| {
+                Error::decoding(offset, "EnrollmentSummary eventTypeFilter exceeds u32")
+            })?));
             offset = new_offset;
         }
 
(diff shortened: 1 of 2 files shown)


jscott3201 (commented):
[Adversary] NOT READY

### P1 — The hardened enrollment filter rejects the conformant `BACnetRecipientProcess` wire form and emits a nonconformant one

`BACnetRecipientProcess` contains an untagged `BACnetRecipient` followed by an untagged `Unsigned32`. A device recipient is a *primitive* context tag `[0]` containing the four object-id octets; the repository's shared recipient codec documents and implements that form at `crates/bacnet-encoding/src/constructed/recipient.rs:16-28,65-79`.

This decoder instead only recognizes an opening `[0]` wrapper and then requires a context `[1]` process identifier (`crates/bacnet-services/src/enrollment_summary.rs:127-164`); its encoder emits that same double-wrapped/context-tagged layout (`crates/bacnet-services/src/enrollment_summary.rs:62-71`). Consequently, a standards-form request such as `09 00 1E 0C 02 00 00 01 21 07 1F` (ack filter 0; enrollment filter with device 8:1 and process 7) reaches line 159 with `0C` and is rejected as not context tag 1. Conversely, an outbound device filter is serialized with `0E 0C … 0F 19 …`, which a conformant peer will not parse as `BACnetRecipientProcess`.

The new regression also explicitly accepts an enrollment filter with no recipient at `crates/bacnet-services/src/enrollment_summary_width_tests.rs:107-124`, even though the recipient is mandatory. This makes the service filter non-interoperable in both directions while the PR claims nested recipient/process validation. Decode and encode the actual `BACnetRecipientProcess` grammar (and reject a missing recipient) rather than pinning the proprietary layout.

jscott3201 (commented):
[Dataflow] NOT READY

**P1 – this changes a published public Rust API without a SemVer/migration declaration.** `RecipientProcess::device` is now a required `ObjectIdentifier` at `crates/bacnet-services/src/enrollment_summary.rs:32-36`, whereas the reviewed base exposes `Option<ObjectIdentifier>`. A downstream `bacnet-services = "0.10.1"` consumer that previously constructed the public filter as `RecipientProcess { device: Some(device), process_identifier: 7 }` now fails to compile. The workspace continues to declare version `0.10.1` (`Cargo.toml:42-61`) and the changelog says the project follows SemVer (`CHANGELOG.md:3-8`), but this PR neither changes the version nor documents this breaking migration. The required device may be the correct way to prevent emission of the formerly representable invalid no-recipient form, but it must be shipped as an explicitly declared breaking API change (or retain source compatibility through an appropriate transition).

Verified data flow (not separate findings):
- The new request bytes correctly place the device CHOICE as primitive context `[0]` followed by application `Unsigned` process ID inside `[1]` (`enrollment_summary.rs:61-65`); decoding mirrors that framing and rejects trailing nested data (`118-165`).
- Checked conversions reject values that cannot fit their public destination types in both request and ACK decoding (`106-109`, `148-154`, `213-242`, `372-437`); application ACK tags reject context/reserved L/V/T forms (`12-14`, `341-346`, `359-364`, `380-385`, `401-406`, `421-426`).
- Handler semantics are unchanged: it decodes the request (`get_enrollment_summary.rs:20`), applies only state, notification-class, and priority filters (`44-86`), then encodes the ACK (`102-103`). The Python producer/consumer builds the request and decodes that ACK at `enrollment_alarm_covmulti_who_writegroup.rs:45-59`; its public Python signature does not expose `RecipientProcess`.

jscott3201 (commented):
[Adversary] READY

Reviewed exactly `9c60d4bdf8f5a10d86ddfd391a1c31bbf43a5a90...dea44b384963f0f8a7d06e48c6a9f292f0821421`.

No material diff defects found. The device-form RecipientProcess wire sequence is now context `[0]` object identifier followed by application Unsigned (`crates/bacnet-services/src/enrollment_summary.rs:61-65`); decoding rejects missing/unsupported/malformed recipient or process members and trailing nested bytes (`:118-159`). Checked conversion boundaries preserve in-range leading-zero aliases while rejecting overflow for every narrowed request/ACK field (`:106-109`, `:148-154`, `:213-242`, `:372-437`). The ACK path validates each required application tag including reserved L/V/T forms, rejects partial entries through fallible tag/bounds decoding, and retains the 10,000-entry cap (`:331-437`).

jscott3201 (commented):
[Dataflow] READY — no material defects found in `9c60d4b...ca4f18d`.

Evidence:
- Public `Option<ObjectIdentifier>` compatibility remains (`crates/bacnet-services/src/enrollment_summary.rs:27-35`); `device=None` emits no enrollment filter (`:60-71`), while `Some(device)` emits `[1] / [0] object-id / application Unsigned / [/1]`, matching the shared recipient framing (`crates/bacnet-encoding/src/constructed/recipient.rs:65-80`).
- Request decoding performs checked conversions before constructing public fields and rejects non-consumed/nested trailing input (`enrollment_summary.rs:107-170`, `:196-285`).
- ACK tag validation and checked widths occur before entries reach Rust/Python consumers (`enrollment_summary.rs:341-445`); Python forwards the decoded typed fields without narrowing (`crates/rusty-bacnet/src/client/client_methods/enrollment_alarm_covmulti_who_writegroup.rs:59-85`).
- Regression coverage includes exact recipient bytes, `device=None` omission, aliases/`u64::MAX`, reserved L/V/T forms, and trailing data (`crates/bacnet-services/src/enrollment_summary_width_tests.rs:61-258`).

Scope separation: the server handler’s filter/output semantics are unchanged by this compare; it receives the decoded request at `crates/bacnet-server/src/handlers/alarm_event/get_enrollment_summary.rs:20` and its logic was not modified.

jscott3201 (commented):
[Adversary] NOT READY

- **High — device-less filters are silently widened** (`crates/bacnet-services/src/enrollment_summary.rs:64-71`). `RecipientProcess.device` remains public as an `Option`, but `encode` emits the entire `[1] enrollmentFilter` only when it is `Some`. A caller constructing `enrollment_filter: Some(RecipientProcess { device: None, process_identifier: 7 })` receives no encode error (the method returns `()`); the wire request has no enrollment filter, so a conforming peer processes it as an unfiltered request rather than as the requested recipient/process filter. The new regression at `crates/bacnet-services/src/enrollment_summary_width_tests.rs:151-172` explicitly asserts that loss of constraint. Reject this unrepresentable value or model/encode the missing recipient alternative instead of silently changing a targeted query into a broad one.

jscott3201 (commented):
[Runtime] NOT READY

1. **P1 — `enrollmentFilter` no longer uses the Clause 21 `BACnetRecipientProcess` framing** (`crates/bacnet-services/src/enrollment_summary.rs:64-67`, `crates/bacnet-services/src/enrollment_summary.rs:121-151`). `BACnetRecipientProcess` is `recipient [0] BACnetRecipient` followed by `process-identifier [1] Unsigned32`; for the device choice that means opening `[0]`, primitive device `[0]`, closing `[0]`, then context unsigned `[1]`. This implementation instead emits/accepts the device choice directly and an application Unsigned. A conforming request for Device,7/process 7 is `09 00 1E 0E 0C 02 00 00 07 0F 19 07 1F`; decoding it fails at the `is_context(0)` check because the next tag is the required opening `[0]`. Conversely, the bytes emitted here are not accepted as a `BACnetRecipientProcess` by conforming peers. The asserted fixture at `crates/bacnet-services/src/enrollment_summary_width_tests.rs:130-147` pins the non-standard self-round-trip rather than an external/conformant wire fixture.

2. **P1 — valid ACK entries that omit `notificationClass` are rejected** (`crates/bacnet-services/src/enrollment_summary.rs:422-440`). Clause 13.8 defines the final `notificationClass Unsigned` member as optional, but the decoder unconditionally reads a fifth tag. The valid one-entry payload `C4 00 00 00 01 91 00 91 00 21 01` (object-id, event type, event state, priority, no notification class) reaches line 423 with `offset == data.len()` and returns a decode error; if another entry follows, its object-id is rejected as the expected Unsigned instead. `raw_ack` always appends this field (`crates/bacnet-services/src/enrollment_summary_width_tests.rs:34-47`), so the tests do not exercise the optional-field boundary or repeated entries with it omitted.

jscott3201 (commented):
[Dataflow] READY

Reviewed exactly `9c60d4bdf8f5a10d86ddfd391a1c31bbf43a5a90...3a1a209e7f99a51520584cbba7696e7fe7da1cca`; the PR still reports that base/head pair.

No material diff defect found. The request decoder consumes the stated nested path—outer enrollment filter `[1]`, recipient member `[0]`, device choice `[0]`, then process member `[1]`—and rejects incomplete or extra members (`crates/bacnet-services/src/enrollment_summary.rs:116-184`). All newly narrowed unsigned decode paths use checked conversions (`:110-113`, `:167-173`, `:192-206`, `:232-261`, `:287-294`, `:392-416`, `:434-454`).

ACK processing validates the application tag/type before each required field and treats an absent optional notification class as public `0` without consuming the next entry's object-identifier (`crates/bacnet-services/src/enrollment_summary.rs:359-465`); the representation is documented at `:319-328`. The Python boundary forwards that decoded value unchanged (`crates/rusty-bacnet/src/client/client_methods/enrollment_alarm_covmulti_who_writegroup.rs:53-76`).

Separate baseline observations: `RecipientProcess.device: None` has long encoded an outer filter containing only process `[1]` (`crates/bacnet-services/src/enrollment_summary.rs:61-69`); that is not a conformant producer under the stated grammar and the new decoder appropriately rejects it. This review does not count that pre-existing encoder behavior as a defect in this diff. The server's notification-class/priority narrowing and filter semantics are unchanged (`crates/bacnet-server/src/handlers/alarm_event/get_enrollment_summary.rs:50-99`).

jscott3201 (commented):
[Adversary] READY

Reviewed the authoritative range `9c60d4bdf8f5a10d86ddfd391a1c31bbf43a5a90...3a1a209e7f99a51520584cbba7696e7fe7da1cca` (base and head still match). No material diff defects found.

Checked strict nested recipient/process and priority framing plus complete request consumption (`enrollment_summary.rs:95-302`); ACK application-tag/LVT validation, optional notification-class handling at end and between entries, width boundaries, item cap, and malformed/truncated paths (`enrollment_summary.rs:12-14,351-468`).

jscott3201 (commented):
[Runtime] NOT READY

- **[P1] Do not emit an enrollment filter without its mandatory recipient — `crates/bacnet-services/src/enrollment_summary.rs:61-68`, `crates/bacnet-services/src/enrollment_summary.rs:122-128`.** `RecipientProcess.device` remains public as an `Option`, but `encode` always opens enrollment-filter `[1]` and emits process-identifier `[1]` even when `device` is `None`; it only emits recipient `[0]` inside the `Some` branch. A concrete `GetEnrollmentSummaryRequest { enrollment_filter: Some(RecipientProcess { device: None, process_identifier: 7 }), .. }` therefore produces a malformed BACnetRecipientProcess. The hardened decoder now rejects that encoder output at the mandatory-recipient check, so a previously constructible public request no longer round-trips and a peer will reject the request. The new framing test covers only `device: Some(...)` and misses this failure mode.

I otherwise verified the 11 checked narrowing conversions, strict request/ACK tags and trailing-data rejection, priority ordering, omitted ACK notification-class handling across repeated entries, the 10,000-entry cap, and index/progress paths for malformed-input panic safety.

jscott3201 (commented):
[Dataflow] NOT READY — reviewed `9c60d4bdf8f5a10d86ddfd391a1c31bbf43a5a90...147df347e15d9aff2132fa0c975f0e5688630a85`.

- **P1 — Invalid Python priority ranges now panic instead of returning a Python error.** A call such as `await client.get_enrollment_summary(addr, min_priority=2, max_priority=1)` builds `PriorityFilter { min: 2, max: 1 }` at `crates/rusty-bacnet/src/client/client_methods/enrollment_alarm_covmulti_who_writegroup.rs:29-35`, then invokes `req.encode()` at `:54`. This change makes that path call `try_encode(...).expect(...)` at `crates/bacnet-services/src/enrollment_summary.rs:62-64`; the newly introduced validation rejects the range at `:78-85`. Consequently the Python async call panics before `confirmed_request` rather than mapping the validation failure to `PyResult` (as response decoding does at `enrollment_alarm_covmulti_who_writegroup.rs:59`). Use `try_encode(...).map_err(to_py_err)?` or validate the range as a `PyValueError` before starting the future.

Verified otherwise: public Rust request/ACK field types and valid encoding flow remain intact; request decoding checks the targeted widths/tags and consumes trailing data; ACK decoding defaults an omitted notification class to zero and the Python output continues to expose that value. The server handler is unchanged.

jscott3201 (commented):
[Adversary] NOT READY

- [P1] `notificationClassFilter` and ACK `notificationClass` are specified as `Unsigned` in Clause 13.8, not `Unsigned16`. The new conversions reject a conformant value such as 65,536: an ACK whose optional final value is `0x23 0x01 0x00 0x00` now fails at `crates/bacnet-services/src/enrollment_summary.rs:481-484`, and the equivalent request filter fails at `crates/bacnet-services/src/enrollment_summary.rs:317-323`. The public `u16` fields at `crates/bacnet-services/src/enrollment_summary.rs:52-53,356-357` also make such values impossible to represent or encode. The added regressions explicitly classify `0x01_00_00` as an overflow (`crates/bacnet-services/src/enrollment_summary_width_tests.rs:84-96,214-224`). This turns valid peers that use notification-class instances above 65,535 into decode/request failures rather than supporting the service's specified `Unsigned` domain.

jscott3201 (commented):
[Dataflow] READY

Reviewed exactly `9c60d4bdf8f5a10d86ddfd391a1c31bbf43a5a90...af31c2c78fe81945c91edb55c297118da3067889`; the PR base/head still match the target.

No material dataflow defect found. Checked conversions reject values above their public `u32`/`u16`/`u8` destinations before request dispatch or Python ACK projection (`crates/bacnet-services/src/enrollment_summary.rs:131-134,188-204,213-228,253-293,302-309,407-451`). Mandatory request framing and trailing data are rejected (`:117-205,239-288,312-317`); valid encodes remain unchanged (`:78-110,352-360`). A missing enrollment device is rejected by fallible encoding before writes (`:68-77`), while the Python request path always constructs `enrollment_filter: None`, so Python parameters cannot reach `encode`'s panic path (`crates/rusty-bacnet/src/client/client_methods/enrollment_alarm_covmulti_who_writegroup.rs:15-58`).

ACK application type/LVT validation is enforced (`enrollment_summary.rs:12-14,375-451`), omitted optional notification class represents as `0` and correctly leaves a following object identifier for the next entry (`:454-480`), and priority min/max ordering remains unvalidated (`:289-292`). Boundary/framing/default regressions cover these cases (`enrollment_summary_width_tests.rs:58-268`).

jscott3201 (commented):
[Adversary] NOT READY

**P1 — Valid device enrollment filters are rejected, while locally encoded ones use a non-conformant layout.** `crates/bacnet-services/src/enrollment_summary.rs:82-89` wraps the device choice in a constructed `[0]` pair and writes `processIdentifier` as context `[1]`; `:143-179` then requires exactly that form. `BACnetRecipient::device [0]` is a primitive context tag containing the four object-ID octets, and the following process identifier is an untagged application Unsigned (`crates/bacnet-encoding/src/constructed/recipient.rs:16-28,65-80`).

For example, a valid filter for Device 7 / process 7 is `09 00 1e 0c 02 00 00 07 21 07 1f`: acknowledgment filter, opening enrollment filter, primitive device `[0]`, application Unsigned process ID, closing enrollment filter. The new decoder rejects it at `enrollment_summary.rs:143-149` because it demands an opening `[0]`, and local requests emitted by `:82-89` will be rejected by conforming peers. Parse/emit the actual `BACnetRecipientProcess` member framing instead of making the pre-existing private layout mandatory.

jscott3201 (commented):
[Adversary] READY

Supersedes my prior review. I incorrectly applied the untagged `BACnetDestination` layout to `BACnetRecipientProcess`. Under the cited RecipientProcess grammar, the framing required by `crates/bacnet-services/src/enrollment_summary.rs:82-89,143-205`—outer enrollment `[1]`, recipient member `[0]`, nested device-choice `[0]`, and context process member `[1]`—is correct.

Re-reviewed `9c60d4bdf8f5a10d86ddfd391a1c31bbf43a5a90...af31c2c78fe81945c91edb55c297118da3067889`: the checked conversions reject out-of-range numerical values while retaining leading-zero values that fit; malformed/duplicate/trailing request members and invalid/reserved ACK field tags fail closed; optional notification class and the ACK entry cap are handled. No remaining material defect found.

jscott3201 (commented):
[Runtime] READY

No material runtime defect found in `9c60d4bdf8f5a10d86ddfd391a1c31bbf43a5a90...af31c2c78fe81945c91edb55c297118da3067889`.

Verified:
- Request encoding validates a missing recipient device before the first buffer write, and the device recipient is encoded with outer constructed recipient `[0]`, inner primitive device `[0]`, followed by process `[1]` (`crates/bacnet-services/src/enrollment_summary.rs:68-89`).
- Request decoding requires the recipient/device/process and priority-filter members, rejects trailing members, and uses checked narrowing for all destination widths (`crates/bacnet-services/src/enrollment_summary.rs:113-317`).
- ACK decoding validates application class/type and reserved L/V/T forms, checks all narrow conversions, distinguishes an omitted notification class from the next object-identifier entry, and enforces the decoded-entry cap (`crates/bacnet-services/src/enrollment_summary.rs:362-483`).
- The regressions exercise fitting leading-zero representations, immediate-overflow and `u64::MAX` values, exact recipient framing, no-write validation, optional notification-class boundaries across repeated entries, wrong tags/LVTs, and trailing input (`crates/bacnet-services/src/enrollment_summary_width_tests.rs:58-269`).

The head SHA still matches the review target. Head CI reports successful Test, Clippy, Rustfmt, file-size, and secret-scan checks.

PR #336, opened 2026-08-15
fix(services): harden alarm operation decoding

jscott3201:
## Summary
- reject AcknowledgeAlarm and LifeSafetyOperation numeric fields above their existing `u32` contracts
- require the Standard context tags for every request member and consume the full payload
- share checked context-field decoding across service modules

## Verification
- file-size and secret checks

Closes part of #309.

### DIFF
--- a/crates/bacnet-services/src/alarm_event/acknowledge_alarm.rs
+++ b/crates/bacnet-services/src/alarm_event/acknowledge_alarm.rs
@@ -1,5 +1,7 @@
 use super::*;
 
+use crate::common::{decode_context, decode_context_u32};
+
 // ---------------------------------------------------------------------------
 // AcknowledgeAlarm
 // ---------------------------------------------------------------------------
@@ -37,39 +39,18 @@ impl AcknowledgeAlarmRequest {
         let mut offset = 0;
 
         // [0]
-        let (_tag, pos) = tags::decode_tag(data, offset)?;
-        let end = pos + _tag.length as usize;
-        if end > data.len() {
-            return Err(Error::decoding(
-                pos,
-                "AcknowledgeAlarm truncated at process-id",
-            ));
-        }
-        let acknowledging_process_identifier = primitives::decode_unsigned(&data[pos..end])? as u32;
+        let (acknowledging_process_identifier, end) =
+            decode_context_u32(data, offset, 0, "AcknowledgeAlarm process-id")?;
         offset = end;
 
         // [1]
-        let (_tag, pos) = tags::decode_tag(data, offset)?;
-        let end = pos + _tag.length as usize;
-        if end > data.len() {
-            return Err(Error::decoding(
-                pos,
-                "AcknowledgeAlarm truncated at object-id",
-            ));
-        }
-        let event_object_identifier = ObjectIdentifier::decode(&data[pos..end])?;
+        let (content, end) = decode_context(data, offset, 1, "AcknowledgeAlarm object-id")?;
+        let event_object_identifier = ObjectIdentifier::decode(content)?;
         offset = end;
 
         // [2]
-        let (_tag, pos) = tags::decode_tag(data, offset)?;
-        let end = pos + _tag.length as usize;
-        if end > data.len() {
-            return Err(Error::decoding(
-                pos,
-                "AcknowledgeAlarm truncated at event-state",
-            ));
-        }
-        let event_state_acknowledged = primitives::decode_unsigned(&data[pos..end])? as u32;
+        let (event_state_acknowledged, end) =
+            decode_context_u32(data, offset, 2, "AcknowledgeAlarm event-state")?;
         offset = end;
 
         // [3] timestamp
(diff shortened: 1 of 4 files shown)


jscott3201 (commented):
[Runtime] READY

No material runtime defects found in `a20b1cd1aea82929f8deef0078317eece0984513...d7188d95fac50da4e5365ef559e336a65475bcfd`.

Evidence:
- `crates/bacnet-services/src/common.rs:12-44` centralizes context-tag ownership, checked end-offset arithmetic, truncation rejection, and fallible `u64`→`u32` narrowing.
- `crates/bacnet-services/src/alarm_event/acknowledge_alarm.rs:42-73` consumes the six required fields under their exact tags and rejects trailing bytes.
- `crates/bacnet-services/src/life_safety.rs:43-69` enforces required tags, preserves unknown in-range `LifeSafetyOperation` values through `from_raw`, accepts optional `[3]` only in its defined position, and rejects trailing data.
- Regression tests exercise leading-zero `u32::MAX`, `2^32` rejection in each narrowed field, wrong tags, trailing fields, and shared-helper callers at `crates/bacnet-services/src/alarm_event/tests/service_round_trip.rs:51-99`, `crates/bacnet-services/src/life_safety.rs:129-176`, and `crates/bacnet-services/src/common.rs:321-373`.

GitHub's Rustfmt, Clippy, and Ubuntu test checks are green for the reviewed head.

jscott3201 (commented):
[Dataflow] READY

Reviewed exact range `a20b1cd1aea82929f8deef0078317eece0984513...d7188d95fac50da4e5365ef559e336a65475bcfd`. No material data-flow defects found.

- The crate-private helper validates the required primitive context tag and bounded content before conversion, and `u32::try_from` rejects values above the public field width (`crates/bacnet-services/src/common.rs:12-44`). Existing shared consumers retain their tag/optional-field state transitions (`crates/bacnet-services/src/common.rs:73-96`, `:143-169`).
- AcknowledgeAlarm consumes required fields [0]-[5]; timestamp decoding verifies its enclosing opening/closing tag (`crates/bacnet-services/src/alarm_event/acknowledge_alarm.rs:41-73`; `crates/bacnet-encoding/src/primitives/mod.rs:748-771`). It rejects trailing bytes before constructing the request, and the server decodes before choosing a transition or mutating the object (`crates/bacnet-server/src/handlers/alarm_event/acknowledge_alarm.rs:11-27`).
- LifeSafetyOperation preserves an arbitrary in-range `u32` via `LifeSafetyOperation::from_raw`, accepts [3] only when present exactly once, and rejects any remaining data (`crates/bacnet-services/src/life_safety.rs:42-69`; `crates/bacnet-types/src/enums/life_safety.rs:72-86`).
- Valid producer paths are unchanged: Rust/Python AcknowledgeAlarm use the same typed request/encoder (`crates/bacnet-client/src/client/file_list.rs:27-53`, `crates/rusty-bacnet/src/client/client_methods/object_device_alarm.rs:161-193`), and Python LifeSafetyOperation preserves the optional object identifier into the request (`crates/rusty-bacnet/src/client/client_methods/file_list_private_text_life.rs:371-399`). Server dispatch maps decoder failures to an Error APDU rather than a SimpleACK (`crates/bacnet-server/src/server/requests/mod.rs:275-280`, `:342-346`).

jscott3201 (commented):
[Adversary] NOT READY

**P2 — Reject non-canonical context-tag headers before accepting the field** (`crates/bacnet-services/src/common.rs:18-19`; `crates/bacnet-encoding/src/tags.rs:182-188,217-245`)

`decode_context` matches only the normalized `Tag`, while `decode_tag` accepts both an extended tag-number encoding whose following byte is `0` and an extended-length encoding whose decoded length is `1`. Clause 20.2.1 assigns the extended tag-number form to tags 15–254 and L/V/T `101` to lengths greater than 4. Thus, take any otherwise-valid AcknowledgeAlarm or LifeSafetyOperation request with a leading canonical `[0] Unsigned(1)` field (`09 01`) and replace it with either `FD 00 01 01` or `0D 01 01`. Both malformed headers normalize to context tag 0, length 1, content 1, so the request succeeds instead of being rejected. For AcknowledgeAlarm this reaches `crates/bacnet-server/src/handlers/alarm_event/acknowledge_alarm.rs:27` and changes the object's acknowledged-transition state.

Make the tag decoder (or the service helper while the raw header is available) reject non-canonical extended tag numbers/lengths, and add these two malformed-header cases for both services.

jscott3201 (commented):
[Adversary] READY

Supersedes my earlier finding: non-canonical tag-header acceptance is pre-existing in `bacnet_encoding::tags::decode_tag` and is not introduced or changed by this service-layer diff.

Reassessed exactly `a20b1cd1aea82929f8deef0078317eece0984513...d7188d95fac50da4e5365ef559e336a65475bcfd`. No material diff defect remains: the shared helper validates normalized context field identity and bounds before slicing; `decode_context_u32` preserves fitting leading-zero encodings while rejecting values above `u32`; both service decoders now consume the complete request; and LifeSafety's optional `[3]` is rejected when unknown, duplicate, or trailing data follows.

PR #338, opened 2026-08-15
fix(encoding): reject noncanonical tag headers

### LINKED ISSUE #337: Tag decoder accepts non-canonical extended headers
### Affected revision

Observed on `dev` at `a20b1cd`; surfaced by the adversarial review of #336.

### Behavior

`decode_tag` normalizes extended tag-number and extended-length headers without checking whether the extended form is legal for the decoded value. For example, a canonical context `[0]` field with one content octet is `09 01`, but both `FD 00 01 01` (extended tag number 0) and `0D 01 01` (extended length 1) decode to the same `Tag`. Service decoders therefore accept malformed BACnet encodings, including state-changing AcknowledgeAlarm requests.

This predates #336 and affects every `decode_tag` caller.

### Standard evidence

ASHRAE 135-2020 Clause 20.2.1.2 requires tag numbers 0..=14 in the initial octet and reserves the extended form for 15..=254. Clause 20.2.1.3 requires lengths 0..=4 in the initial octet, one-octet extension for 5..=253, the `254` marker for 254..=65535, and the `255` marker for larger values.

### Acceptance

- reject extended tag numbers below 15 and the reserved value 255;
- reject extended-length forms whose decoded length belongs to a shorter form;
- cover primitive and opening/closing tags without changing canonical encodings;
- run the full workspace and feature-matrix tests because the decoder is shared by all BACnet layers.

### Relevant code

- `crates/bacnet-encoding/src/tags.rs`

Found during #336; tracked separately from the checked-narrowing work in #309.

jscott3201:
## Summary
- reject extended tag-number headers for values outside the Standard 15..=254 range
- reject extended-length headers when the decoded length belongs to a shorter form
- preserve the shipped EventEnrollment tag-255 legacy sentinel only inside its constructed codec
- update lenient service fixtures to use canonical non-matching tags

## Standard
ASHRAE 135-2020 Clause 20.2.1.2 assigns the extended tag-number form to 15..=254. Clause 20.2.1.3 assigns the one-, two-, and four-octet length extensions to 5..=253, 254..=65535, and 65536..=2^32-1 respectively.

## Verification
- file-size and secret checks

Closes #337.

### DIFF
--- a/crates/bacnet-encoding/src/constructed/event_parameter.rs
+++ b/crates/bacnet-encoding/src/constructed/event_parameter.rs
@@ -185,13 +185,21 @@ pub fn decode_event_parameter(
     // can reject ASHRAE-reserved tag number 255 everywhere else.
     if data.get(offset..offset.saturating_add(2)) == Some(&[0xFE, 0xFF]) {
         let body_start = offset + 2;
-        let (body, end) = tags::extract_raw_context(data, body_start, u8::MAX)?;
+        let body_end = data.len().checked_sub(2).ok_or_else(|| {
+            Error::decoding(offset, "BACnetEventParameter: truncated legacy wrapper")
+        })?;
+        if body_end < body_start || data.get(body_end..) != Some(&[0xFF, 0xFF]) {
+            return Err(Error::decoding(
+                offset,
+                "BACnetEventParameter: missing legacy closing tag",
+            ));
+        }
         return Ok((
             EP::Opaque {
                 tag: u8::MAX,
-                data: body.to_vec(),
+                data: data[body_start..body_end].to_vec(),
             },
-            end,
+            data.len(),
         ));
     }
 
--- a/crates/bacnet-encoding/src/constructed/event_parameter.rs
+++ b/crates/bacnet-encoding/src/constructed/event_parameter.rs
@@ -151,6 +151,10 @@ pub fn encode_event_parameter(buf: &mut BytesMut, value: &BACnetEventParameter) 
             tags::encode_closing_tag(buf, 2);
             tags::encode_closing_tag(buf, 9);
         }
+        BACnetEventParameter::Opaque { tag, data } if *tag == u8::MAX => {
+            // Legacy pre-framing EventEnrollment values used an Octet String.
+            primitives::encode_app_octet_string(buf, data);
+        }
         BACnetEventParameter::Opaque { tag, data } => {
             // Preserved alternative: its captured bytes are the complete
             // SEQUENCE body, re-emitted under its own opening/closing pair.
@@ -193,6 +197,22 @@ pub fn decode_event_parameter(
 
     let (tag, pos) = tags::decode_tag(data, offset)?;
     let what = "BACnetEventParameter";
+
+    if tag.class == tags::TagClass::Application && tag.number == tags::app_tag::OCTET_STRING {
+        let end = pos
+            .checked_add(tag.length as usize)
+            .ok_or_else(|| Error::decoding(pos, "BACnetEventParameter: length overflow"))?;
+        if end > data.len() {
+            return Err(Error::buffer_too_short(end, data.len()));
+        }
+        return Ok((
+            EP::Opaque {
+                tag: u8::MAX,
+                data: data[pos..end].to_vec(),
+            },
+            end,
+        ));
+    }
 
     if RESERVED_EVENT_TAGS.contains(&tag.number) {
         return Err(Error::decoding(
--- a/crates/bacnet-encoding/src/constructed/event_parameter.rs
+++ b/crates/bacnet-encoding/src/constructed/event_parameter.rs
@@ -175,6 +175,22 @@ pub fn decode_event_parameter(
     offset: usize,
 ) -> Result<(BACnetEventParameter, usize), Error> {
     use BACnetEventParameter as EP;
+
+    // Pre-framing EventEnrollment writes use [255] as an internal wrapper.
+    // Keep this shipped compatibility form local so the generic tag parser
+    // can reject ASHRAE-reserved tag number 255 everywhere else.
+    if data.get(offset..offset.saturating_add(2)) == Some(&[0xFE, 0xFF]) {
+        let body_start = offset + 2;
+        let (body, end) = tags::extract_raw_context(data, body_start, u8::MAX)?;
+        return Ok((
+            EP::Opaque {
+                tag: u8::MAX,
+                data: body.to_vec(),
+            },
+            end,
+        ));
+    }
+
     let (tag, pos) = tags::decode_tag(data, offset)?;
     let what = "BACnetEventParameter";
 
--- a/crates/bacnet-encoding/src/constructed/tests/event_parameter.rs
+++ b/crates/bacnet-encoding/src/constructed/tests/event_parameter.rs
@@ -267,6 +267,21 @@ fn opaque_unmodeled_alternatives_preserved() {
     }
 }
 
+#[test]
+fn legacy_opaque_sentinel_stays_local_to_event_parameters() {
+    let value = BACnetEventParameter::Opaque {
+        tag: u8::MAX,
+        data: vec![0xFF, 0x01, 0x02],
+    };
+    let mut encoded = BytesMut::new();
+    encode_event_parameter(&mut encoded, &value);
+
+    assert!(tags::decode_tag(&encoded, 0).is_err());
+    let (decoded, end) = decode_event_parameter(&encoded, 0).unwrap();
+    assert_eq!(decoded, value);
+    assert_eq!(end, encoded.len());
+}
+
 // --- Negatives ----------------------------------------------------------------
 
 #[test]
--- a/crates/bacnet-encoding/src/constructed/tests/event_parameter.rs
+++ b/crates/bacnet-encoding/src/constructed/tests/event_parameter.rs
@@ -283,14 +283,14 @@ fn legacy_opaque_sentinel_stays_local_to_event_parameters() {
     assert_eq!(decoded, value);
     assert_eq!(end, encoded.len());
 
-    let historical = [0xFE, 0xFF, 1, 2, 3, 0xFF, 0xFF];
+    let historical = [0xFE, 0xFF, 1, 0xFF, 0xFF, 2, 0xFF, 0xFF];
     assert!(tags::decode_tag(&historical, 0).is_err());
     let (decoded, end) = decode_event_parameter(&historical, 0).unwrap();
     assert_eq!(
         decoded,
         BACnetEventParameter::Opaque {
             tag: u8::MAX,
-            data: vec![1, 2, 3],
+            data: vec![1, 0xFF, 0xFF, 2],
         }
     );
     assert_eq!(end, historical.len());
--- a/crates/bacnet-encoding/src/constructed/tests/event_parameter.rs
+++ b/crates/bacnet-encoding/src/constructed/tests/event_parameter.rs
@@ -276,10 +276,24 @@ fn legacy_opaque_sentinel_stays_local_to_event_parameters() {
     let mut encoded = BytesMut::new();
     encode_event_parameter(&mut encoded, &value);
 
-    assert!(tags::decode_tag(&encoded, 0).is_err());
+    let (tag, _) = tags::decode_tag(&encoded, 0).unwrap();
+    assert_eq!(tag.class, tags::TagClass::Application);
+    assert_eq!(tag.number, tags::app_tag::OCTET_STRING);
     let (decoded, end) = decode_event_parameter(&encoded, 0).unwrap();
     assert_eq!(decoded, value);
     assert_eq!(end, encoded.len());
+
+    let historical = [0xFE, 0xFF, 1, 2, 3, 0xFF, 0xFF];
+    assert!(tags::decode_tag(&historical, 0).is_err());
+    let (decoded, end) = decode_event_parameter(&historical, 0).unwrap();
+    assert_eq!(
+        decoded,
+        BACnetEventParameter::Opaque {
+            tag: u8::MAX,
+            data: vec![1, 2, 3],
+        }
+    );
+    assert_eq!(end, historical.len());
 }
 
 // --- Negatives ----------------------------------------------------------------
--- a/crates/bacnet-encoding/src/constructed/tests/event_parameter.rs
+++ b/crates/bacnet-encoding/src/constructed/tests/event_parameter.rs
@@ -283,14 +283,14 @@ fn legacy_opaque_sentinel_stays_local_to_event_parameters() {
     assert_eq!(decoded, value);
     assert_eq!(end, encoded.len());
 
-    let historical = [0xFE, 0xFF, 1, 0xFF, 0xFF, 2, 0xFF, 0xFF];
+    let historical = [0xFE, 0xFF, 1, 0xFF, 0xFF, 0x2F, 2, 0xFF, 0xFF];
     assert!(tags::decode_tag(&historical, 0).is_err());
     let (decoded, end) = decode_event_parameter(&historical, 0).unwrap();
     assert_eq!(
         decoded,
         BACnetEventParameter::Opaque {
             tag: u8::MAX,
-            data: vec![1, 0xFF, 0xFF, 2],
+            data: vec![1, 0xFF, 0xFF, 0x2F, 2],
         }
     );
     assert_eq!(end, historical.len());
--- a/crates/bacnet-encoding/src/lib.rs
+++ b/crates/bacnet-encoding/src/lib.rs
@@ -12,3 +12,6 @@ pub mod npdu;
 pub mod primitives;
 pub mod segmentation;
 pub mod tags;
+
+#[cfg(test)]
+mod tags_canonical_tests;
--- a/crates/bacnet-encoding/src/tags.rs
+++ b/crates/bacnet-encoding/src/tags.rs
@@ -178,6 +178,9 @@ pub fn decode_tag(data: &[u8], offset: usize) -> Result<(Tag, usize), Error> {
         TagClass::Application
     };
     let lvt = initial & 0x07;
+    if class == TagClass::Application && lvt > 5 {
+        return Err(Error::decoding(offset, "reserved application tag L/V/T"));
+    }
 
     if tag_number == 0x0F {
         if pos >= data.len() {
@@ -587,15 +590,6 @@ mod tests {
     }
 
     #[test]
-    fn decode_rejects_invalid_extended_tag_numbers() {
-        for initial in [0xF1, 0xF9, 0xFE, 0xFF] {
-            for tag_number in [0, 1, 14, 255] {
-                assert!(decode_tag(&[initial, tag_number], 0).is_err());
-            }
-        }
-    }
-
-    #[test]
     fn decode_extended_length() {
         // Tag 2, Application, length 100: 0x25, 100
         let (tag, pos) = decode_tag(&[0x25, 100], 0).unwrap();
@@ -614,35 +608,6 @@ mod tests {
         assert_eq!(tag.number, 2);
         assert_eq!(tag.length, 100000);
         assert_eq!(pos, 6);
-    }
-
-    #[test]
-    fn decode_rejects_noncanonical_extended_lengths() {
-        for length in 0..=4 {
-            assert!(decode_tag(&[0x0D, length], 0).is_err());
-        }
-
-        assert!(decode_tag(&[0x0D, 254, 0x00, 0xFD], 0).is_err());
-        assert!(decode_tag(&[0x0D, 255, 0x00, 0x00, 0xFF, 0xFF], 0).is_err());
-    }
-
-    #[test]
-    fn decode_accepts_canonical_extended_length_boundaries() {
-        let cases: &[(&[u8], u32)] = &[
-            (&[0x0D, 5], 5),
-            (&[0x0D, 253], 253),
-            (&[0x0D, 254, 0x00, 0xFE], 254),
-            (&[0x0D, 254, 0xFF, 0xFF], 65_535),
-            (&[0x0D, 255, 0x00, 0x01, 0x00, 0x00], 65_536),
-        ];
-
-        for &(encoded, expected_length) in cases {
-            let (tag, pos) = decode_tag(encoded, 0).unwrap();
-            assert_eq!(tag.number, 0);
-            assert_eq!(tag.class, TagClass::Context);
-            assert_eq!(tag.length, expected_length);
-            assert_eq!(pos, encoded.len());
-        }
     }
 
     #[test]
--- a/crates/bacnet-encoding/src/tags.rs
+++ b/crates/bacnet-encoding/src/tags.rs
@@ -184,6 +184,12 @@ pub fn decode_tag(data: &[u8], offset: usize) -> Result<(Tag, usize), Error> {
             return Err(Error::decoding(pos, "truncated extended tag number"));
         }
         tag_number = data[pos];
+        if !(15..=254).contains(&tag_number) {
+            return Err(Error::decoding(
+                pos,
+                format!("invalid extended tag number {tag_number}"),
+            ));
+        }
         pos += 1;
     }
 
@@ -224,12 +230,24 @@ pub fn decode_tag(data: &[u8], offset: usize) -> Result<(Tag, usize), Error> {
         pos += 1;
 
         match ext {
-            0..=253 => ext as u32,
+            0..=4 => {
+                return Err(Error::decoding(
+                    pos - 1,
+                    format!("non-canonical extended tag length {ext}"),
+                ));
+            }
+            5..=253 => ext as u32,
             254 => {
                 if pos + 2 > data.len() {
                     return Err(Error::decoding(pos, "truncated 2-byte extended length"));
                 }
                 let len = u16::from_be_bytes([data[pos], data[pos + 1]]) as u32;
+                if len < 254 {
+                    return Err(Error::decoding(
+                        pos,
+                        format!("non-canonical 2-byte extended tag length {len}"),
+                    ));
+                }
                 pos += 2;
                 len
             }
@@ -239,6 +257,12 @@ pub fn decode_tag(data: &[u8], offset: usize) -> Result<(Tag, usize), Error> {
                 }
                 let len =
                     u32::from_be_bytes([data[pos], data[pos + 1], data[pos + 2], data[pos + 3]]);
+                if len < 65_536 {
+                    return Err(Error::decoding(
+                        pos,
+                        format!("non-canonical 4-byte extended tag length {len}"),
+                    ));
+                }
                 pos += 4;
                 len
             }
@@ -563,6 +587,15 @@ mod tests {
     }
 
     #[test]
+    fn decode_rejects_invalid_extended_tag_numbers() {
+        for initial in [0xF1, 0xF9, 0xFE, 0xFF] {
+            for tag_number in [0, 1, 14, 255] {
+                assert!(decode_tag(&[initial, tag_number], 0).is_err());
+            }
+        }
+    }
+
+    #[test]
     fn decode_extended_length() {
         // Tag 2, Application, length 100: 0x25, 100
         let (tag, pos) = decode_tag(&[0x25, 100], 0).unwrap();
@@ -581,6 +614,35 @@ mod tests {
         assert_eq!(tag.number, 2);
         assert_eq!(tag.length, 100000);
         assert_eq!(pos, 6);
+    }
+
+    #[test]
+    fn decode_rejects_noncanonical_extended_lengths() {
+        for length in 0..=4 {
+            assert!(decode_tag(&[0x0D, length], 0).is_err());
+        }
+
+        assert!(decode_tag(&[0x0D, 254, 0x00, 0xFD], 0).is_err());
+        assert!(decode_tag(&[0x0D, 255, 0x00, 0x00, 0xFF, 0xFF], 0).is_err());
+    }
+
+    #[test]
+    fn decode_accepts_canonical_extended_length_boundaries() {
+        let cases: &[(&[u8], u32)] = &[
+            (&[0x0D, 5], 5),
+            (&[0x0D, 253], 253),
+            (&[0x0D, 254, 0x00, 0xFE], 254),
+            (&[0x0D, 254, 0xFF, 0xFF], 65_535),
+            (&[0x0D, 255, 0x00, 0x01, 0x00, 0x00], 65_536),
+        ];
+
+        for &(encoded, expected_length) in cases {
+            let (tag, pos) = decode_tag(encoded, 0).unwrap();
+            assert_eq!(tag.number, 0);
+            assert_eq!(tag.class, TagClass::Context);
+            assert_eq!(tag.length, expected_length);
+            assert_eq!(pos, encoded.len());
+        }
     }
 
     #[test]
--- a/crates/bacnet-encoding/src/tags_canonical_tests.rs
+++ b/crates/bacnet-encoding/src/tags_canonical_tests.rs
@@ -1,0 +1,48 @@
+use crate::tags::{self, TagClass};
+
+#[test]
+fn rejects_invalid_extended_tag_numbers() {
+    for initial in [0xF1, 0xF9, 0xFE, 0xFF] {
+        for tag_number in [0, 1, 14, 255] {
+            assert!(tags::decode_tag(&[initial, tag_number], 0).is_err());
+        }
+    }
+}
+
+#[test]
+fn rejects_noncanonical_extended_lengths() {
+    for length in 0..=4 {
+        assert!(tags::decode_tag(&[0x0D, length], 0).is_err());
+    }
+
+    assert!(tags::decode_tag(&[0x0D, 254, 0x00, 0xFD], 0).is_err());
+    assert!(tags::decode_tag(&[0x0D, 255, 0x00, 0x00, 0xFF, 0xFF], 0).is_err());
+}
+
+#[test]
+fn accepts_canonical_extended_length_boundaries() {
+    let cases: &[(&[u8], u32)] = &[
+        (&[0x0D, 5], 5),
+        (&[0x0D, 253], 253),
+        (&[0x0D, 254, 0x00, 0xFE], 254),
+        (&[0x0D, 254, 0xFF, 0xFF], 65_535),
+        (&[0x0D, 255, 0x00, 0x01, 0x00, 0x00], 65_536),
+    ];
+
+    for &(encoded, expected_length) in cases {
+        let (tag, pos) = tags::decode_tag(encoded, 0).unwrap();
+        assert_eq!(tag.number, 0);
+        assert_eq!(tag.class, TagClass::Context);
+        assert_eq!(tag.length, expected_length);
+        assert_eq!(pos, encoded.len());
+    }
+}
+
+#[test]
+fn rejects_reserved_application_lvt_forms() {
+    for lvt in [6, 7] {
+        let encoded = [0x20 | lvt, 5, 0, 0, 0, 0, 1];
+        assert!(tags::decode_tag(&encoded, 0).is_err());
+        assert!(crate::primitives::decode_application_value(&encoded, 0).is_err());
+    }
+}
--- a/crates/bacnet-server/src/event_enrollment/mod.rs
+++ b/crates/bacnet-server/src/event_enrollment/mod.rs
@@ -401,9 +401,9 @@ pub fn evaluate_event_enrollments(
             // full ASN.1 CHOICE framing).
             Ok(PropertyValue::ApplicationData(bytes)) => {
                 match bacnet_encoding::constructed::decode_event_parameter(&bytes, 0) {
-                    Ok((ep, _)) => ep,
+                    Ok((ep, consumed)) if consumed == bytes.len() => ep,
                     // Malformed framed value: nothing to evaluate.
-                    Err(_) => continue,
+                    _ => continue,
                 }
             }
             // Legacy flat application-tagged form (downstream/custom object
--- a/crates/bacnet-server/src/handlers/tests/framed_properties.rs
+++ b/crates/bacnet-server/src/handlers/tests/framed_properties.rs
@@ -134,7 +134,7 @@ fn legacy_event_parameters_wire_write_is_canonicalized() {
     let oid = ee.object_identifier();
     db.add(Box::new(ee)).unwrap();
 
-    let payload = vec![1, 0xff, 0xff, 2];
+    let payload = vec![1, 0xff, 0xff, 0x3f, 2];
     let mut historical = vec![0xfe, 0xff];
     historical.extend_from_slice(&payload);
     historical.extend_from_slice(&[0xff, 0xff]);
--- a/crates/bacnet-server/src/handlers/tests/framed_properties.rs
+++ b/crates/bacnet-server/src/handlers/tests/framed_properties.rs
@@ -128,6 +128,33 @@ fn event_parameters_framed_wire_round_trip() {
 }
 
 #[test]
+fn legacy_event_parameters_wire_write_is_canonicalized() {
+    let mut db = ObjectDatabase::new();
+    let ee = EventEnrollmentObject::new(1, "EE-1", 0).unwrap();
+    let oid = ee.object_identifier();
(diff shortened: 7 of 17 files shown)


jscott3201 (commented):
[Dataflow] NOT READY

**P1 — the tag-255 EventEnrollment compatibility value cannot traverse ReadProperty/WriteProperty framing.** `tags.rs:187-191` now rejects the `[0xFE, 0xFF]` opening header before the special case in `constructed/event_parameter.rs:182-190` can run. That special case is therefore only effective when `decode_event_parameter` is invoked directly. A newly created EventEnrollment stores `Opaque { tag: 0xFF, data: [] }` (`bacnet-objects/src/event_enrollment/mod.rs:78-81`) and reads it out as the same framed bytes (`:258-264`). The server places those bytes inside the ReadProperty `[3]` wrapper (`bacnet-server/src/handlers/read_property.rs:35-47`), but the client/service decoder walks that wrapper with `extract_context_value` (`bacnet-services/src/read_property.rs:212`), which calls strict `decode_tag` and fails on its first inner byte. Thus even a normal ReadProperty of an unconfigured EventEnrollment's Event_Parameters now errors instead of returning the legacy opaque value. The analogous WriteProperty wrapper is also parsed by `extract_context_value` at `bacnet-services/src/write_property.rs:114`, so a framed legacy value cannot reach the object decoder either. Preserve this one compatibility representation through the enclosing abstract-syntax framing (and the subsequent generic application-value consumer) or encode it in a form that those boundaries can safely skip; otherwise the promised local raw bypass is unreachable on transport-facing paths.

jscott3201 (commented):
[Runtime] NOT READY

1. **High — the tag-255 compatibility escape is not reached on the actual ReadProperty/WriteProperty wire paths.** `decode_event_parameter` recognizes `[0xFE, 0xFF]` locally at `crates/bacnet-encoding/src/constructed/event_parameter.rs:179-194`, but an EventEnrollment starts with that sentinel (`crates/bacnet-objects/src/event_enrollment/mod.rs:78-81`) and its read arm emits it as `ApplicationData` (`crates/bacnet-objects/src/event_enrollment/mod.rs:260-264`). Before a caller can invoke the specialized decoder, both `ReadPropertyACK::decode` and `WritePropertyRequest::decode` walk the enclosing `[3]` value with `extract_context_value` (`crates/bacnet-services/src/read_property.rs:194,212`; `crates/bacnet-services/src/write_property.rs:107-115`). That walker calls the now-strict generic `decode_tag`, which rejects the inner extended tag number 255 at `crates/bacnet-encoding/src/tags.rs:182-193`. Concrete repro: encode/decode a ReadPropertyACK whose property value is the default empty sentinel `[FE FF FF FF]`; decoding now fails with `invalid extended tag number 255`. The same bytes cannot survive a WriteProperty request decode. Thus the direct codec test at `crates/bacnet-encoding/src/constructed/tests/event_parameter.rs:270-283` passes while the shipped compatibility value—and even a new object's default Event_Parameters—breaks under real service framing. This needs wire-level default/legacy sentinel coverage.

2. **Medium — application L/V/T 6 and 7 are still accepted as alternate extended-length markers.** Opening/closing handling is restricted to context tags at `crates/bacnet-encoding/src/tags.rs:196-221`, but the following branch treats every remaining `lvt >= 5` as an extended length and accepts an extension byte in `5..=253` (`crates/bacnet-encoding/src/tags.rs:223-239`). For example, `decode_tag(&[0x26, 0x05, 0, 0, 0, 0, 1], 0)` returns an application Unsigned tag of length 5, and `decode_application_value` accepts it as `Unsigned(1)`, even though Clause 20.2.1.3 requires L/V/T 5 for that extension and reserves 6/7 for context opening/closing. `[0x27, ...]` behaves likewise. The modified I-Am test only rejects the `ext=4` instance incidentally; it does not close this shared-parser canonicality gap.

jscott3201 (commented):
[Adversary] NOT READY

### High — reserved application L/V/T forms are still accepted

`decode_tag` recognizes L/V/T values 6 and 7 as opening/closing markers only when the class is Context (`crates/bacnet-encoding/src/tags.rs:196-221`); an Application tag with either value instead falls into the extended-length branch (`crates/bacnet-encoding/src/tags.rs:223-239`). For example, the malformed application Unsigned `[0x26, 0x05, 0, 0, 0, 0, 1]` is decoded as application tag 2 with length 5, and `decode_application_value` returns `PropertyValue::Unsigned(1)` (`crates/bacnet-encoding/src/primitives/mod.rs:464,491-510`). The analogous L/V/T=7 form is accepted too.

This is a tag-header validity gap in the change: the code already identifies application L/V/T 6 and 7 as reserved (`crates/bacnet-services/src/who_is.rs:135-139`). It is reachable from an inbound WriteProperty `propertyValue`: the server repeatedly passes each value to `decode_application_value` and accepts a successful result (`crates/bacnet-server/src/handlers/write_property.rs:455-475,517-531`). Thus a request containing that byte sequence can, for example, set EventEnrollment `NOTIFICATION_CLASS` to 1 because its write arm accepts an Unsigned (`crates/bacnet-objects/src/event_enrollment/mod.rs:350-354`) despite the malformed application header.

Reject application-class L/V/T 6 and 7 in the generic tag decoder, and cover both forms through `decode_application_value`/the WriteProperty path.

jscott3201 (commented):
[Dataflow] NOT READY

## P1 — historical sentinel cannot cross ReadProperty/WriteProperty framing

`decode_event_parameter` adds a local exception for the historical `[255]` sentinel (`0xFE 0xFF … 0xFF 0xFF`) at `crates/bacnet-encoding/src/constructed/event_parameter.rs:183-195`.  However, a service decoder consumes the enclosing `[3] propertyValue` first via `extract_context_value` (`crates/bacnet-services/src/write_property.rs:114-116`; similarly `crates/bacnet-services/src/read_property.rs:194,212`). That walker invokes `decode_tag`, whose new `15..=254` check rejects the sentinel's extended tag number `255` (`crates/bacnet-encoding/src/tags.rs:185-196`), before control can reach the local compatibility decoder.

Concrete regression: an historical EventEnrollment `EVENT_PARAMETERS` property value containing the formerly emitted empty opaque encoding `FE FF FF FF`, enclosed in WriteProperty `[3]`, now causes `WritePropertyRequest::decode` to fail. The equivalent historical bytes in a ReadPropertyACK also fail to decode. Thus old stored/captured sentinel traffic cannot be migrated over the transport boundary despite the direct-codec compatibility branch.

The new framing test only exercises the replacement application Octet String form (`crates/bacnet-services/src/read_property.rs:320-358`); its direct historical decode assertion is in `crates/bacnet-encoding/src/constructed/tests/event_parameter.rs:286-297`, outside ReadProperty/WriteProperty framing. It does not cover the failing sequence above.

## Verified flows

* Canonical tag-length boundaries are rejected/accepted at `crates/bacnet-encoding/src/tags.rs:181-270`, with boundary coverage in `crates/bacnet-encoding/src/tags_canonical_tests.rs:13-39`.
* The new canonical sentinel form is an application Octet String (`crates/bacnet-encoding/src/constructed/event_parameter.rs:154-157,201-214`), so a payload such as `FF 01 02` is length-skipped safely by the outer property-value walker.
* The EventEnrollment read arm emits the constructed codec output (`crates/bacnet-objects/src/event_enrollment/mod.rs:258-265`), and the write arm maps application Octet String input back to the sentinel representation (`crates/bacnet-objects/src/event_enrollment/mod.rs:389-407`).

## Residual gaps

No server-handler end-to-end test covers `EVENT_PARAMETERS` legacy opaque input through WriteProperty, object storage, and ReadProperty; the existing handler round trip covers only OutOfRange (`crates/bacnet-server/src/handlers/tests/framed_properties.rs:97-120`).

jscott3201 (commented):
[Runtime] NOT READY

## High

- **The legacy tag-255 compatibility path is unreachable through ReadProperty/WriteProperty service framing.** `decode_event_parameter` locally recognizes historical `FE FF ... FF FF` at `crates/bacnet-encoding/src/constructed/event_parameter.rs:183-195`, but both service decoders first walk the bytes inside context `[3]` with the generic parser (`crates/bacnet-services/src/read_property.rs:194-195,212-213`; `crates/bacnet-services/src/write_property.rs:106-115`). The new range check rejects the inner `FE FF` at `crates/bacnet-encoding/src/tags.rs:185-196`, before `decode_event_parameter` can see it. Concrete reproducer for an empty legacy EventEnrollment sentinel (object type 9/instance 1, property 83) is the service payload `0C 02 40 00 01 19 53 3E FE FF FF FF 3F`; both `ReadPropertyACK::decode` and `WritePropertyRequest::decode` return an invalid-extended-tag error. This is also the common default legacy value. The added service test does not cover that wire form: it calls the new encoder at `crates/bacnet-services/src/read_property.rs:325-330`, which converts the sentinel to an application Octet String, while the only historical-form test invokes `decode_event_parameter` directly at `crates/bacnet-encoding/src/constructed/tests/event_parameter.rs:286-296`. Observable consequence: a new peer cannot decode a previously emitted legacy ReadProperty value (or a WriteProperty that replays one), despite the claimed compatibility.

## Verification

- Verified PR base/head through GitHub: `84e2457defeb0afcc2f93332bd85745dd06fa4f6...a467447fea4dfff85272bde5b6a65734b6c2b883`.
- Inspected the complete PR diff and exact-head surrounding implementations for tag parsing, raw/context extraction, EventParameter framing, primitives, ReadProperty, WriteProperty, and EventEnrollment object/server consumers.
- No local cargo command was run because this reviewer environment exposes no command-execution tool. Head CI reports successful Ubuntu tests, clippy, rustfmt, file-size, and secret checks; the broader test matrix is skipped.

## Residual risks

- Historical raw opaque payloads containing their own `FF FF` delimiter remain ambiguous under `extract_raw_context` (`crates/bacnet-encoding/src/tags.rs:378-405`); this is documented and predates this patch, but it limits how complete legacy recovery can be.

jscott3201 (commented):
[Adversary] NOT READY

### High — the historical-wrapper path is lossy for valid legacy payload bytes
`decode_event_parameter` recognizes the reserved wrapper and delegates its unstructured body to `extract_raw_context` (`crates/bacnet-encoding/src/constructed/event_parameter.rs:183-195`). That scanner treats every raw `0xFF, 0xFF` pair as a structural closing tag (`crates/bacnet-encoding/src/tags.rs:395-405`), even though the compatibility payload is documented as verbatim/raw. Thus a historical value emitted by the old encoder with payload bytes containing that pair is silently truncated rather than preserved; the documented limitation confirms this ambiguity (`tags.rs:378-380`). The new coverage only exercises payload `[1,2,3]` (`constructed/tests/event_parameter.rs:286-296`).

Reproducer: decode `[0xFE,0xFF, 0x00,0x00, 0xFF,0xFF, 0x01,0x02, 0xFF,0xFF]` at offset 0. Actual result is `Ok((Opaque { tag: 255, data: vec![0,0] }, 6))`; the old wrapper's payload is `[0,0,0xFF,0xFF,1,2]` and its complete encoding ends at 10. A caller that enforces full consumption (the EventEnrollment write arm does at `bacnet-objects/src/event_enrollment/mod.rs:402-406`) rejects the otherwise supported legacy value; a caller that does not (the evaluator discards the returned end at `bacnet-server/src/event_enrollment/mod.rs:402-407`) can accept only the prefix and ignore the tail. This contradicts the claimed preservation of shipped legacy opaque framing and provides a delimiter-smuggling form for malformed trailing bytes.

### Attacks checked
- Generic invalid extended numbers 0..=14 and 255, for primitive/opening/closing forms: rejected before callers see a tag (`tags.rs:185-197`).
- Noncanonical extended lengths 0..=4, two-byte <254, and four-byte <65536: rejected (`tags.rs:226-272`); canonical boundaries are covered in `tags_canonical_tests.rs:23-38`.
- Application L/V/T 6/7: rejected before extension parsing (`tags.rs:180-183`).
- Truncated extended-number/length headers: error paths at `tags.rs:185-188`, `229-231`, `244-246`, and `258-260`.
- Exact historical `FE FF ... FF FF` is confined to `decode_event_parameter`; generic `decode_tag` still rejects it. The application-Octet-String normalization also first passes through generic canonical header validation (`event_parameter.rs:198-215`).

### Residual risk
Apart from the finding, the legacy/raw path remains linear-time and copies the complete payload; it has no raw-payload size or nesting bound of its own (`tags.rs:381-410`). Upstream APDU framing must remain the resource limit.

jscott3201 (commented):
[Dataflow] NOT READY

## P1 — legacy WPM extraction can consume subsequent properties and corrupt the migrated value

The new legacy branch scans *backward through the entire remaining request* for the last `FF FF` + enclosing-property-close pattern (`crates/bacnet-services/src/common.rs:47-63`). `BACnetPropertyValue::decode` passes the complete WPM APDU rather than a boundary for this property (`crates/bacnet-services/src/common.rs:187-197`), and WPM then trusts its returned cursor to start the next element (`crates/bacnet-services/src/wpm.rs:73-89`).

Concrete input: make a single WPM access-spec contain two historical `EVENT_PARAMETERS` `BACnetPropertyValue`s. Each has `[2] FE FF <payload> FF FF 2F`. When decoding the first, the reverse loop chooses the second value's `FF FF 2F`, returns a slice spanning both values, and advances the cursor past the second closing tag. WPM consequently produces one property rather than two. The handler recognizes the widened first slice as a legacy wrapper and accepts it, because `decode_event_parameter` takes every byte from `FE FF` through the final `FF FF` as opaque payload (`crates/bacnet-encoding/src/constructed/event_parameter.rs:186-203`), then stores that corrupted concatenation as an OctetString (`crates/bacnet-server/src/handlers/write_property.rs:459-469`). The later write is silently skipped.

The same occurs if a following property's valid abstract-syntax content ends `FF FF` immediately before its `[2]` close. This violates the WPM element-cursor invariant and turns valid historical traffic into data corruption instead of either two writes or a clean rejection.

The added coverage only decodes one standalone `BACnetPropertyValue` (`crates/bacnet-services/src/common.rs:318-345`) and has no multi-property WPM case.

## Verified flows

* The prior simple ReadProperty/WriteProperty framing gap is addressed: both use the property-aware extractor (`crates/bacnet-services/src/read_property.rs:196,214`; `crates/bacnet-services/src/write_property.rs:113-116`), and the historical framing test covers an embedded `FF FF` plus WriteProperty priority (`crates/bacnet-services/src/read_property.rs:365-390`).
* Historical wrapper decoding now treats only the terminal two bytes as the wrapper close, preserving embedded `FF FF` (`crates/bacnet-encoding/src/constructed/event_parameter.rs:186-203`); canonical readback is an application Octet String (`crates/bacnet-encoding/src/constructed/event_parameter.rs:154-157,209-222`).
* The server WriteProperty route accepts a fully consumed historical wrapper and canonicalizes it; the end-to-end test proves that single-write path (`crates/bacnet-server/src/handlers/write_property.rs:459-469`; `crates/bacnet-server/src/handlers/tests/framed_properties.rs:131-155`).
* Evaluator parsing now requires full EventParameter consumption (`crates/bacnet-server/src/event_enrollment/mod.rs:399-407`).

## Residual gaps

No WPM decode/handler test covers a historical EventParameters value followed by another property, particularly another historical wrapper or a value ending `FF FF`.

jscott3201 (commented):
[Runtime] NOT READY

## High

- **The new `BACnetPropertyValue` legacy extractor can consume later properties in a valid multi-value service.** `extract_property_value` searches backward through the entire remaining service buffer for the last `FF FF <outer-close>` sequence (`crates/bacnet-services/src/common.rs:47-63`), while WPM passes that unbounded buffer for each list member (`crates/bacnet-services/src/wpm.rs:73-89`). Concrete WPM payload: `0C 02 40 00 01 1E 09 53 2E FE FF FF FF 2F 09 11 2E 22 FF FF 2F 1F` (EventEnrollment instance 1; first write is the empty historical EVENT_PARAMETERS wrapper; second writes NOTIFICATION_CLASS=65535). For the first value, the reverse scan selects the second value's `FF FF 2F`, so the decoder returns one corrupted EVENT_PARAMETERS property and silently consumes the NOTIFICATION_CLASS property. The server's exact-end compatibility branch then accepts the combined bytes as opaque data (`crates/bacnet-server/src/handlers/write_property.rs:459-469`). The focused `BACnetPropertyValue` tests use only one value in the buffer (`crates/bacnet-services/src/common.rs:315-345`), so they do not exercise this boundary failure.

- **Historical EVENT_PARAMETERS still cannot cross ReadPropertyMultiple.** RPM knows the property identifier before extracting property value `[4]`, but both branches still use strict `tags::extract_context_value` (`crates/bacnet-services/src/rpm.rs:213-266`) rather than the EVENT_PARAMETERS-only helper. An old peer's valid legacy result `0C 02 40 00 01 1E 29 53 4E FE FF FF FF 4F 1F` (the common empty sentinel) reaches the inner `FE FF`, which the canonical parser rejects as extended tag 255 at `crates/bacnet-encoding/src/tags.rs:185-196`. Thus ReadProperty and BACnetPropertyValue were repaired, but a new client still cannot read the same historical property through RPM.

## Verification

- Verified the PR still targets `84e2457defeb0afcc2f93332bd85745dd06fa4f6...d4d53561339d3b87a8ccff59730113c7afc4b0dc`.
- Inspected the complete final diff and exact-head surrounding paths for canonical tag parsing, EventParameter exact-end decoding, ReadProperty, WriteProperty, BACnetPropertyValue, WPM, RPM, server write canonicalization, and focused tests.
- No local cargo command was run because this reviewer environment exposes no command-execution tool. Head CI reports successful Ubuntu tests, clippy, rustfmt, file-size, and secret checks; broader matrix jobs are skipped.

## Residual risks

- The generic raw-context scanner's documented delimiter ambiguity remains for non-sentinel opaque constructed payloads (`crates/bacnet-encoding/src/tags.rs:378-405`).

jscott3201 (commented):
[Adversary] NOT READY

### High — terminal-wrapper extraction can consume later WPM property/spec entries
The new property-scoped exception reverse-scans the entire slice it was passed and chooses the last `FF FF <outer-close>` sequence, rather than a delimiter belonging to the current `BACnetPropertyValue` (`crates/bacnet-services/src/common.rs:53-63`). `BACnetPropertyValue::decode` passes its unbounded `data` slice to that function (`common.rs:196`), and the WPM decoder likewise passes the complete service payload while iterating properties (`crates/bacnet-services/src/wpm.rs:81-88`). Therefore a first historical EventParameters field can absorb a later property's terminal wrapper and its surrounding frames without an error.

Reproducer: encode a `WritePropertyMultipleRequest` containing two `WriteAccessSpecification`s for EventEnrollment objects 1 and 2, each with one `BACnetPropertyValue { property_identifier: EVENT_PARAMETERS, value: vec![0xFE, 0xFF, payload_byte, 0xFF, 0xFF], priority: None }`. Decode the encoded request. The first property's reverse scan selects the second property's final `FF FF 2F`; it returns after the second property's close, the WPM loop consumes the second spec's `1F` as the first spec's close, and decoding succeeds with one spec/property instead of two. The first opaque value contains the swallowed framing and the second intended write is silently dropped. This violates structural framing and WPM's all-or-nothing expectation.

The exact-terminal EventParameter decoder and full-consumption evaluator checks address the prior embedded-delimiter truncation, but they do not bound this earlier extraction step. The new tests cover one isolated property wrapper (`common.rs:315-345`) and Read/WriteProperty, not an EventParameters entry followed by another property/spec.

### Attacks checked
- Embedded `FF FF` in an isolated historical wrapper: preserved by terminal decoding (`constructed/event_parameter.rs:186-203`).
- Truncated wrapper, incorrect terminal close, and trailing bytes handed directly to the EventParameter decoder: rejected by the terminal check.
- Generic invalid extended tag numbers, noncanonical lengths, reserved application L/V/T forms, and truncated headers: rejected in `bacnet-encoding/src/tags.rs:180-272`.
- Evaluator now requires full consumption (`bacnet-server/src/event_enrollment/mod.rs:402-407`).

### Residual risk
Legacy raw framing is inherently delimiter-ambiguous without a length or escaping scheme; compatibility handling must stay bounded by a trusted enclosing field rather than scan across sibling service elements.

jscott3201 (commented):
[Dataflow] NOT READY

## P1 — forward delimiter scan rejects legacy payloads containing the enclosing close octet

The historical wrapper holds arbitrary raw bytes, but `extract_property_value` now takes the *first* `FF FF <outer-close>` sequence (`crates/bacnet-services/src/common.rs:53-63`). That sequence is not unambiguous in an unescaped raw payload.

Concrete input for a WritePropertyMultiple `EVENT_PARAMETERS` value: `[2] FE FF 01 FF FF 2F 02 FF FF 2F`, where `FE FF`/terminal `FF FF` are the historical tag-255 wrapper and the final `2F` closes property-value `[2]`. The raw legacy payload is `01 FF FF 2F 02`. The loop selects the embedded `FF FF 2F`, returns only `FE FF 01 FF FF`, and advances the cursor to the raw `02`. `BACnetPropertyValue::decode` then hands a truncated value/cursor to the WPM loop (`crates/bacnet-services/src/common.rs:187-240`; `crates/bacnet-services/src/wpm.rs:73-89`), which fails on the remaining payload. Single WriteProperty similarly rejects it when its priority/trailing parse sees bytes after the premature cursor (`crates/bacnet-services/src/write_property.rs:113-145`); ReadPropertyACK rejects the premature cursor as trailing data (`crates/bacnet-services/src/read_property.rs:196-198,214-216`).

This is a valid historical opaque payload under the compatibility contract: the direct EventParameter decoder deliberately preserves every byte between the initial wrapper and its terminal `FF FF` (`crates/bacnet-encoding/src/constructed/event_parameter.rs:186-203`). Thus the outer framing layer and the constructed decoder disagree about the value boundary, causing accidental rejection of legacy traffic.

The new multi-result tests place `FF FF` before a non-closing byte (`2` or an application-data byte), not the ambiguous `FF FF 2F`/`3F`/`4F` sequence, so they do not exercise this cursor transition (`crates/bacnet-services/src/wpm.rs:157-184`; `crates/bacnet-services/src/rpm.rs:468-499`).

## Verified flows

* Canonical tag-header rejection and boundary acceptance are implemented at `crates/bacnet-encoding/src/tags.rs:181-270` and covered by `crates/bacnet-encoding/src/tags_canonical_tests.rs:13-48`.
* Canonical opaque encoding is an application Octet String and round-trips through the constructed codec (`crates/bacnet-encoding/src/constructed/event_parameter.rs:154-157,209-222`).
* Property-aware extraction is used in ReadProperty, WriteProperty, WPM (`BACnetPropertyValue`), and RPM: `crates/bacnet-services/src/read_property.rs:196,214`; `crates/bacnet-services/src/write_property.rs:113-116`; `crates/bacnet-services/src/common.rs:187-197`; `crates/bacnet-services/src/rpm.rs:240-271`.
* Both simple and multiple write handler paths share the fully-consumed historical-wrapper canonicalization branch at `crates/bacnet-server/src/handlers/write_property.rs:455-469`; the simple WriteProperty handler path is proven to read back canonical bytes at `crates/bacnet-server/src/handlers/tests/framed_properties.rs:131-155`.

## Residual gaps

No tests cover historical payloads containing `FF FF` followed by the enclosing property close (`2F` WPM, `3F` WriteProperty/ReadProperty, `4F` RPM), nor a server-handler WPM historical write through canonical readback.

jscott3201 (commented):
[Runtime] READY

No blocker/high/medium findings.

Verified the complete `84e2457defeb0afcc2f93332bd85745dd06fa4f6...07760faed3a148603fff272031a2b9f295dcd905` diff. Canonical tag-number, extended-length, truncation, sanity-limit, and application L/V/T validation is ordered safely in `crates/bacnet-encoding/src/tags.rs:163-291`. Historical EventParameter decoding is exact-end and preserves interior `FF FF` bytes at `crates/bacnet-encoding/src/constructed/event_parameter.rs:183-203`, while new sentinel output is canonical Octet String framing at `crates/bacnet-encoding/src/constructed/event_parameter.rs:154-157`.

The service repair now selects the first complete legacy-close plus property-close sequence at `crates/bacnet-services/src/common.rs:47-71`, preventing cross-sibling consumption. That property-aware path is used by `BACnetPropertyValue` (`crates/bacnet-services/src/common.rs:187-197`), both ReadProperty paths, WriteProperty, and both RPM success branches (`crates/bacnet-services/src/rpm.rs:227-274`). The WPM and RPM regressions include a following sibling whose value ends in `FF FF` (`crates/bacnet-services/src/wpm.rs:157-182`; `crates/bacnet-services/src/rpm.rs:471-496`). Server write canonicalization also requires exact consumption (`crates/bacnet-server/src/handlers/write_property.rs:459-469`).

Commands/checks: verified the exact PR head and complete diff through GitHub; inspected all changed files and surrounding service/server callers. No local cargo command was available in this reviewer environment. Head clippy, rustfmt, file-size, and secret checks passed; the Ubuntu test check was still running at final observation, and broader matrix jobs were skipped.

Residual risk: the historical raw wrapper is intrinsically ambiguous if opaque payload bytes themselves contain `FF FF` immediately followed by the enclosing context closing octet; the first-complete-terminator policy favors preserving service sibling boundaries. The generic raw-context scanner retains its pre-existing documented delimiter ambiguity (`crates/bacnet-encoding/src/tags.rs:378-405`).

jscott3201 (commented):
[Adversary] NOT READY

### Medium — the new “first inner-close + outer-close” rule still rejects/truncates valid raw legacy payloads
The compatibility body is documented as arbitrary octets, but `extract_property_value` takes the *first* `FF FF <current outer close>` triple (`crates/bacnet-services/src/common.rs:53-63`). That triple can occur inside the raw legacy body. For a `BACnetPropertyValue` (outer close `2F`), encode an EventParameters value

`[FE, FF, 01, FF, FF, 2F, 02, FF, FF]`

where the intended raw payload is `[01, FF, FF, 2F, 02]`. The complete pre-existing wire field is `2E FE FF 01 FF FF 2F 02 FF FF 2F`. The scanner selects the `2F` at payload position 5 and returns only `[FE, FF, 01, FF, FF]` (`common.rs:60-62`), leaving the rest as apparent sibling/trailing bytes. Standalone `BACnetPropertyValue::decode` succeeds with that partial value; normal WPM/RPM/ReadProperty consumers then reject the remaining bytes rather than accepting the valid historical value. The same failure occurs for outer `3F` and `4F` in WriteProperty and RPM.

This is a concrete backward-compatibility gap: the old encoder placed raw bytes between `FE FF` and `FF FF` without escaping them, so this byte sequence is representable and must be preserved if the wrapper is supported. The new tests exercise `FF FF` only when it is not followed by the enclosing close (`constructed/tests/event_parameter.rs:286-296`; `wpm.rs:158-182`), so they do not cover this delimiter-bearing payload.

### Attacks checked
- First-marker extraction prevents the prior cross-sibling WPM/RPM swallow for payloads without `FF FF <outer-close>`.
- Isolated embedded `FF FF` is retained by terminal EventParameter decoding (`constructed/event_parameter.rs:186-203`).
- Legacy detection is property-gated to `EVENT_PARAMETERS` (`common.rs:53-55`, `handlers/write_property.rs:459-470`).
- Generic invalid extended numbers, noncanonical lengths, reserved application L/V/T forms, and truncated tag headers remain rejected (`bacnet-encoding/src/tags.rs:180-272`).
- Evaluator enforces full EventParameter consumption (`bacnet-server/src/event_enrollment/mod.rs:402-407`).

### Residual risk
An unescaped delimiter framing cannot simultaneously preserve arbitrary payload bytes and derive a reliable enclosing boundary from the same byte stream. The compatibility form needs an unambiguous length/boundary supplied by its actual enclosing parser, or an explicit policy that rejects delimiter-bearing historical values.

jscott3201 (commented):
[Dataflow] NOT READY

## P1 — grammar-shaped raw payloads still satisfy the new candidate-boundary heuristic

The legacy payload is documented as arbitrary octets (`crates/bacnet-services/src/common.rs:73-78`), but a candidate close is accepted when the next raw byte merely decodes as a permitted following context tag (`crates/bacnet-services/src/common.rs:79-89`; `matches_property_boundary` at `54-63`). Consequently, raw data can still impersonate the enclosing close plus the next grammar field.

Concrete WPM input: an `EVENT_PARAMETERS` `[2]` value with historical bytes `FE FF 01 FF FF 2F 09 00 FF FF`, followed by its real `[2]` close and the list `[1]` close. Its valid opaque payload is `01 FF FF 2F 09 00`. At the embedded `FF FF 2F`, the byte `09` is a valid primitive context `[0]` tag, so `PropertyValueBoundary::Context(0)` accepts the false candidate (`crates/bacnet-services/src/common.rs:207-220`). The first value is truncated to `FE FF 01 FF FF`; the WPM loop then treats raw `09 00` as a sibling property's identifier and fails when the next raw byte is not opening `[2]` (`crates/bacnet-services/src/wpm.rs:73-89`; `crates/bacnet-services/src/common.rs:244-254`). Thus valid historical opaque traffic is still rejected.

The same ambiguity remains in each context-aware path: a raw `FF FF 3F` followed by a context `[4]` header matches WriteProperty's priority boundary (`crates/bacnet-services/src/write_property.rs:113-126`); raw `FF FF 4F` followed by context `[2]` matches RPM's next-result boundary (`crates/bacnet-services/src/rpm.rs:244-253,276-285`). The added tests place a non-boundary byte (`2`) after the embedded close-like sequence, so they do not exercise a grammar-shaped collision (`crates/bacnet-services/src/wpm.rs:157-184`; `crates/bacnet-services/src/rpm.rs:486-517`).

## Verified flows

* Canonical tag rejection is bounded correctly at `crates/bacnet-encoding/src/tags.rs:181-270`, with boundary tests in `crates/bacnet-encoding/src/tags_canonical_tests.rs:13-48`.
* The structured codec preserves embedded close-like bytes until the terminal historical wrapper close and emits canonical application Octet String on readback (`crates/bacnet-encoding/src/constructed/event_parameter.rs:154-157,186-222`).
* ReadProperty, WriteProperty, RPM, WPM, CreateObject, and COV now route historical EventParameters through property-aware extraction: `crates/bacnet-services/src/read_property.rs:193-230`; `crates/bacnet-services/src/write_property.rs:113-126`; `crates/bacnet-services/src/rpm.rs:241-292`; `crates/bacnet-services/src/wpm.rs:81-89`; `crates/bacnet-services/src/object_mgmt.rs:105-129`; `crates/bacnet-services/src/cov.rs:340-372`.
* The server's shared write-value conversion requires full historical-wrapper consumption and the simple WriteProperty integration test proves canonical readback (`crates/bacnet-server/src/handlers/write_property.rs:455-469`; `crates/bacnet-server/src/handlers/tests/framed_properties.rs:131-155`).

## Residual gaps

No collision tests place a syntactically valid next-boundary tag immediately after an embedded `FF FF 2F`/`3F`/`4F`, and no handler-level WPM/CreateObject/COV legacy-property integration test verifies the final stored/read-back behavior.

jscott3201 (commented):
[Runtime] NOT READY

## Medium

- **A boundary-looking tag inside the opaque payload still wins even when the remaining service suffix proves it cannot be the real boundary.** `matches_property_boundary` validates only the next tag header (`crates/bacnet-services/src/common.rs:54-63`), and `extract_property_value` immediately returns the first candidate that passes that one-tag check (`crates/bacnet-services/src/common.rs:80-89`). For WriteProperty, `Context(4)` is accepted as a candidate boundary (`crates/bacnet-services/src/write_property.rs:116-125`), but priority validity and exact-end are checked only afterward (`crates/bacnet-services/src/write_property.rs:129-151`). Concrete old-format request: `0C 02 40 00 01 19 53 3E FE FF 01 FF FF 3F 49 08 02 FF FF 3F`. The historical opaque payload is `01 FF FF 3F 49 08 02`; the final `FF FF 3F` is the actual legacy/property terminator. The scanner instead stops at the embedded `FF FF 3F` because the following `49` looks like priority tag [4]. WriteProperty parses priority 8 and then rejects `02 FF FF 3F` as trailing data, even though continuing to the final candidate yields an exact, valid historical value. The regression test embeds `FF FF 3F` only before byte `02`, which cannot look like a service boundary (`crates/bacnet-services/src/read_property.rs:376-401`), so this failure mode remains untested. The same one-tag predicate can prematurely select embedded `Context(0)`/`Context(2)` candidates in list/RPM payloads before the candidate suffix is shown malformed.

## Commands/checks

- Verified exact target `84e2457defeb0afcc2f93332bd85745dd06fa4f6...6815e42da6195466f1bf039ac82f7b9abd20aea2` and inspected the complete diff plus all boundary-aware callers.
- No local cargo command was available in this reviewer environment.
- Head Ubuntu tests, clippy, rustfmt, file-size, and secret checks passed; broader matrix/security jobs were skipped.

## Residual risk

- Some historical payloads are intrinsically ambiguous when they contain a complete terminator followed by a fully valid service suffix. The concrete case above is not intrinsically ambiguous because the first candidate leaves trailing bytes and the final candidate consumes the request exactly.

jscott3201 (commented):
[Adversary] NOT READY

### Medium — an arbitrary legacy payload can forge the new immediate-boundary predicate and fabricate a sibling property
Candidate validation only checks that the byte immediately after `FF FF <outer-close>` *looks like* an allowed next tag (`crates/bacnet-services/src/common.rs:80-88`), while the body is explicitly arbitrary octets (`common.rs:76-78`). A raw body can therefore contain a complete fake terminator plus a syntactically valid successor.

Reproducer: encode a one-property WPM request whose sole `BACnetPropertyValue` is EventParameters and whose raw value is:

`FE FF AA FF FF 2F 09 53 2E FE FF BB FF FF`

(`09 53` is context [0] property-id 83 / EVENT_PARAMETERS; `2E` opens its value.) The surrounding encoder appends the real first-property `2F` and list `1F`.

The first scan accepts the embedded `FF FF 2F` because the next raw byte is `09`, a permitted `Context(0)` boundary (`common.rs:215-219`). It returns a first opaque value `[FE FF AA FF FF]`. The WPM loop then parses the raw `09 53 2E FE FF BB FF FF` as a *second* EventParameters property; its terminal is the original property's real `FF FF 2F`, followed by the allowed list close `1F`. `WritePropertyMultipleRequest::decode` succeeds with two properties instead of the one encoded, changing the stored final EventParameters value and defeating the sibling guard (`crates/bacnet-services/src/wpm.rs:81-88`).

The new test exercises an embedded `FF FF 2F` followed by raw `0x02`, which is not a valid boundary (`wpm.rs:158-182`); it does not exercise a forged valid boundary. RPM has the analogous Context(2)/Closing(1) predicate (`rpm.rs:244-285`).

### Attacks checked
- Generic `decode_tag` still rejects extended tag numbers outside 15..=254, noncanonical extended lengths, and application L/V/T 6/7 (`bacnet-encoding/src/tags.rs:180-272`).
- The special service extraction remains gated to `EVENT_PARAMETERS` (`common.rs:73-75`), and the write handler independently gates historical decoding to that property (`bacnet-server/src/handlers/write_property.rs:459-470`).
- Exact terminal EventParameter decoding and evaluator full-consumption checks are present (`constructed/event_parameter.rs:186-203`; `bacnet-server/src/event_enrollment/mod.rs:402-407`).

### Residual risk
An unescaped raw wrapper cannot safely infer its terminator from delimiter bytes plus a locally plausible next tag: arbitrary payload bytes can reproduce both. A compatibility strategy needs an unambiguous length/boundary from a trusted outer framing layer, or must reject ambiguous legacy streams.

jscott3201 (commented):
[Dataflow] READY

No blocker, high, or medium findings.

## Verified flows

* Generic decoding rejects reserved application L/V/T values, extended tag numbers outside 15..=254, and noncanonical extended lengths while retaining canonical boundary values: `crates/bacnet-encoding/src/tags.rs:181-270`; tests `crates/bacnet-encoding/src/tags_canonical_tests.rs:13-48`.
* The EventParameter codec confines the historical tag-255 exception to the constructed decoder, preserves raw bytes through the terminal historical marker, and emits the canonical application Octet String form: `crates/bacnet-encoding/src/constructed/event_parameter.rs:154-157,186-222`; test `crates/bacnet-encoding/src/constructed/tests/event_parameter.rs:270-301`.
* Legacy abstract-syntax extraction now requires a grammar-valid boundary and rejects a second candidate rather than choosing a cursor that could fabricate/suppress a property: `crates/bacnet-services/src/common.rs:54-114`. For single-value paths, End/ContextToEnd requires the optional suffix to consume the remainder: `crates/bacnet-services/src/common.rs:57-72,214-224`; `crates/bacnet-services/src/write_property.rs:113-126`; `crates/bacnet-services/src/read_property.rs:193-230`.
* Multiple-result/property containers use their actual next-element or list-close boundaries: RPM `crates/bacnet-services/src/rpm.rs:241-292`; WPM `crates/bacnet-services/src/wpm.rs:73-89`; CreateObject `crates/bacnet-services/src/object_mgmt.rs:105-129`; COV `crates/bacnet-services/src/cov.rs:340-372`. WPM/RPM include both sibling-preservation and ambiguous-boundary rejection coverage: `crates/bacnet-services/src/wpm.rs:157-208`; `crates/bacnet-services/src/rpm.rs:486-541`.
* Historical WriteProperty reaches the shared handler conversion only after full constructed-codec consumption, stores the sentinel as OctetString, and reads back canonical bytes: `crates/bacnet-server/src/handlers/write_property.rs:455-469`; `crates/bacnet-server/src/handlers/tests/framed_properties.rs:131-155`.
* The evaluator now requires full EventParameter consumption before evaluating: `crates/bacnet-server/src/event_enrollment/mod.rs:399-407`.

## Residual risks

The unescaped legacy delimiter is inherently ambiguous when arbitrary raw payload bytes imitate an enclosing close followed by a permitted grammar boundary. The implementation deliberately rejects multiple candidates (`crates/bacnet-services/src/common.rs:87-108`) under the stated fail-closed policy; such historical traffic will not migrate, but it is not split into fabricated properties. No server-handler end-to-end WPM/CreateObject/COV canonicalization test is present; their decode paths are covered and the shared write-value conversion is the same.

jscott3201 (commented):
[Runtime] READY

No blocker/high/medium findings.

Verified the complete `84e2457defeb0afcc2f93332bd85745dd06fa4f6...1c497fc574d35b916af7a6f882d7a8d8557d99f7` diff and exact-head surrounding callers. `ContextToEnd` now requires an optional priority candidate to consume the complete remaining service payload (`crates/bacnet-services/src/common.rs:55-72`; `crates/bacnet-services/src/write_property.rs:116-125`), so the previously distinguishable embedded `FF FF 3F 49 08 ...` case reaches the real final terminator. Legacy extraction records one grammar-valid candidate and fails closed on a second (`crates/bacnet-services/src/common.rs:75-114`). The WPM/RPM tests cover both a distinguishable close-like payload and grammar-ambiguous payload rejection (`crates/bacnet-services/src/wpm.rs:157-204`; `crates/bacnet-services/src/rpm.rs:489-535`). Canonical parsing, exact-end EventParameter decoding, server canonicalization, and all property-aware service paths remain intact.

No local cargo command was available in this reviewer environment. Head clippy, rustfmt, file-size, and secret checks passed; Ubuntu tests were still running at final observation, and broader matrix/security checks were skipped.

Residual risk: historical tag-255 framing is unescaped, so byte strings admitting two complete service-grammar parses cannot be recovered uniquely. Rejecting those inputs is the intended safe policy; canonical Octet String re-encoding removes this ambiguity for newly stored values.

jscott3201 (commented):
[Adversary] READY

No blocker, high, or medium findings survived the final adversarial review.

### Attacks checked
- Extended tag numbers 0..=14 and 255, every tag form, and truncated extended-number headers: generic rejection remains at `crates/bacnet-encoding/src/tags.rs:180-196`.
- Extended-length short forms, two-/four-octet noncanonical boundaries, application L/V/T 6/7, and truncated lengths: rejected at `tags.rs:180-272`.
- Historical `FE FF` payloads containing `FF FF`, `FF FF 2F`, `FF FF 3F`, and `FF FF 4F`: the compatibility decoder preserves isolated terminal wrappers; property extraction considers the enclosing grammar (`bacnet-services/src/common.rs:73-111`).
- Forged immediate sibling boundaries in WPM/RPM: more than one grammar-valid candidate is rejected before decoding a property (`common.rs:80-106`); corresponding regressions cover the prior WPM/RPM shapes (`wpm.rs:184-208`, `rpm.rs:517-540`).
- WriteProperty optional-priority candidates: only a context [4] that consumes the complete remaining suffix is eligible (`common.rs:61-69`, `write_property.rs:116-129`).
- Scope: the service bypass is gated to `EVENT_PARAMETERS` (`common.rs:73-75`; `bacnet-server/src/handlers/write_property.rs:459-470`); generic tag parsing remains strict. The event evaluator requires full value consumption (`bacnet-server/src/event_enrollment/mod.rs:402-407`).

### Residual risk
The deliberately fail-closed policy rejects historical streams with multiple grammar-valid wrapper boundaries. This is unavoidable for unescaped arbitrary legacy bytes and avoids silent truncation or sibling fabrication.

Issue #339, opened 2026-08-15
Implement ReadRange sequence-number semantics and First Sequence Number


jscott3201:
### Affected revision

Observed on `dev` at `677fadf` while hardening ReadRange decoding under #309.

### Behavior

`handle_read_range` currently treats a By-Sequence `reference_seq` as a one-based list position and always encodes `first_sequence_number: None`, including nonempty By-Sequence responses.

ANSI/ASHRAE 135-2020 Clause 15.8.1.1.4.2 defines sequence numbers as values assigned when records are added, not list positions. Clause 15.8.1.2 requires First Sequence Number for nonempty responses to By-Sequence or By-Time requests.

This prevents a client from validating the response condition without rejecting responses from the in-tree server.

### Relevant code

- `crates/bacnet-server/src/handlers/file.rs:60-81`
- `crates/bacnet-server/src/handlers/file.rs:101-109`
- `crates/bacnet-services/src/read_range.rs`

### Suggested direction

- derive record sequence numbers from the object/log model rather than list position;
- select By-Sequence ranges using those values;
- emit First Sequence Number when the returned item count is nonzero;
- add server/client integration coverage for positive and negative counts, wrapped or purged log buffers, and empty results;
- keep By-Time support as an explicit supported or rejected path until timestamp selection is implemented.

PR #340, opened 2026-08-15
fix(services): harden ReadRange decoding

jscott3201:
## Summary

- Reject ReadRange numeric overflow and malformed request/ACK framing before server lookup or client exposure.
- Preserve the existing public `u32`/`i32` fields and accept numerically in-range leading-zero Unsigned encodings.
- Reduce the #309 direct-cast inventory from 72 to 64 sites.

## Standard scope

- ANSI/ASHRAE 135-2020 Clause 15.8.1.1: request fields, prohibited special property identifiers, nonzero array index, optional Range CHOICE, nonzero `INTEGER16` Count, and specific By-Time datetime.
- Clause 15.8.1.2: Result Flags, Item Count, Item Data, and conditional First Sequence Number.
- Clause 21: `ReadRange-Request` and `ReadRange-ACK` formal productions.
- Clause 3: definitions of specific date, time, and datetime.

No conformance ledger or public support claim changes are included.

## Implementation

- Replace eight unchecked `u64 as u32` conversions with value-based checked decoding.
- Require the specified context and application tags, complete Range members, exact three-bit Result Flags, and complete service payload consumption.
- Reject `ALL`, `REQUIRED`, and `OPTIONAL`, zero request array indexes, invalid By-Time components, zero or out-of-range `INTEGER16` counts, impossible First Sequence Number values, and empty/nonempty Item Count contradictions.
- Validate echoed object, property, and array identifiers in the Rust client before returning an ACK.
- Include `first_sequence_number` in the Python result dictionary.

Valid wire output and public Rust/Python method signatures are unchanged. The Python result dictionary gains the decoded `first_sequence_number` field.

## Tests

Commands run:

```bash
cargo check -p rusty-bacnet --tests --locked
cargo test --workspace --exclude rusty-bacnet --locked
cargo test --workspace --exclude rusty-bacnet --locked --features bacnet-types/serde,bacnet-transport/ipv6,bacnet-transport/sc-tls
cargo clippy --workspace --exclude rusty-bacnet --all-targets --locked
cargo fmt --all -- --check
git diff --cached --check
bash .github/scripts/check-file-size.sh
bash .github/scripts/check-no-secrets.sh
```

New tests cover every narrowed field, `u32::MAX` with a leading zero, overflow and `u64::MAX`, `INTEGER16` boundaries, wrong tags, invalid CHOICE members, incomplete/trailing sequences, malformed Result Flags, nonspecific datetimes, prohibited property selectors, Item Count contradictions, and request/ACK correlation.

## Known limitations

- Public `u32` fields do not represent the Unsigned64 reference values required for object types with Unsigned64 `Total_Record_Count`; this PR preserves the existing API boundary.
- The in-tree server still models By-Sequence selection as list position and omits First Sequence Number. That behavior is tracked in #339; the client rejects impossible First Sequence Number values but remains compatible with an omitted value.
- ComplexACK service-choice correlation belongs to the client TSM/APDU layer and is tracked in #341.
- Shared nested context-tag validation remains tracked in #327.

## Performance

No hot-path or allocation strategy changes; benchmarks were not required.

Progresses #309.
Follow-ups: #339, #341.

### DIFF
--- a/crates/bacnet-client/src/client/file_list.rs
+++ b/crates/bacnet-client/src/client/file_list.rs
@@ -1,5 +1,46 @@
 use super::*;
 
+fn validate_read_range_ack(
+    request: &bacnet_services::read_range::ReadRangeRequest,
+    ack: &bacnet_services::read_range::ReadRangeAck,
+) -> Result<(), Error> {
+    if ack.object_identifier != request.object_identifier {
+        return Err(Error::decoding(
+            0,
+            "ReadRange ACK object identifier does not match the request",
+        ));
+    }
+    if ack.property_identifier != request.property_identifier {
+        return Err(Error::decoding(
+            0,
+            "ReadRange ACK property identifier does not match the request",
+        ));
+    }
+    if ack.property_array_index != request.property_array_index {
+        return Err(Error::decoding(
+            0,
+            "ReadRange ACK array index does not match the request",
+        ));
+    }
+
+    let sequence_range = matches!(
+        request.range.as_ref(),
+        Some(
+            bacnet_services::read_range::RangeSpec::BySequenceNumber { .. }
+                | bacnet_services::read_range::RangeSpec::ByTime { .. }
+        )
+    );
+    let permits_first_sequence_number = sequence_range && ack.item_count > 0;
+    if ack.first_sequence_number.is_some() && !permits_first_sequence_number {
+        return Err(Error::decoding(
+            0,
+            "ReadRange ACK first sequence number is invalid for the request range",
+        ));
+    }
+
+    Ok(())
+}
+
 impl<T: TransportPort + 'static> BACnetClient<T> {
     /// Get event information from a remote device.
     pub async fn get_event_information(
@@ -80,7 +121,9 @@ impl<T: TransportPort + 'static> BACnetClient<T> {
             .confirmed_request(destination_mac, ConfirmedServiceChoice::READ_RANGE, &buf)
             .await?;
 
-        ReadRangeAck::decode(&response_data)
+        let ack = ReadRangeAck::decode(&response_data)?;
+        validate_read_range_ack(&request, &ack)?;
+        Ok(ack)
     }
 
     /// Read file data from a remote device (stream or record access).
@@ -191,5 +234,93 @@ impl<T: TransportPort + 'static> BACnetClient<T> {
             .await?;
 
         Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use bacnet_services::read_range::{RangeSpec, ReadRangeAck, ReadRangeRequest};
+    use bacnet_types::enums::{ObjectType, PropertyIdentifier};
+    use bacnet_types::primitives::{Date, ObjectIdentifier, Time};
+
+    fn request(range: Option<RangeSpec>) -> ReadRangeRequest {
+        ReadRangeRequest {
+            object_identifier: ObjectIdentifier::new(ObjectType::TREND_LOG, 1).unwrap(),
+            property_identifier: PropertyIdentifier::LOG_BUFFER,
+            property_array_index: Some(1),
+            range,
+        }
+    }
+
+    fn ack(request: &ReadRangeRequest, item_count: u32, first: Option<u32>) -> ReadRangeAck {
+        ReadRangeAck {
+            object_identifier: request.object_identifier,
+            property_identifier: request.property_identifier,
+            property_array_index: request.property_array_index,
+            result_flags: (true, true, false),
+            item_count,
+            item_data: Vec::new(),
+            first_sequence_number: first,
+        }
+    }
+
+    #[test]
+    fn read_range_ack_must_echo_the_request() {
+        let request = request(Some(RangeSpec::ByPosition {
+            reference_index: 1,
+            count: 1,
+        }));
+        let valid = ack(&request, 1, None);
+        assert!(validate_read_range_ack(&request, &valid).is_ok());
+
+        let mut wrong_object = valid.clone();
+        wrong_object.object_identifier = ObjectIdentifier::new(ObjectType::TREND_LOG, 2).unwrap();
+        assert!(validate_read_range_ack(&request, &wrong_object).is_err());
+
+        let mut wrong_property = valid.clone();
+        wrong_property.property_identifier = PropertyIdentifier::PRESENT_VALUE;
+        assert!(validate_read_range_ack(&request, &wrong_property).is_err());
+
+        let mut wrong_index = valid;
+        wrong_index.property_array_index = Some(2);
+        assert!(validate_read_range_ack(&request, &wrong_index).is_err());
+    }
+
+    #[test]
+    fn first_sequence_number_must_match_the_requested_range() {
+        let by_sequence = request(Some(RangeSpec::BySequenceNumber {
+            reference_seq: 1,
+            count: 1,
+        }));
+        assert!(validate_read_range_ack(&by_sequence, &ack(&by_sequence, 1, Some(1))).is_ok());
+        assert!(validate_read_range_ack(&by_sequence, &ack(&by_sequence, 1, None)).is_ok());
+        assert!(validate_read_range_ack(&by_sequence, &ack(&by_sequence, 0, None)).is_ok());
+        assert!(validate_read_range_ack(&by_sequence, &ack(&by_sequence, 0, Some(1))).is_err());
+
+        let by_time = request(Some(RangeSpec::ByTime {
+            reference_time: (
+                Date {
+                    year: 126,
+                    month: 3,
+                    day: 1,
+                    day_of_week: 7,
+                },
+                Time {
+                    hour: 14,
+                    minute: 30,
+                    second: 0,
+                    hundredths: 0,
+                },
+            ),
+            count: 1,
+        }));
+        assert!(validate_read_range_ack(&by_time, &ack(&by_time, 1, Some(1))).is_ok());
+
+        let by_position = request(Some(RangeSpec::ByPosition {
+            reference_index: 1,
+            count: 1,
+        }));
+        assert!(validate_read_range_ack(&by_position, &ack(&by_position, 1, Some(1))).is_err());
     }
 }
--- a/crates/bacnet-services/src/read_range.rs
+++ b/crates/bacnet-services/src/read_range.rs
@@ -8,18 +8,47 @@ use bacnet_types::error::Error;
 use bacnet_types::primitives::{Date, ObjectIdentifier, Time};
 use bytes::BytesMut;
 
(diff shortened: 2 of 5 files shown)


jscott3201 (commented):
[Runtime] NOT READY

Reviewed head: aea7da2d2803a1426b77069a8ba00357dbf5fd36

**High — A valid specific By-Time date is rejected when day-of-week is unspecified.** `crates/bacnet-services/src/read_range.rs:217-224` requires `date.day_of_week` to be `1..=7`. Clause 3's specific-date definition requires the year, month, and day-of-month to be specified, but permits day-of-week to be unspecified; that date combined with a fully specified time satisfies Clause 15.8's By-Time requirement. Concrete reproduction: decode `0c 05 00 00 01 19 83 7e a4 7e 03 01 ff b4 0e 1e 00 00 31 01 7f` (TrendLog:1, Log_Buffer, By-Time date 2026-03-01 with day-of-week `0xff`, time 14:30:00.00, Count 1). The new condition returns `ReadRange byTime requires a specific datetime`, so a conformant request is rejected before the server's object lookup. The same `Date` is accepted by the public encoder (`crates/bacnet-services/src/read_range.rs:109-120`), making this valid encode/decode round trip fail. The regression test at `crates/bacnet-services/src/read_range_width_tests.rs:253-273` currently codifies the rejection by treating offset 5 / `0xff` as nonspecific.

Residual low risks: `item_data` remains raw and therefore cannot receive property-specific semantic validation in this generic decoder. The explicitly scoped server selection/First Sequence Number behavior (#339), Unsigned64 expansion, and nested-tag handling (#327) remain out of scope.

jscott3201 (commented):
[Adversary] NOT READY

Reviewed head: aea7da2d2803a1426b77069a8ba00357dbf5fd36

Findings:

1. **Medium — valid By-Time requests with an unspecified day-of-week are rejected.** Clause 3's specific-date definition requires a specific year, month, and day-of-month; the redundant day-of-week may be unspecified. `crates/bacnet-services/src/read_range.rs:217-224` instead requires `day_of_week` to be `1..=7`. A request whose Date content is `[0x7e, 0x03, 0x01, 0xff]` (2026-03-01 with unspecified day-of-week), followed by a fully specified Time and Count 1, now fails decoding and the server returns an error rather than serving a conforming ReadRange request. Allow `Date::UNSPECIFIED` for day-of-week while continuing to reject unspecified/special year, month, and day values.

2. **Medium — Item Count is exposed without even basic consistency validation against Item Data.** `crates/bacnet-services/src/read_range.rs:355-369` decodes `item_count` and copies the `[5]` contents independently. Thus an ACK with `itemCount = 1` and an immediately closed, empty `[5]` decodes successfully; with matching echoed fields and no First Sequence Number, `crates/bacnet-client/src/client/file_list.rs:1-39` also accepts it and returns `item_count == 1` with empty `item_data`. The inverse (`itemCount = 0` with nonempty Item Data) also passes. Clause 15.8.1.2 defines Item Count as the number of items in Item Data, so these are malformed ACKs and can mislead callers or unsafe downstream preallocation. At minimum reject the unambiguous empty/nonempty contradictions before returning the ACK; exact counting may remain property-syntax-aware.

Residual risks: the explicitly excluded nested context matching (#327), server sequence semantics/missing First Sequence Number (#339), Unsigned64 API width, and encoder validation remain. I found no additional panic or silent narrowing path in the changed decoders.

jscott3201 (commented):
[Dataflow] NOT READY
Reviewed head: aea7da2d2803a1426b77069a8ba00357dbf5fd36

Findings:

1. **HIGH — missing mandatory Range CHOICE still reaches object lookup.** Clause 21 makes `range` mandatory, but `ReadRangeRequest::decode` accepts a payload ending after property/optional array index (`crates/bacnet-services/src/read_range.rs:252-265`). A confirmed READ_RANGE APDU with only [0] and [1] therefore returns `Ok(range: None)`; the dispatch reaches `handle_read_range`, which calls decode and then `db.get` (`crates/bacnet-server/src/handlers/file.rs:14-16`). The same malformed request is publicly constructible through Python's `range_type=None` (`crates/rusty-bacnet/src/client/client_methods/object_device_alarm.rs:248-263`). This violates the required failure-before-lookup behavior and preserves the old non-standard whole-list extension.

2. **HIGH — the Rust client does not correlate ComplexACK service choice.** Client APDU dispatch completes the TSM for any unsegmented ComplexACK matching only source/invoke ID, without checking `ack.service_choice` (`crates/bacnet-client/src/client/dispatch.rs:31-50`). `read_range` then decodes the payload and applies only object/property/array and First Sequence Number checks (`crates/bacnet-client/src/client/file_list.rs:121-123`). A peer sending a valid ReadRange-ACK payload inside a ComplexACK labelled for another service is consequently returned to Rust and Python. This is an APDU-level mismatched ACK that the new payload echo checks do not catch.

3. **MEDIUM — Item Count and Item Data are not cross-validated.** The decoder parses both fields but never enforces that the sequence contains exactly `item_count` items (`crates/bacnet-services/src/read_range.rs:356-371`). For example, `itemCount=0` with `itemData` containing one valid application NULL is accepted; the Rust client validator does not reject it, and Python exposes the inconsistent pair (`crates/rusty-bacnet/src/client/client_methods/object_device_alarm.rs:288-291`). A positive count with empty Item Data is likewise accepted. This is a malformed ACK reaching callers despite the stated Clause 15.8.1.2 scope.

4. **MEDIUM — Python drops First Sequence Number after Rust validation.** A valid By-Sequence/By-Time ACK's `first_sequence_number` is decoded and permitted by the Rust client, but the Python dictionary never includes it (`crates/rusty-bacnet/src/client/client_methods/object_device_alarm.rs:273-291`). Thus the compatibility-tolerated omission is handled, but a present sequence number is silently lost at the Rust/Python boundary, preventing Python callers from using the sequence metadata required to interpret/paginate the returned records.

Verified flows:
- All eight former unchecked Unsigned-to-u32 sites are now covered by checked `u32::try_from` paths, and the width tests cover leading-zero `u32::MAX`, overflow, and `u64::MAX`.
- Wrong context/application tags, incomplete range members, zero array indexes, prohibited property selectors, INTEGER16 bounds/zero, specific By-Time components, Result Flags, and trailing service bytes fail in the service decoder before the handler's object lookup.
- Rust object/property/array echo mismatches fail before the ACK is returned; Python delegates through that Rust method and maps the error.

Residual risks intentionally deferred: the server's By-Sequence list-position semantics and omitted First Sequence Number (#339), shared nested context validation (#327), and the public Unsigned64 boundary. These do not explain the findings above.

jscott3201:
Review follow-up on `b66eb9b`:

- Added empty/nonempty Item Count consistency checks and hostile ACK tests. Exact item cardinality remains property-syntax-aware because Item Data is `SEQUENCE OF ABSTRACT-SYNTAX.&Type`.
- Added `first_sequence_number` to the Python result dictionary and updated `docs/python-api.md`.
- Kept omitted Range valid: Clause 15.8.1.1.4 calls it optional and the Clause 21 production ends the Range CHOICE with `OPTIONAL`.
- Kept `0xFF` day-of-week invalid for By-Time: Clause 15.8 requires a specific datetime, and Clause 3 defines a specific date as containing no unspecified octets or special date values.
- Tracked APDU-wide ComplexACK service-choice correlation separately in #341; server sequence semantics remain in #339.

The full local default/feature workspace matrices, bindings check, clippy, formatting, file-size, secret, and diff checks pass on the follow-up tree.

jscott3201 (commented):
[Runtime] READY

Reviewed head: b66eb9b6e53aef6b83c3d2e690e0513972f28534

No blocker, high, or medium findings.

The complete `base_sha...head_sha` change now enforces the ReadRange request/ACK tag classes and sequence framing, checked `u32` narrowing with in-range leading-zero compatibility, nonzero `INTEGER16` count bounds, specific By-Time date/time fields, exact three-bit Result Flags, complete payload consumption, Item Count empty/nonempty consistency, and request-correlated First Sequence Number conditions. The Python result also exposes the decoded `first_sequence_number`.

Residual/out-of-scope risks remain the public Unsigned64 boundary, server sequence semantics (#339), APDU service-choice correlation (#341), and shared nested-tag handling (#327). Property-specific interpretation and exact counting of arbitrary raw `item_data` remain downstream responsibilities beyond the new empty/nonempty contradiction check.

jscott3201 (commented):
[Adversary] READY

Reviewed head: b66eb9b6e53aef6b83c3d2e690e0513972f28534

No blocker, high, or medium findings survived the final adversarial pass. The changed decoders reject checked-width overflow and malformed mandatory/range framing, enforce complete payload consumption, and the Item Count parity check at `crates/bacnet-services/src/read_range.rs:359-376` correctly rejects both zero-with-data and nonzero-with-empty-data without a panic path. The Rust client validates echoed identifiers and impossible present First Sequence Number contexts before returning the ACK (`crates/bacnet-client/src/client/file_list.rs:3-39,124-126`), and the Python result now exposes the validated optional value.

Residual risks (non-blocking and already scoped): exact Item Count cardinality and property-specific primitive semantics cannot be established from generic raw Item Data; nested context matching remains #327; server sequence behavior/missing First Sequence Number remains #339; APDU-wide service-choice correlation remains #341; and the existing `u32` API/encoder boundaries remain unchanged.

jscott3201 (commented):
[Dataflow] READY
Reviewed head: b66eb9b6e53aef6b83c3d2e690e0513972f28534

No blocker, high, or medium findings remain for this service tranche.

Verified flows:
- Confirmed READ_RANGE dispatch reaches `handle_read_range`, whose first operation is `ReadRangeRequest::decode` before `db.get` (`crates/bacnet-server/src/handlers/file.rs:14-16`). The decoder now rejects malformed tags, incomplete members, prohibited selectors, zero array indexes, invalid counts/datetimes, invalid choices, and trailing bytes; the locally verified Standard permits an omitted Range CHOICE.
- All eight former narrowing sites use checked value conversion, with leading-zero `u32::MAX`, overflow/`u64::MAX`, and INTEGER16 boundary tests in `crates/bacnet-services/src/read_range_width_tests.rs`.
- ACK decoding enforces mandatory context tags, exact three-bit Result Flags, Item Data framing/Item Count empty-vs-nonempty consistency, conditional First Sequence Number, and complete payload consumption (`crates/bacnet-services/src/read_range.rs:317-409`).
- Rust client validation checks object, property, and array-index echoes and rejects impossible First Sequence Number presence before returning the ACK (`crates/bacnet-client/src/client/file_list.rs:3-48,121-123`). Python receives that validated result and now exposes `first_sequence_number` (`crates/rusty-bacnet/src/client/client_methods/object_device_alarm.rs:273-293`); the public API documentation is updated (`docs/python-api.md:674`).

Residual risks are the explicitly deferred Unsigned64 public boundary, shared nested context-tag validation (#327), and server By-Sequence/list-position plus omitted First Sequence Number behavior (#339). ComplexACK service-choice correlation is isolated in #341 and was treated as outside this one-service tranche.

Issue #346, opened 2026-08-15
BACnetPropertyStates variants use incorrect Clause 21 context tags


jscott3201:
## Problem

`BACnetPropertyStates` and its service codec diverge from the ANSI/ASHRAE 135-2020 Clause 21 CHOICE after tag 13. This prevents the 22 remaining #309 casts in `alarm_event/property_states.rs` from being safely converted as a numeric-only slice because several decoded values currently become the wrong public variant.

## Evidence

Clause 21 assigns, among others:

- `[14]` restart-reason, `[15]` door-alarm-state, `[16]` action, `[17]` door-secured-status, `[18]` door-status, `[19]` door-value.
- `[38]` lighting-in-progress, `[39]` lighting-operation, `[40]` lighting-transition, `[42]` binary-lighting-value.
- `[43]` timer-state, `[44]` timer-transition, `[52]` lift-car-direction, `[53]` lift-car-door-command.

The current enum/codec maps `[14]` to `DoorAlarmState`, `[15]` to `Action`, `[16]` to `DoorSecuredStatus`, `[17]` to `DoorStatus`, `[18]` to `DoorValue`, `[38]` to `TimerState`, `[39]` to `TimerTransition`, `[40]` to `LiftCarDirection`, and `[42]` to `LiftCarDoorCommand`. It also omits many Standard alternatives.

Local source anchor: extracted Standard text lines 60464-60548. Code: `crates/bacnet-types/src/constructed/mod.rs` and `crates/bacnet-services/src/alarm_event/property_states.rs`.

## Expected work

1. Correct the public variant/tag model and add the missing Standard alternatives.
2. Preserve proprietary tags 64-254 and implement tag 63 extended-value semantics.
3. Enforce primitive context tags, strict Boolean encoding, and checked destination widths.
4. Add a table-driven golden test for every modeled tag plus hostile overflow and malformed-form cases.
5. Update all notification-parameter consumers and then remove the associated #309 casts.

Treat this as an API/wire-model correction, not a mechanical cast replacement.

Issue #345, opened 2026-08-15
Audit service codecs do not match the Clause 21 wire model


jscott3201:
## Problem

`crates/bacnet-services/src/audit.rs` exposes request models and codecs that diverge from ANSI/ASHRAE 135-2020 Clauses 13.19-13.21 and the Clause 21 productions. This prevents the four remaining #309 casts in that module from being treated as a numeric-only hardening slice.

## Evidence

Clause 21 defines:

- `ConfirmedAuditNotification-Request` and `UnconfirmedAuditNotification-Request` as `notifications [0] SEQUENCE OF BACnetAuditNotification`.
- `AuditLogQuery-Request` as audit log `[0]`, query parameters `[1]`, optional start sequence `[2]`, and requested count `[3]`.
- `BACnetAuditNotification` with optional source timestamp, source user ID `[8]`, source user role `[9]`, target device `[10]`, target property `[12]`, target priority `[13]`, and result `[16]`.

The current code:

- Models one notification directly and omits the request-level `[0] SEQUENCE OF` wrapper.
- Requires source timestamp and shifts/replaces fields after invoke ID: raw source-user-info `[8]`, target device `[9]`, target property `[11]`, target priority `[12]`, and result `[15]`.
- Models `AuditLogQueryRequest` as `acknowledgment_filter [0]` plus opaque trailing bytes, which is unrelated to the Standard request production.
- Cites Clauses 15.2.8/15.2.9 instead of 13.19-13.21.

Local source anchors: extracted Standard text lines 54804-54806, 55001-55012, 55218-55220, and 56414-56452.

## Expected work

1. Define Standard-shaped request and base types, including notification lists and both query alternatives.
2. Decide and document the public Rust/Python migration because existing fields and constructors are incompatible.
3. Add byte-level golden, boundary, malformed-tag, list-limit, and end-to-end client/server tests.
4. Then remove the four direct narrowing casts tracked by #309.

Do not broaden public support or conformance claims until the wire model and runtime paths are complete.

jscott3201:
PR-0107 is complete through PR #457 (squash merge ea274472c8839d2218376934990dbb5c2f9945ec; immutable reviewed head 0702604554a72fe3500722c86017839e563b25bf).

[...]

Issue #171, opened 2026-07-27
Event_Time_Stamps and Event_Message_Texts are BACnetARRAY[3] but the read path discards the array index


jscott3201:
### Affected revision

Observed on `dev` at commit `3990f3e`.

### Behavior

`macro_rules! read_event_properties!` (`crates/bacnet-objects/src/common.rs:353-354`) takes `($self:expr, $property:expr)` and no array-index parameter. Its `EVENT_TIME_STAMPS` arm (`:403-430`) and `EVENT_MESSAGE_TEXTS` arm (`:431-443`) unconditionally build a three-element `PropertyValue::List`. All three analog call sites invoke it without an index (`crates/bacnet-objects/src/analog/mod.rs:121`, `:358`, `:637`) even though the enclosing `read_property` has `array_index: Option<u32>` in scope. The three binary object types have the same defect in their own arms rather than through the macro: `crates/bacnet-objects/src/binary/mod.rs:133-137`, `:361` and `:599` each return a hardcoded three-element list while `array_index` is in scope at `:85`.

Upstream, `crates/bacnet-server/src/handlers/read_property.rs:36-37` lists both properties in the `is_array_property` whitelist, so an indexed read passes the gate at `:43-48` instead of being rejected, `:51` forwards the index to the object, and `:59` echoes `property_array_index` back into the ACK. The client is told it received element N and is handed all three.

Clause 12.1.5.1, Array Properties (p164): "An index of 0 (zero) shall specify the count of the number of data elements as datatype Unsigned. If the array index is omitted, it means that all of the elements of the array are to be accessed. An array index N, greater than zero, shall specify the Nth element in the sequence." Table 12-2 (p168) types both properties as `BACnetARRAY[3]`, of `BACnetTimeStamp` and `CharacterString` respectively.

The whitelist itself is right — these genuinely are arrays. The gap is that nothing below it honours the index.

This constrains #123 and #128 directly: once `Event_Time_Stamps` carries real per-transition values, a client reading index 2 for the TO_FAULT timestamp still gets all three.

Separately, each element is emitted as a bare `PropertyValue::Unsigned` rather than a `BACnetTimeStamp` CHOICE, so even an unindexed read is not the declared datatype. That is the object-side half of the Time Stamp encoding issue filed alongside this one and should be fixed in the same pass.

### Suggested direction

- give `read_event_properties!` an `$array_index` parameter and thread the caller's `array_index` through all three analog call sites;
[...]

PR #347, opened 2026-08-15
fix(services): harden GetEventInformation end to end

jscott3201:
## Summary

- Reject malformed or overflowing GetEventInformation request and ACK fields without changing public types.
- Remove per-summary transition-bit allocations from ACK decoding.
- Make server responses resumable in numeric object-identifier order and preserve Notification Class priorities and valid stored sequence timestamps.
- Reduce the #309 direct-cast inventory from 51 to 48 sites.

## Standard scope

- ANSI/ASHRAE 135-2020 Clause 13.12: GetEventInformation request, event-summary, and More Events semantics.
- Clause 21: `GetEventInformation-Request` and `GetEventInformation-ACK` productions.
- Clause 20.2.10: three-bit `BACnetEventTransitionBits` encoding and zero padding.

No public Rust/Python signature or conformance-ledger claim changes.

## Implementation

- Require optional request field `[0]` to own the complete nonempty payload.
- Decode Event State, Notify Type, and all three Event Priorities with value-based `u32` checks. Numerically valid five-octet leading-zero encodings remain accepted.
- Require exact context tags `[0]` through `[6]`, exactly three timestamps, exactly three application Unsigned priorities, matching constructed closings, a strict More Events Boolean, and complete ACK consumption.
- Parse fixed-width transition fields directly from their two content octets, avoiding the generic decoder owned `Vec` for up to 20,000 fields at the accepted summary limit.
- Cap decoded event summaries at `MAX_DECODED_ITEMS`.
- Sort server candidates by encoded object identifier and treat the request cursor as an exclusive numeric boundary. Pagination continues when the cursor object was removed between pages.
- Match Notification Class objects by their `NOTIFICATION_CLASS` property rather than assuming the class number equals the object instance, then project the three-item `PRIORITY` list into the ACK.
- Forward object API Event Time Stamps sequence values in the BACnet `0..=65535` range. A malformed array defaults to three zero sequence numbers without aborting other summaries. Other timestamp alternatives remain lossy in the object API pending #171.
- Correct touched alarm/event documentation to the 135-2020 Clause 13.5, 13.8-13.9, and 13.12 numbering.

## Tests

Commands run:

```bash
cargo test -p bacnet-services --locked
cargo test -p bacnet-server --locked
cargo test -p bacnet-client --locked
cargo check -p rusty-bacnet --tests --locked
cargo test --workspace --exclude bacnet-benchmarks --exclude rusty-bacnet --all-targets --locked
cargo test --workspace --exclude bacnet-benchmarks --exclude rusty-bacnet --all-targets --no-default-features --locked
cargo clippy --workspace --all-targets --locked -- -A missing-docs -A unused
cargo fmt --all -- --check
git diff --check
bash .github/scripts/check-file-size.sh
bash .github/scripts/check-no-secrets.sh
```

Decoder coverage includes all five numeric values at padded `u32::MAX`, `u32::MAX + 1`, and `u64::MAX`; every request/ACK context tag; wrong priority datatypes; two/four-element timestamp and priority sequences; malformed transition bits and Boolean values; trailing data; and the 10,000/10,001 boundary.

Server integration coverage decodes generated ACKs to verify a class number that differs from its Notification Class object instance, custom priorities, the sequence-number ceiling, malformed timestamp containment, and pagination across 27 reverse-inserted active objects after deleting the first-page cursor.

## Review

- Runtime: READY on `00b5c1ae8d00fe3862b979b9a9903331690ed29d`.
- Dataflow: READY on `00b5c1ae8d00fe3862b979b9a9903331690ed29d`.

## Known follow-ups

- Audit service codecs require a public wire-model correction before their #309 casts can be removed: #345.
- `BACnetPropertyStates` variants require a public tag-model correction before their #309 casts can be removed: #346.
- EventNotification and nested notification-parameter casts remain separate slices.

Progresses #309.

### DIFF
--- a/crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs
+++ b/crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs
@@ -19,16 +19,20 @@ pub fn handle_get_event_information(
     let request = GetEventInformationRequest::decode(service_data)?;
 
     let mut summaries = Vec::new();
-    let mut skipping = request.last_received_object_identifier.is_some();
     let mut more_events = false;
+    let cursor = request
+        .last_received_object_identifier
+        .map(|identifier| identifier.encode());
+    let mut object_identifiers = db.list_objects();
+    object_identifiers.sort_unstable_by_key(ObjectIdentifier::encode);
 
-    for (oid, object) in db.iter_objects() {
-        if skipping {
-            if Some(oid) == request.last_received_object_identifier {
-                skipping = false;
-            }
+    for oid in object_identifiers {
+        if cursor.is_some_and(|cursor| oid.encode() <= cursor) {
             continue;
         }
+        let Some(object) = db.get(&oid) else {
+            continue;
+        };
 
         // Clause 12.12: Event_Detection_Enable controls whether the object is
         // considered by the event-summarization services. Checked after the
@@ -50,7 +54,7 @@ pub fn handle_get_event_information(
                     .read_property(PropertyIdentifier::NOTIFICATION_CLASS, None)
                     .ok()
                     .and_then(|v| match v {
-                        PropertyValue::Unsigned(n) => Some(n as u32),
+                        PropertyValue::Unsigned(n) => u32::try_from(n).ok(),
                         _ => None,
                     })
                     .unwrap_or(0);
@@ -85,8 +89,20 @@ pub fn handle_get_event_information(
                                 .ok()
                         })
                         .and_then(|v| match v {
-                            PropertyValue::OctetString(bytes) if bytes.len() == 3 => {
-                                Some([bytes[0] as u32, bytes[1] as u32, bytes[2] as u32])
+                            PropertyValue::List(items) => {
+                                let [
+                                    PropertyValue::Unsigned(offnormal),
+                                    PropertyValue::Unsigned(fault),
+                                    PropertyValue::Unsigned(normal),
+                                ] = items.as_slice()
+                                else {
+                                    return None;
+                                };
+                                Some([
+                                    u32::try_from(*offnormal).ok()?,
+                                    u32::try_from(*fault).ok()?,
+                                    u32::try_from(*normal).ok()?,
+                                ])
                             }
                             _ => None,
                         })
@@ -107,7 +123,21 @@ pub fn handle_get_event_information(
                     .read_property(PropertyIdentifier::EVENT_TIME_STAMPS, None)
                     .ok()
                     .and_then(|v| match v {
-                        PropertyValue::List(items) if items.len() == 3 => None,
+                        PropertyValue::List(items) => {
+                            let [
+                                PropertyValue::Unsigned(offnormal),
+                                PropertyValue::Unsigned(fault),
+                                PropertyValue::Unsigned(normal),
+                            ] = items.as_slice()
+                            else {
+                                return None;
+                            };
+                            Some([
+                                BACnetTimeStamp::SequenceNumber(*offnormal),
+                                BACnetTimeStamp::SequenceNumber(*fault),
+                                BACnetTimeStamp::SequenceNumber(*normal),
+                            ])
+                        }
                         _ => None,
                     })
                     .unwrap_or([
--- a/crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs
+++ b/crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs
@@ -25,6 +25,7 @@ pub fn handle_get_event_information(
         .map(|identifier| identifier.encode());
     let mut object_identifiers = db.list_objects();
     object_identifiers.sort_unstable_by_key(ObjectIdentifier::encode);
+    let notification_class_identifiers = db.find_by_type(ObjectType::NOTIFICATION_CLASS);
 
     for oid in object_identifiers {
         if cursor.is_some_and(|cursor| oid.encode() <= cursor) {
@@ -79,34 +80,41 @@ pub fn handle_get_event_information(
                     })
                     .unwrap_or(0);
 
-                let event_priorities =
-                    ObjectIdentifier::new(ObjectType::NOTIFICATION_CLASS, notification_class)
-                        .ok()
-                        .and_then(|nc_oid| db.get(&nc_oid))
-                        .and_then(|nc_obj| {
-                            nc_obj
-                                .read_property(PropertyIdentifier::PRIORITY, None)
-                                .ok()
-                        })
-                        .and_then(|v| match v {
-                            PropertyValue::List(items) => {
-                                let [
-                                    PropertyValue::Unsigned(offnormal),
-                                    PropertyValue::Unsigned(fault),
-                                    PropertyValue::Unsigned(normal),
-                                ] = items.as_slice()
-                                else {
-                                    return None;
-                                };
-                                Some([
-                                    u32::try_from(*offnormal).ok()?,
-                                    u32::try_from(*fault).ok()?,
-                                    u32::try_from(*normal).ok()?,
-                                ])
-                            }
-                            _ => None,
-                        })
-                        .unwrap_or([0, 0, 0]);
+                let matching_notification_class =
+                    notification_class_identifiers.iter().find_map(|nc_oid| {
+                        let nc_obj = db.get(nc_oid)?;
+                        matches!(
+                            nc_obj.read_property(PropertyIdentifier::NOTIFICATION_CLASS, None),
+                            Ok(PropertyValue::Unsigned(class_number))
+                                if class_number == u64::from(notification_class)
+                        )
+                        .then_some(nc_obj)
+                    });
+                let event_priorities = matching_notification_class
+                    .and_then(|nc_obj| {
+                        nc_obj
+                            .read_property(PropertyIdentifier::PRIORITY, None)
+                            .ok()
+                    })
+                    .and_then(|v| match v {
+                        PropertyValue::List(items) => {
+                            let [
+                                PropertyValue::Unsigned(offnormal),
+                                PropertyValue::Unsigned(fault),
+                                PropertyValue::Unsigned(normal),
+                            ] = items.as_slice()
+                            else {
+                                return None;
+                            };
+                            Some([
+                                u32::try_from(*offnormal).ok()?,
+                                u32::try_from(*fault).ok()?,
+                                u32::try_from(*normal).ok()?,
+                            ])
+                        }
+                        _ => None,
+                    })
+                    .unwrap_or([0, 0, 0]);
 
                 let acked = object
                     .read_property(PropertyIdentifier::ACKED_TRANSITIONS, None)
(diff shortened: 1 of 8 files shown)


jscott3201 (commented):
[Runtime] NOT READY

Reviewed head: ae7628949363eb6d1d179aae04168b8d48d05bb3

### Medium

- `crates/bacnet-services/src/alarm_event/get_event_information.rs:11-12` materializes the generic bit-string decoder's owned `Vec` before checking that this fixed-width field has exactly one data octet. Because the helper is reached twice per summary, a valid ACK at the accepted 10,000-summary boundary performs 20,000 peer-driven allocate/copy/free cycles; a malformed transition field can likewise copy its entire declared content before `bits.len() != 1` rejects it. This creates avoidable allocator and latency pressure in the network decode path and regresses the formerly allocation-free transition parsing. Check the two content octets directly (or use a borrowed bit-string decode) before any allocation. The new boundary test exercises the count but does not detect this allocation failure mode.

The numeric range checks, exact tags/cardinalities, zero-padding check, strict Boolean, full consumption, and 10,000/10,001 result behavior otherwise match the requested scope. Residual risk: malformed-input coverage is matrix-based rather than fuzz/property-based.

jscott3201 (commented):
[Dataflow] NOT READY
Reviewed head: ae7628949363eb6d1d179aae04168b8d48d05bb3

### HIGH

1. **Server-generated ACKs expose `[0, 0, 0]` for every event priority.** `crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs:77-92` only accepts the notification class `PRIORITY` result as `PropertyValue::OctetString`; the in-tree `NotificationClass::read_property(PRIORITY, None)` returns a three-element `PropertyValue::List` (`crates/bacnet-objects/src/notification_class/mod.rs:99-106`). Thus a class configured with priorities `[4, 5, 6]` reaches `GetEventInformationAck::encode` as `[0, 0, 0]`, and the new public decoder correctly exposes the wrong values. This bypasses the requested value flow for all three Event Priorities.

2. **Pagination is not a stable object-identifier cursor.** The handler starts skipping only after it encounters an exact matching OID and then continues over `db.iter_objects()` (`crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs:25-31`); that iterator is over the unordered `HashMap` (`crates/bacnet-objects/src/database.rs:16-18,206-208`). If the last OID is removed between pages, `skipping` never clears and the response silently returns no remaining events. HashMap order also makes the page boundary dependent on insertion/rehash/server lifetime rather than OID order, so a 25-item response can omit or repeat active summaries across requests. The `More Events` bit is therefore not a reliable state-machine continuation signal.

### MEDIUM

3. **The server never forwards stored event timestamps.** The timestamp extraction is unconditional fallback: both the three-item `PropertyValue::List` arm and the catch-all return `None` (`crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs:105-116`), so every server-generated `EventSummary` is encoded with three `SequenceNumber(0)` values even when the object has nonzero history. The ACK codec remains wire-compatible, but the public decoded `EventSummary.event_timestamps` does not represent the server’s event state.

### Verified flows

- Empty request bytes decode to no cursor; a nonempty request must be context `[0]`, owns the complete OID payload, and has no trailing bytes (`crates/bacnet-services/src/alarm_event/get_event_information.rs:63-79`). The server passes those bytes directly into the handler (`crates/bacnet-server/src/server/requests/mod.rs:215-224`), which applies detection-enable exclusion, non-NORMAL filtering, and its 25-summary response cap (`get_event_information.rs:17,33-46`).
- ACK decoding enforces the Clause 21 order and exact context tags/constructed closings, three timestamps, and three application Unsigned priorities (`crates/bacnet-services/src/alarm_event/get_event_information.rs:109-221`). The shared helpers preserve padded `u32::MAX` and reject `u32::MAX + 1`/`u64::MAX` before assigning the public `u32` fields (`crates/bacnet-services/src/common.rs:38-50`; `get_event_information.rs:22-44`).
- Transition fields require one data octet, `unused_bits == 5`, and zero low padding; More Events accepts only context content `[0]` or `[1]`; both request and ACK reject trailing bytes (`get_event_information.rs:4-20`; `common.rs:52-80`; `get_event_information.rs:222-236`). The 10,000-item decode cap is checked before parsing another summary (`get_event_information.rs:117-127`).
- Existing ACK encoding uses the same context tags, application Unsigned priorities, shared timestamp-choice codec, and valid server defaults; the time, sequence-number, and DateTime alternatives remain covered by the existing timestamp tests. Audit #345, BACnetPropertyStates #346, and event-notification codecs were not considered.

No blocker found, but the two HIGH integration defects must be resolved before this end-to-end change is ready.

jscott3201 (commented):
[Dataflow] READY

Reviewed head: e6a036e43cb656b1aac7dcf05a29a43b31bad2b2

No material data-flow or state-transition findings remain.

Evidence reviewed:
- `crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs:20-31` sorts encoded object identifiers and applies an exclusive `<=` cursor boundary before lookup, so a deleted cursor still resumes correctly.
- `crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs:43-45` sets More Events only after the 25-entry eligible-summary limit is exceeded; `:77-103` projects NotificationClass `PropertyValue::List` priorities; `:120-143` projects the three object-API sequence timestamps.
- `crates/bacnet-services/src/alarm_event/get_event_information.rs:137-191` traces the fixed summary fields through strict decoding, timestamp choices, and application Unsigned priorities; `:207-219` decodes More Events and rejects trailing bytes; the encoder preserves the same timestamp/priority fields.
- `crates/bacnet-server/src/handlers/tests/device_event.rs:325-384` covers custom priorities, nonzero sequence timestamps, 25-entry pagination, More Events, and deletion of the first-page cursor.

The documented object-API flattening limitation for non-sequence timestamp alternatives remains confined to `crates/bacnet-objects/src/event/history.rs:28-35` and is not newly regressed here. PR CI reports the relevant test, clippy, rustfmt, file-size, and no-secret checks successful.

jscott3201 (commented):
[Runtime] NOT READY

Reviewed head: e6a036e43cb656b1aac7dcf05a29a43b31bad2b2

Findings, ordered by shipping impact:

1. **High — valid Notification Class references can still project the wrong priorities.** `crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs:82-112` treats the event object's `Notification_Class` number as the *object instance* and only reads `PRIORITY` from that OID. The repository permits a NotificationClass object's `Notification_Class` property to differ from its instance, and the notification-delivery resolver already falls back to locating a class by that property. Concrete case: NotificationClass object instance 5 has `Notification_Class = 7` and priorities `[100,150,200]`; an active event object references class 7. GetEventInformation probes object `(NotificationClass,7)`, finds none, and emits `[0,0,0]`. The new test uses instance 5 with class 5, so it does not exercise this valid configuration.

2. **Medium — an out-of-range flattened timestamp aborts the entire response.** `crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs:123-149` accepts arbitrary `PropertyValue::Unsigned(u64)` elements and constructs `BACnetTimeStamp::SequenceNumber` without enforcing its `0..=65535` wire constraint. A custom object returning `[Unsigned(65536), Unsigned(0), Unsigned(0)]` passes this projection, then `ack.encode(buf)?` fails when the shared timestamp encoder rejects 65536. The observable result is that one malformed object's property prevents any ACK for an otherwise valid GetEventInformation request. The new fixture itself stores `[u64; 3]`, but tests only in-range values.

Verified follow-up: `crates/bacnet-services/src/alarm_event/get_event_information.rs:10-16` now validates and unpacks the borrowed two-octet transition content directly; there is no per-transition success-path `Vec` allocation. The decoder tests also construct and parse the real 10,000-summary boundary and reject summary 10,001.

jscott3201 (commented):
[Runtime] READY

Reviewed head: 00b5c1ae8d00fe3862b979b9a9903331690ed29d

No material runtime findings remain.

Prior findings are resolved:

- `crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs:28,83-119` enumerates NotificationClass objects and matches the referenced class against each object's `NOTIFICATION_CLASS` property rather than assuming that class number equals object instance. `crates/bacnet-server/src/handlers/tests/device_event.rs:303-333` exercises class number 7 on object instance 5 and decodes the expected `[100, 150, 200]` priorities from the ACK.
- `crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs:131-160` checks all three flattened sequence values through `u16::try_from` before constructing timestamps. Thus 65535 remains encodable, while any array containing 65536 takes the all-zero fallback instead of reaching the fallible ACK encoder. `crates/bacnet-server/src/handlers/tests/device_event.rs:336-365` covers both cases through an encoded and decoded ACK.

Complete-diff runtime checks:

- `crates/bacnet-services/src/alarm_event/get_event_information.rs:10-16` validates and unpacks borrowed transition content directly, with no per-transition success-path allocation.
- The ACK decoder uses checked content bounds, strict tags/cardinalities/Boolean/trailing-data validation, and rejects item 10,001 before parsing it; `crates/bacnet-services/src/alarm_event/tests/get_event_information_decode.rs:298-315` constructs and decodes the real 10,000-summary boundary.
- `crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs:23-35` retains numeric OID sorting and the exclusive cursor boundary, and the decoded integration test still covers a removed first-page cursor.

Head CI completed successfully for tests, Clippy, rustfmt, file-size, and secret checks.

jscott3201 (commented):
[Dataflow] READY

Reviewed head: 00b5c1ae8d00fe3862b979b9a9903331690ed29d

No material data-flow regression found in the follow-up.

Evidence:
- `crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs:20-32` preserves request decoding, encoded-OID sorting, and the exclusive cursor boundary; `:37-45` preserves Event_Detection_Enable filtering, the 25-summary cap, and More Events transition.
- `crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs:82-115` scans Notification Class objects by their `NOTIFICATION_CLASS` property and projects the exact three-item object API `PropertyValue::List` into ACK priorities. The non-instance-matching case is covered at `crates/bacnet-server/src/handlers/tests/device_event.rs:303-326`.
- `crates/bacnet-server/src/handlers/alarm_event/get_event_information.rs:128-159` requires exactly three Unsigned timestamp values, constrains each to `0..=65535`, and defaults the complete malformed array locally. Valid `11`, `65535`, and invalid `65536` flows are exercised at `crates/bacnet-server/src/handlers/tests/device_event.rs:328-362`.
- The unchanged ACK encoder/public decoder path remains consistent for application Unsigned priorities and timestamp choices in `crates/bacnet-services/src/alarm_event/get_event_information.rs:222-274`; the pagination/deleted-cursor integration remains covered at `crates/bacnet-server/src/handlers/tests/device_event.rs:364-400`.

The documented #171 flattening of non-sequence timestamp alternatives remains unchanged and is not newly regressed. Static inspection used the complete base-to-head diff and head/base file reads; current CI showed clippy, rustfmt, file-size, and no-secret checks successful while the Ubuntu test job was still in progress.

Issue #351, opened 2026-08-15
Correct remaining NotificationParameters wire models


jscott3201:
## Problem

Several public `NotificationParameters` alternatives do not match the ANSI/ASHRAE 135-2020 Clause 21 `BACnetNotificationParameters` production. These are model corrections rather than framing or numeric-width fixes.

## Evidence

- `AccessEvent.access_credential` is modeled as `BACnetDeviceObjectPropertyReference`; Clause 21 requires `BACnetDeviceObjectReference` at `[4]`.
- `AccessEvent.authentication_factor` is mandatory in the Rust model and encoder; Clause 21 marks `[5] BACnetAuthenticationFactor` optional.
- `ChangeOfTimer.last_state_change`, `initial_timeout`, and `expiration_time` are mandatory in the Rust model and always encoded; Clause 21 marks `[3]`, `[4]`, and `[5]` optional.
- `NoneParams` emits choice `[20]`, but Clause 21 states that context tag `[20]` is intentionally omitted from `BACnetNotificationParameters`.
- The public model does not represent `complex-event-type [6] SEQUENCE OF BACnetPropertyValue`.

Local Standard anchors: extracted text lines 58642, 58681-58699, and 58733-58776. Current model and codecs are in `crates/bacnet-services/src/alarm_event/notification_parameters/`.

## Scope

- Correct the public variants and their encoder/decoder fields.
- Define the source-compatibility migration for changed field types and optionality.
- Add exact Clause 21 wire vectors for AccessEvent, ChangeOfTimer, complex event type, and the omitted choice `[20]`.
- Update all server producers and consumers before changing support claims.

Boundary ownership for the enclosing EventNotification `[12]` wrapper remains tracked in #349.

jscott3201:
Resolved by #432 (`dev` merge `2d59ca8edab6d56458ef9b3694c8a5ed621e1675`; reviewed head `3d51b5dcd251bc9ae6285e4c713ac34fed8117dd`).

The merged change corrects the AccessEvent credential type, makes its authentication factor optional, makes the ChangeOfTimer suffix fields optional, adds ComplexEventType, rejects omitted choice `[20]`, adds exact Clause 21 vectors, and records the source-compatibility changes in the changelog.

[...]

PR #352, opened 2026-08-15
fix(services): harden notification parameter boundaries

### LINKED ISSUE #349: Harden EventNotification framing for opaque notification parameters
## Problem

`EventNotificationRequest::decode` uses a depth scan to locate the outer `[12]` close for notification-parameter variants other than ChangeOfLifeSafety. Legacy variants such as Extended `[9]` and AccessEvent `[13]` preserve opaque payload bytes, so those bytes cannot safely be reparsed as BACnet tags. The scan can stop at the wrong boundary and can accept a missing or wrong outer close or trailing data.

This was found while reviewing #348. That PR hardens only ChangeOfLifeSafety `[8]`; broadening it would require changing boundary ownership across the notification-parameter family.

## Scope

- Make notification-parameter decoding report its exact consumed offset, or provide an equivalent unambiguous boundary contract.
- Require a matching outer closing `[12]` for every variant.
- Reject siblings after the selected variant and bytes after closing `[12]`.
- Preserve opaque payload bytes for Extended `[9]` and AccessEvent `[13]`.
- Avoid parsing opaque bytes as nested BACnet tags.

## Tests

- Missing, wrong, and truncated outer `[12]` closing tags for structured and opaque variants.
- Extra siblings before closing `[12]`.
- Trailing data after closing `[12]`.
- Opaque payloads containing bytes that resemble opening and closing tags.
- Existing round trips for every `NotificationParameters` variant.

jscott3201:
## Summary
- anchor the final EventNotification `[12]` close at the end of the request and require one matching NotificationParameters variant immediately before it
- traverse raw NotificationParameters fields as encoded BACnet TLV sequences, preserving tag-like content octets while rejecting malformed nesting, same-tag siblings, and truncated close aliases
- preserve the public decoder contract for callers that pass the enclosing event-values `[12]` close
- require exact context tags and four-bit StatusFlags across the affected variants
- replace 11 unchecked `u32` narrowings and one `u16` narrowing with exact context decoding and checked conversion
- validate ChangeOfTimer DateTime fields and exact variant consumption

## Scope

Closes #349 and advances #309. The direct unchecked-cast inventory falls from 38 to 26.

This PR preserves the public NotificationParameters variants and encoder wire shapes. Raw `Vec<u8>` fields now explicitly contain encoded BACnet TLV sequences; their values remain byte-preserving. The separate model migration for AccessEvent, ChangeOfTimer, complex event type `[6]`, and the omitted choice `[20]` is tracked in #351. BACnetPropertyStates remains in #346 and audit service models remain in #345.

## Validation
- `cargo test -p bacnet-server event_notification --locked`

### DIFF
--- a/crates/bacnet-services/src/alarm_event/event_notification.rs
+++ b/crates/bacnet-services/src/alarm_event/event_notification.rs
@@ -159,62 +159,35 @@ impl EventNotificationRequest {
         // [12] eventValues — optional
         let mut event_values = None;
         if offset < data.len() {
-            let (peek, _) = tags::decode_tag(data, offset)?;
-            if !peek.is_opening || peek.number != 12 {
+            let (opening, inner_start) = tags::decode_tag(data, offset)?;
+            if !opening.is_opening_tag(12) {
                 return Err(Error::decoding(
                     offset,
                     "EventNotification expected opening tag 12 for eventValues",
                 ));
             }
-            // Skip opening tag [12]
-            let (_, inner_start) = tags::decode_tag(data, offset)?;
-            event_values = Some(NotificationParameters::decode(data, inner_start)?);
-            let (variant, variant_start) = tags::decode_tag(data, inner_start)?;
-            if variant.number == 8 {
-                let (_, variant_end) =
-                    tags::extract_context_value(data, variant_start, variant.number)?;
-                let (closing, next) = tags::decode_tag(data, variant_end)?;
-                if !closing.is_closing_tag(12) {
-                    return Err(Error::decoding(
-                        variant_end,
-                        "EventNotification expected closing tag 12 after eventValues",
-                    ));
-                }
-                if next != data.len() {
-                    return Err(Error::decoding(
-                        next,
-                        "EventNotification unexpected trailing data",
-                    ));
-                }
-                offset = next;
-            } else {
-                // Legacy variants may contain opaque bytes that cannot be scanned as tags.
-                let mut scan = inner_start;
-                let mut depth: usize = 1;
-                while depth > 0 && scan < data.len() {
-                    let (tag, next) = tags::decode_tag(data, scan)?;
-                    if tag.is_opening {
-                        depth += 1;
-                        scan = next;
-                    } else if tag.is_closing {
-                        depth -= 1;
-                        if depth == 0 {
-                            offset = next;
-                        } else {
-                            scan = next;
-                        }
-                    } else {
-                        let end = next.saturating_add(tag.length as usize);
-                        if end > data.len() {
-                            return Err(Error::decoding(
-                                next,
-                                "EventNotification: truncated tag in eventValues",
-                            ));
-                        }
-                        scan = end;
-                    }
-                }
+            let closing_offset = data.len().checked_sub(1).ok_or_else(|| {
+                Error::decoding(offset, "EventNotification missing closing tag 12")
+            })?;
+            if closing_offset < inner_start {
+                return Err(Error::decoding(
+                    inner_start,
+                    "EventNotification empty eventValues",
+                ));
             }
+            let (closing, next) = tags::decode_tag(data, closing_offset)?;
+            if !closing.is_closing_tag(12) || next != data.len() {
+                return Err(Error::decoding(
+                    closing_offset,
+                    "EventNotification expected closing tag 12 after eventValues",
+                ));
+            }
+            event_values = Some(NotificationParameters::decode_bounded(
+                data,
+                inner_start,
+                closing_offset,
+            )?);
+            offset = next;
         }
         let _ = offset;
 
--- a/crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs
+++ b/crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs
@@ -1,29 +1,25 @@
 use super::decode_helpers::{
     closing_tag_start, decode_context_status_flags, decode_context_u16,
-    extract_trailing_raw_context,
+    extract_trailing_raw_context, finish_variant as check_variant_end,
 };
 use super::decode_timer::{decode_change_of_discrete_value, decode_change_of_timer};
 use super::*;
 use crate::common::{decode_context, decode_context_u32};
 
 impl NotificationParameters {
-    /// Decode notification parameters from a position just past the opening
-    /// tag of the eventValues wrapper.
+    /// Decode one notification-parameter choice ending at the end of `data`.
     pub fn decode(data: &[u8], offset: usize) -> Result<Self, Error> {
-        Self::decode_impl(data, offset, None)
+        Self::decode_impl(data, offset, data.len())
     }
 
     pub(crate) fn decode_bounded(data: &[u8], offset: usize, end: usize) -> Result<Self, Error> {
-        Self::decode_impl(data, offset, Some(end))
+        Self::decode_impl(data, offset, end)
     }
 
-    fn decode_impl(data: &[u8], offset: usize, end: Option<usize>) -> Result<Self, Error> {
-        let data = match end {
-            Some(end) => data.get(..end).ok_or_else(|| {
-                Error::decoding(end, "NotificationParameters boundary exceeds input")
-            })?,
-            None => data,
-        };
+    fn decode_impl(data: &[u8], offset: usize, end: usize) -> Result<Self, Error> {
+        let data = data
+            .get(..end)
+            .ok_or_else(|| Error::decoding(end, "NotificationParameters boundary exceeds input"))?;
         // Peek the inner opening tag to determine the variant
         if offset >= data.len() {
             return Err(Error::decoding(
@@ -39,16 +35,14 @@ impl NotificationParameters {
             ));
         }
         let variant_tag = inner_tag.number;
-        let variant_body_end = end
-            .map(|_| {
-                closing_tag_start(
-                    data,
-                    data.len(),
-                    variant_tag,
-                    "NotificationParameters variant",
-                )
-            })
-            .transpose()?;
+        let variant_body_end = closing_tag_start(
+            data,
+            data.len(),
+            variant_tag,
+            "NotificationParameters variant",
+        )?;
+        let finish_variant =
+            |value, consumed| check_variant_end(value, consumed, variant_body_end, variant_tag);
 
         match variant_tag {
             // [1] Change of state
@@ -77,10 +71,13 @@ impl NotificationParameters {
                     return Err(Error::decoding(sf_pos, "ChangeOfState: truncated flags"));
                 }
                 let status_flags = decode_status_flags(&data[sf_pos..sf_end]);
-                Ok(Self::ChangeOfState {
-                    new_state,
-                    status_flags,
-                })
+                finish_variant(
+                    Self::ChangeOfState {
+                        new_state,
+                        status_flags,
+                    },
+                    sf_end,
+                )
             }
             // [2] Change of value
             2 => {
@@ -131,10 +128,13 @@ impl NotificationParameters {
                     return Err(Error::decoding(sf_pos, "ChangeOfValue: truncated flags"));
                 }
                 let status_flags = decode_status_flags(&data[sf_pos..sf_end]);
-                Ok(Self::ChangeOfValue {
-                    new_value,
-                    status_flags,
-                })
+                finish_variant(
+                    Self::ChangeOfValue {
+                        new_value,
+                        status_flags,
+                    },
+                    sf_end,
+                )
             }
             // [5] Out of range
             5 => {
@@ -170,13 +170,15 @@ impl NotificationParameters {
                     return Err(Error::decoding(p, "OutOfRange: truncated exceeded_limit"));
                 }
                 let exceeded_limit = primitives::decode_real(&data[p..end])?;
-                let _ = (pos, end);
-                Ok(Self::OutOfRange {
-                    exceeding_value,
-                    status_flags,
-                    deadband,
-                    exceeded_limit,
-                })
+                finish_variant(
+                    Self::OutOfRange {
+                        exceeding_value,
+                        status_flags,
+                        deadband,
+                        exceeded_limit,
+                    },
+                    end,
+                )
             }
             // [10] Buffer ready
             10 => {
@@ -201,13 +203,16 @@ impl NotificationParameters {
                 let (previous_notification, pos) =
                     decode_context_u32(data, pos, 1, "BufferReady previous-notification")?;
                 // [2] current-notification
-                let (current_notification, _) =
+                let (current_notification, end) =
                     decode_context_u32(data, pos, 2, "BufferReady current-notification")?;
-                Ok(Self::BufferReady {
-                    buffer_property,
-                    previous_notification,
-                    current_notification,
-                })
+                finish_variant(
+                    Self::BufferReady {
+                        buffer_property,
+                        previous_notification,
+                        current_notification,
+                    },
+                    end,
+                )
             }
             // [11] Unsigned range
             11 => {
@@ -241,12 +246,14 @@ impl NotificationParameters {
                     ));
                 }
                 let exceeded_limit = primitives::decode_unsigned(&data[p..end])?;
-                let _ = (pos, end);
-                Ok(Self::UnsignedRange {
-                    exceeding_value,
-                    status_flags,
-                    exceeded_limit,
-                })
+                finish_variant(
+                    Self::UnsignedRange {
+                        exceeding_value,
+                        status_flags,
+                        exceeded_limit,
+                    },
+                    end,
+                )
             }
             // [0] Change of bitstring
             0 => {
@@ -269,10 +276,13 @@ impl NotificationParameters {
                     ));
                 }
                 let status_flags = decode_status_flags(&data[sf_pos..sf_end]);
-                Ok(Self::ChangeOfBitstring {
-                    referenced_bitstring: (unused, bits),
-                    status_flags,
-                })
+                finish_variant(
+                    Self::ChangeOfBitstring {
+                        referenced_bitstring: (unused, bits),
+                        status_flags,
+                    },
+                    sf_end,
+                )
             }
             // [3] Command failure
             3 => {
@@ -297,18 +307,21 @@ impl NotificationParameters {
                 if !t.is_opening || t.number != 2 {
                     return Err(Error::decoding(pos, "CommandFailure: expected opening [2]"));
                 }
-                let (feedback_value, _after) = extract_trailing_raw_context(
+                let (feedback_value, after) = extract_trailing_raw_context(
                     data,
                     p,
                     2,
                     variant_body_end,
                     "CommandFailure feedback-value",
                 )?;
-                Ok(Self::CommandFailure {
-                    command_value,
-                    status_flags,
-                    feedback_value,
-                })
+                finish_variant(
+                    Self::CommandFailure {
+                        command_value,
+                        status_flags,
+                        feedback_value,
+                    },
+                    after,
+                )
             }
             // [4] Floating limit
             4 => {
@@ -350,13 +363,15 @@ impl NotificationParameters {
                     return Err(Error::decoding(p, "FloatingLimit: truncated error_limit"));
                 }
                 let error_limit = primitives::decode_real(&data[p..end])?;
-                let _ = (pos, end);
-                Ok(Self::FloatingLimit {
-                    reference_value,
-                    status_flags,
-                    setpoint_value,
-                    error_limit,
-                })
+                finish_variant(
+                    Self::FloatingLimit {
+                        reference_value,
+                        status_flags,
+                        setpoint_value,
+                        error_limit,
+                    },
+                    end,
+                )
             }
             // [8] Change of life safety
             8 => {
@@ -381,19 +396,15 @@ impl NotificationParameters {
                 let status_flags = bits >> 4;
                 let (operation_expected, pos) =
                     decode_context_u32(data, pos, 3, "ChangeOfLifeSafety operation-expected")?;
-                let (closing, _) = tags::decode_tag(data, pos)?;
-                if !closing.is_closing_tag(8) {
-                    return Err(Error::decoding(
-                        pos,
-                        "ChangeOfLifeSafety expected closing tag 8",
-                    ));
-                }
-                Ok(Self::ChangeOfLifeSafety {
-                    new_state,
-                    new_mode,
-                    status_flags,
-                    operation_expected,
-                })
+                finish_variant(
+                    Self::ChangeOfLifeSafety {
+                        new_state,
+                        new_mode,
+                        status_flags,
+                        operation_expected,
+                    },
+                    pos,
+                )
             }
             // [9] Extended
             9 => {
@@ -408,18 +419,21 @@ impl NotificationParameters {
                 if !t.is_opening_tag(2) {
                     return Err(Error::decoding(pos, "Extended: expected opening [2]"));
                 }
-                let (parameters, _after) = extract_trailing_raw_context(
+                let (parameters, after) = extract_trailing_raw_context(
                     data,
                     p,
                     2,
                     variant_body_end,
                     "Extended parameters",
                 )?;
-                Ok(Self::Extended {
-                    vendor_id,
-                    extended_event_type,
-                    parameters,
-                })
+                finish_variant(
+                    Self::Extended {
+                        vendor_id,
+                        extended_event_type,
+                        parameters,
+                    },
+                    after,
+                )
             }
             // [13] Access event
             13 => {
@@ -467,21 +481,24 @@ impl NotificationParameters {
                         "AccessEvent: expected opening [5] for authentication-factor",
                     ));
                 }
-                let (authentication_factor, _after) = extract_trailing_raw_context(
+                let (authentication_factor, after) = extract_trailing_raw_context(
                     data,
                     p,
                     5,
                     variant_body_end,
                     "AccessEvent authentication-factor",
                 )?;
-                Ok(Self::AccessEvent {
-                    access_event,
-                    status_flags,
-                    access_event_tag,
-                    access_event_time,
-                    access_credential,
-                    authentication_factor,
-                })
+                finish_variant(
+                    Self::AccessEvent {
+                        access_event,
+                        status_flags,
+                        access_event_tag,
+                        access_event_time,
+                        access_credential,
+                        authentication_factor,
+                    },
+                    after,
+                )
             }
             // [14] Double out of range
             14 => {
@@ -523,13 +540,15 @@ impl NotificationParameters {
                     ));
                 }
                 let exceeded_limit = primitives::decode_double(&data[p..end])?;
-                let _ = (pos, end);
-                Ok(Self::DoubleOutOfRange {
-                    exceeding_value,
-                    status_flags,
-                    deadband,
-                    exceeded_limit,
-                })
+                finish_variant(
+                    Self::DoubleOutOfRange {
+                        exceeding_value,
+                        status_flags,
+                        deadband,
+                        exceeded_limit,
+                    },
+                    end,
+                )
             }
             // [15] Signed out of range
             15 => {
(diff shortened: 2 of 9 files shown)


jscott3201 (commented):
[Runtime]
Reviewed head: 6cf6584ab3733d9057601b9f8491b4a71be02540

NOT READY

- **High — a second same-tag NotificationParameters sibling is accepted.** `decode_impl` selects the variant from the first opening tag but only verifies that the final bytes before outer `[12]` are a matching variant close (`crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:41-51`). The BufferReady path then discards the consumed offset after `current-notification` and returns (`crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:203-206`). Concretely, place two complete `[10] BufferReady ... [/10]` values inside one `[12] ... [/12]`: the first opening selects variant 10, the second `[/10]` satisfies the suffix check, and decoding succeeds with the first value while silently ignoring the sibling. This violates the change's exact-one-variant/reject-siblings boundary contract; the hostile sibling coverage uses a different tag, so it does not exercise this same-tag case.

jscott3201 (commented):
[Adversary] NOT READY

Reviewed head: 6cf6584ab3733d9057601b9f8491b4a71be02540

Findings:

1. **[P1] The bounded decoder still accepts a same-tag sibling and other bytes after the selected structured variant.** `crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:42-51` anchors `variant_body_end` at the *last* close matching the first variant, but the BufferReady path at `crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:201-210` discards the consumed offset after `[2]` and returns without requiring it to equal that boundary; `crates/bacnet-services/src/alarm_event/event_notification.rs:185-189` consequently treats the decode as complete. Concrete input: encode a valid EventNotification containing BufferReady, remove only the final outer `[12]` close, append a second complete BufferReady encoding (including its `[10]` close), then restore the `[12]` close. The final `[10]`/`[12]` suffix passes, while the first BufferReady is decoded and its close plus the second sibling are silently ignored. Inserting an extra primitive before the original `[10]` close is accepted for the same reason. This directly defeats the claimed one-variant, sibling, fake-final, and trailing-byte checks.

2. **[P2] The public decoder retains a silent opaque-payload truncation fallback.** `NotificationParameters::decode` passes no end boundary at `crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:12-18`, and `crates/bacnet-services/src/alarm_event/notification_parameters/decode_helpers.rs:31-33` then falls back to delimiter scanning. Concrete input: encode `Extended { parameters: vec![0x61, 0x2f], .. }`, where `0x61,0x2f` is an application OctetString of length one whose content byte equals closing `[2]`. Calling the public `NotificationParameters::decode(encoded, 0)` treats that content byte as the field close and returns `Ok(Extended { parameters: vec![0x61], .. })`, silently corrupting the payload and leaving the actual closes unchecked. Thus delimiter safety applies only to the private bounded EventNotification path, despite the public decode entry point remaining available.

Coverage confirmation: the new overflow and leading-zero tests do exercise all 11 changed `u32` narrowing sites—four BufferReady/reference fields, Extended, two AccessEvent fields, ChangeOfReliability, and three ChangeOfTimer fields—and the one Extended `u16` site. I kept #351, #346, and #345 separate from these findings.

jscott3201 (commented):
[Dataflow] NOT READY
Reviewed head: 6cf6584ab3733d9057601b9f8491b4a71be02540

Findings:

- [P1] Enforce consumption through the variant close for every choice — `crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:38-47,170-178`. The bounded path computes the final variant-close offset, but most branches never compare their final parsed offset with it. For example, BufferReady discards the offset after `[2] current-notification` and returns. A valid `[10]` body followed by any extra context-tagged bytes (e.g. `[13]`) and then the `[10]` close is therefore accepted by EventNotification and the extra bytes are silently dropped. This leaves the selected variant boundary unenforced despite the outer suffix check.

- [P1] Pass the bounded end into ChangeOfDiscreteValue and reject a missing `[1]` field — `crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:735-738`; `crates/bacnet-services/src/alarm_event/notification_parameters/decode_timer.rs:124-133`. The `[21]` call drops `variant_body_end`. With bytes for opening `[21]`, opening/closing `[0]`, then the extended closing `[21]` (`FF 15`) and no status-flags field, `decode_tag` reads that two-byte variant close as `sf_tag`; its zero length makes `decode_status_flags(&[])` return `0`, and the decoder returns success. The final-close precheck does not prevent this because it is not used to constrain this decoder, so a malformed extended-tag variant is accepted and its required field is fabricated."

jscott3201 (commented):
[Luna] NOT READY
Reviewed head: 6cf6584ab3733d9057601b9f8491b4a71be02540

The shared EventNotificationRequest codec is appropriate for Confirmed and Unconfirmed productions, and the bounded trailing-raw helper avoids walking those opaque bodies. The checked numeric conversions also cover the claimed 11 u32 and one u16 narrowings. One framing defect remains:

- **High — the bounded decoder can accept a BACnetNotificationParameters choice without an actual matching close.** `crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:41-50` validates only that the final bytes before the outer close look like the variant close. But the BufferReady branch parses its final field against the full bounded slice and discards the returned end at `crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:203-208`; it never checks that end against `variant_body_end`. For a valid BufferReady request, encode `[2] current-notification` with length 2 and content `[0x00, 0xAF]`, omit the actual choice `[10]` close, and retain the outer `[12]` close `0xCF`. The suffix check interprets the content byte `0xAF` as the `[10]` close, while the field decoder consumes it as numeric content (175), so the malformed request is accepted. A real `[10]` close with an extra field before it is likewise ignored. This violates the Clause 21 choice/SEQUENCE boundary and the stated “one matching choice close” contract for both service productions.

The parser needs an invariant that structured fields cannot consume the validated choice-close suffix and that parsing reaches that suffix; otherwise the final-octet anchor is not a trustworthy delimiter.

jscott3201 (commented):
[Runtime]
Reviewed head: 033596948d61f8ad664e72411b3e30c010d83d26
Decision: NOT READY

**HIGH — Same-tag siblings are still accepted for choices whose last field is raw.** `extract_trailing_raw_context` chooses the field close immediately before the *final* variant close and then returns `variant_body_end` unconditionally (`crates/bacnet-services/src/alarm_event/notification_parameters/decode_helpers.rs:39-53`); the exact-consumption check therefore succeeds (`decode_helpers.rs:4-17`, used for Extended at `notification_parameters/decode.rs:419-429`). Concrete sequence: encode two `NotificationParameters::Extended` values consecutively and call `NotificationParameters::decode(&bytes, 0)`, or place both inside one EventNotification `[12]`. The decoder selects the second `[9]` close as the first choice's boundary and returns one `Extended` whose `parameters` absorb the first `[2]/[9]` closes, the second `[9]` open/header, and its body. The sibling is accepted and data is corrupted. The new regression uses only structured `BufferReady` (`tests/notification_parameters_boundaries.rs:397-411`), so it cannot exercise this bypass. Establish/reject the selected variant boundary independently of its trailing opaque field, and cover same-tag siblings for Extended, CommandFailure, AccessEvent, and ChangeOfReliability.

**MEDIUM — Non-trailing raw fields still reject valid close-like payload bytes, including the crate's own encoding.** `decode_change_of_discrete_value` still calls the byte-scanning `extract_raw_context` (`notification_parameters/decode_timer.rs:104`), whose loop treats every payload `0x0f` as a `[0]` close (`alarm_event/property_states.rs:238-249`), while the encoder emits `new_value` verbatim (`notification_parameters/encode.rs:285-297`). For `ChangeOfDiscreteValue { new_value: vec![0x61, 0x0f], ... }`—an application Octet String of length 1 containing `0x0f`—`encode` followed by `decode` errors because the content byte is mistaken for the field close and the actual close is parsed as status flags. CommandFailure `command_value` and ChangeOfStatusFlags `present_value` use the same scanner. Add delimiter-bearing round trips for these non-trailing raw fields and use a grammar/boundary-aware extractor.

Validation observed at this exact head: CI command `cargo test --workspace --exclude rusty-bacnet --locked --features bacnet-types/serde,bacnet-transport/ipv6,bacnet-transport/sc-tls,bacnet-transport/serial,bacnet-transport/ethernet` succeeded (`Test (ubuntu-latest)`); the cases above are absent from the suite.

jscott3201 (commented):
[Adversary]
Reviewed head: 033596948d61f8ad664e72411b3e30c010d83d26

Decision: NOT READY

## HIGH — Opaque suffix bytes can impersonate all required closes, so truncated input and same-tag siblings are silently accepted

`extract_trailing_raw_context` derives the raw-field close solely from the byte immediately before the final variant close and then treats every preceding byte as payload (`crates/bacnet-services/src/alarm_event/notification_parameters/decode_helpers.rs:39-53`). `Extended` uses that helper at `notification_parameters/decode.rs:419-438`, while `EventNotificationRequest::decode` similarly treats the final byte as the outer `[12]` close at `event_notification.rs:169-189`. These independent suffix assumptions compose into an ambiguous delimiter contract.

Concrete truncation: encode an EventNotification containing `Extended { vendor_id: 42, extended_event_type: 7, parameters: vec![0x61, 0x2f, 0x9f, 0xcf] }`. Its tail is:

`... 61 2f 9f cf  2f 9f cf`

where the final `2f 9f cf` are the real `[2]`, `[9]`, and `[12]` closes. Truncate those final three bytes. The head decoder still returns `Ok`, interpreting the three payload bytes as those closes and silently changing `parameters` from `[61,2f,9f,cf]` to `[61]`. This directly defeats the claimed missing-close/truncated-suffix rejection.

The prior same-tag-sibling concern also remains for raw-final variants. Two empty Extended choices concatenate as:

`9e 09 01 19 01 2e 2f 9f  9e 09 02 19 02 2e 2f 9f`

`NotificationParameters::decode(..., 0)` returns one `Extended` instead of rejecting the sibling; its `parameters` becomes `2f 9f 9e 09 02 19 02 2e`. Under EventNotification `[12]`, the same merge occurs. The new sibling test only exercises structured `BufferReady` (`notification_parameters_boundaries.rs:394-405`), where `finish_variant` can see the earlier consumed offset.

Repro test using the existing `encode_event` helper: truncate `encoded.len() - 3` for the first vector, and for the sibling vector remove outer `cf`, append a second `Extended::encode`, then restore `cf`; both should assert `is_err()`. Run with:

`cargo test -p bacnet-services --locked opaque_suffix_alias -- --nocapture`

Boundary handling must reject ambiguous suffix candidates (including raw bytes that can form field/variant/outer closes) or parse/model the trailing values sufficiently to establish one structural boundary.

## MEDIUM — The public decoder’s documented enclosing-buffer use is broken without a migration path

At the base SHA, `NotificationParameters::decode` was explicitly documented to accept an offset just after EventNotification’s opening `[12]` (`notification_parameters/decode.rs:6-9` at `c1ac43261827f665fbdee8758aefdfd55ff47fe9`). Head now unconditionally treats `data.len()` as the variant boundary (`notification_parameters/decode.rs:10-16`). The private EventNotification caller supplies its own bound, but external callers of this publicly re-exported API do not have access to `decode_bounded`.

Exact regression vector:

`ce 9e 09 01 19 01 2e 2f 9f cf`

This is outer open `[12]`, an empty Extended choice, and outer close `[12]`. `NotificationParameters::decode(&bytes, 1)` succeeded under the prior documented contract; head rejects it because the final `cf` is not closing `[9]`. The new public test only passes a standalone choice ending at the slice end (`notification_parameters_boundaries.rs:429-441`), so it misses this behavior change. A focused regression test can use that literal array and run with `cargo test -p bacnet-services --locked public_decode_enclosing_buffer -- --nocapture`.

Keep the bounded security path, but expose a public boundary/consumed-offset contract (with migration/deprecation) rather than silently repurposing the existing public method.

## Residual risks

- Raw `Vec<u8>` bodies still intentionally bypass malformed TLV-length and nested-close validation; until their wire models are corrected, hostile bodies remain structurally ambiguous.
- The separate AccessEvent/ChangeOfTimer/choice-model mismatches tracked by #351 were treated as out of scope.
- Runtime test execution was unavailable here; the vectors above are derived against the immutable GitHub head and existing test helpers.

jscott3201 (commented):
[Dataflow] NOT READY
Reviewed head: 033596948d61f8ad664e72411b3e30c010d83d26

Findings (severity-ranked):

1. [P1] Same-tag siblings are swallowed by trailing opaque fields. `decode_impl` derives `variant_body_end` from the final selected-variant close (`crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:35-43`), while `extract_trailing_raw_context` derives the raw field close from that suffix and reports `variant_body_end` as consumed (`crates/bacnet-services/src/alarm_event/notification_parameters/decode_helpers.rs:20-38,41-56`). For two valid same-tag `Extended [9]` choices before the outer `[12]` close, the suffix is the second choice's `[2]` close and `[9]` close; the first choice's `parameters` therefore absorbs the first close plus the second choice and the decoder returns success. The same construction applies to `CommandFailure [3]`, `AccessEvent [13]`, and `ChangeOfReliability [19]` (`decode.rs:307-327,429-457`). This violates the claimed exact variant ownership and silently changes the returned opaque payload. Add a framing rule/length or reject this ambiguity; test same-tag siblings for every trailing-opaque variant.

2. [P1] Non-trailing opaque fields still byte-scan delimiter bytes, so valid encoder output can fail to decode. `extract_raw_context` scans for raw opening/closing octets (`crates/bacnet-services/src/alarm_event/property_states.rs:227-257`), and ChangeOfDiscreteValue still invokes it for `[0] new-value` before parsing `[1]` (`crates/bacnet-services/src/alarm_event/notification_parameters/decode_timer.rs:90-109`). Encoding `new_value` containing `0x0f` (the closing marker for context `[0]`) makes the scanner terminate at the payload byte instead of the actual close, after which status-flags parsing fails. The same defect remains for `CommandFailure.command_value [0]` and `ChangeOfStatusFlags.present_value [0]`. This contradicts opaque-byte preservation for encoder-compatible values; use a boundary derived from the known following fields or explicitly constrain these raw values, with round-trip tests containing opening/closing marker bytes.

3. [P2] The public decoder's framing contract regressed incompatibly. `NotificationParameters::decode` now always calls `decode_impl(..., data.len())` and requires the selected variant to end at the end of the supplied slice (`crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:11-24,35-43`). The prior public contract accepted an offset just after the enclosing EventNotification `[12]` opening (base `.../decode.rs:5-8`) and ignored the enclosing close; a caller passing the documented wrapper slice for a normal variant now sees the final `[12]` close instead of the variant close and gets an error. The internal path avoids this only because EventNotification switched to the crate-private bounded path (`crates/bacnet-services/src/alarm_event/event_notification.rs:185-189`). Preserve the old public boundary behavior or provide a compatible public bounded/consumed-offset API and tests.

4. [P2] Exact body consumption does not establish field ownership for several variants. For example, the [5] OutOfRange branch reads each `Tag` and uses only its length before decoding the value; it does not require [0], [1], [2], and [3] (`crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:130-169`). Replacing all four with valid values under wrong context tags still reaches `finish_variant` and is accepted. The same pattern remains in the [0], [1], [2], [4], [11], [14]-[18] primitive branches. Apply expected-tag/class checks consistently or narrow the scope/contract; otherwise exact cursor completion can accept misowned fields.

Residual risks: suffix framing preserves delimiter-like bytes inside trailing opaque payloads, but without an explicit length/type boundary it cannot distinguish a same-tag sibling or a missing close whose bytes form the same suffix. Existing GitHub checks reported Test (ubuntu-latest), Clippy, Rustfmt, no-secret scan, and file-size cap successful; no independent local focused test was run.

jscott3201 (commented):
[Luna]
Reviewed head: 033596948d61f8ad664e72411b3e30c010d83d26
NOT READY

Findings (severity-ranked)

1. HIGH — same-choice siblings can still be swallowed by trailing opaque fields. A malformed request containing two consecutive CommandFailure choices with the same outer tag (and similarly Extended, AccessEvent, or ChangeOfReliability) can be accepted: `<12> <3> ... <2>payload</2></3> <3> ... <2>payload</2></3> </12>`. `event_notification.rs:160-191` removes only the final outer close and calls the bounded decoder. `notification_parameters/decode.rs:38-45` treats the final `</3>` as the selected variant close, so `variant_body_end` is after the second choice. `notification_parameters/decode_helpers.rs:21-41` then accepts the suffix close, and `:44-60` finds the second choice's `</2>` and copies the intervening second-choice bytes into the first `feedback_value`; `:4-19` sees `consumed == body_end` and returns `Ok`. This contradicts the PR's sibling/trailing-data contract and corrupts the decoded opaque value. Reject ambiguous/multiple same-tag choice boundaries rather than allowing the trailing-raw helper to absorb another choice.

2. HIGH — the public `NotificationParameters::decode` framing contract regresses. At base, `notification_parameters/decode.rs:6-8` documents decoding from the position after the EventNotification eventValues opening, where the enclosing `</12>` remains in the input; the old branches returned after decoding their choice and ignored that enclosing suffix. At head, the public method is still exposed at `notification_parameters/decode.rs:9-16`, but `:38-45` requires the input to end with the selected variant's own closing tag. Thus a previously accepted `params.encode()` followed by `tags::encode_closing_tag(..., 12)` now fails, while external callers have no bounded/end parameter. The internal `decode_bounded` workaround does not preserve the re-exported public API contract. Preserve the legacy public framing or expose an explicit boundary/consumed API without silently changing existing callers.

3. MEDIUM — delimiter-like bytes in non-trailing opaque fields still hit the legacy byte scanner. `notification_parameters/property_states.rs:225-256` treats every payload byte equal to the field's opening/closing octet as structural. `notification_parameters/decode_timer.rs:90-120` still uses it for ChangeOfDiscreteValue's raw `[0] new-value`; the same pattern remains for CommandFailure's `[0] command-value` and ChangeOfStatusFlags' `[0] present-value`. A valid raw value containing `0x0f` for `[0]` is mistaken for `</0>`, so the subsequent status-flags field is rejected (or the boundary is mis-owned). The new suffix ownership fixes only trailing raw fields and does not close this remaining public/legacy scan boundary. Apply an unambiguous ownership strategy to the leading raw fields or explicitly scope/document the compatibility limitation and add hostile-byte coverage.

Gaps
- No hostile test covers two same-tag siblings for any of the four trailing-raw variants.
- No test preserves the base public `NotificationParameters::decode(data, offset)` behavior when an enclosing EventNotification `</12>` follows the choice.
- No test puts delimiter-like bytes in CommandFailure command-value, ChangeOfDiscreteValue new-value, or ChangeOfStatusFlags present-value.

Non-blocking follow-ups
- Keep the exact-choice boundary rules centralized and document which decoder owns each enclosing close.
- Add wire-level fixtures for extended variant closing tags [15+] alongside the existing round trips.

Evidence/sources
- GitHub PR #352 `get` and `get_diff`, verified base `c1ac43261827f665fbdee8758aefdfd55ff47fe9` and head `033596948d61f8ad664e72411b3e30c010d83d26`.
- Head/base reads of the changed notification decoder, event decoder, property-state helper, timer decoder, encoder/tests, and `crates/bacnet-encoding/src/tags.rs` (including `extract_raw_context` semantics).
- No local commands or repository modifications were used.

jscott3201 (commented):
[Runtime]
Reviewed head: 01f4a73aa78e07b34382682013269ea08b513cec
Decision: READY

No material findings.

Prior Runtime findings are resolved:
- `extract_raw_context` now performs bounded TLV traversal with a fixed-depth stack, matching-close validation, checked content arithmetic, and length-aware content skipping (`crates/bacnet-services/src/alarm_event/property_states.rs:223-285`). Thus an earlier raw-field close is returned before a same-tag sibling, and delimiter-like octets inside a length-delimited value are not interpreted as tags. Regressions cover all trailing-raw sibling variants and all non-trailing raw fields (`tests/notification_parameters_boundaries.rs:450-513`).
- The public decoder again recognizes an enclosing final `[12]` close while standalone choices remain exact (`notification_parameters/decode.rs:10-18`), and EventNotification retains the explicit final-close bound (`alarm_event/event_notification.rs:160-190`).
- The expected-tag refactor routes primitive fields through exact context extraction and strict four-bit StatusFlags decoding (`notification_parameters/decode_helpers.rs:38-78`); each variant still checks its consumed cursor against the selected variant boundary.

Validation observed at this exact head: GitHub `Test (ubuntu-latest)` succeeded with `cargo test --workspace --exclude rusty-bacnet --locked --features bacnet-types/serde,bacnet-transport/ipv6,bacnet-transport/sc-tls,bacnet-transport/serial,bacnet-transport/ethernet`; Clippy, Rustfmt, file-size, and no-secret checks also succeeded.

Residual test risk: the new local TLV walker has strong delimiter/sibling coverage but no direct boundary test for exactly 32 versus 33 nested context tags or a mismatched inner close. The fixed stack and pre-write depth check make those paths non-panicking on inspection.

jscott3201 (commented):
[Adversary]
Reviewed head: 01f4a73aa78e07b34382682013269ea08b513cec

Decision: READY

No material findings.

The prior vectors are addressed at this head:

- **Opaque suffix truncation alias:** `extract_raw_context` now parses TLVs, skips primitive content by its checked declared length, tracks nested context tags, and requires matching closes (`crates/bacnet-services/src/alarm_event/property_states.rs:219-282`). For an Extended octet string ending in `2f 9f cf`, the tail is `63 2f 9f cf 2f 9f cf`; deleting the real final `[2]/[9]/[12]` closes leaves the declared three-octet value truncated after EventNotification removes the apparent final `cf`, so decoding errors rather than returning a shortened value.
- **Same-tag siblings / extra members:** the raw walker stops at the first structural field close, then every choice compares its consumed offset with the final variant boundary through `finish_variant` (`notification_parameters/decode_helpers.rs:4-17`; Extended call at `notification_parameters/decode.rs:375-386`). A second `[9] Extended` therefore leaves the first decode short of `variant_body_end` and is rejected.
- **Public wrapper compatibility:** `NotificationParameters::decode` recognizes a final enclosing `[12]` close while retaining standalone-choice decoding (`notification_parameters/decode.rs:10-14`). The literal wrapped shape `ce 9e 09 2a 19 07 2e 61 cf 2f 9f cf` preserves the `cf` octet inside the length-delimited value and decodes successfully.
- **Hostile TLVs:** the raw walker checks stack capacity before writes, enforces LIFO tag-number matching, delegates extended-tag/extended-length validation to `decode_tag`, uses checked length addition, bounds content against the slice, special-cases application Boolean’s header-carried value, and returns errors for EOF/unclosed nesting (`property_states.rs:228-282`). I found no panic path in those transitions. Structured fields also now require their owned context tags and exact four-bit StatusFlags.

Reproducible focused commands:

- `cargo test -p bacnet-services --locked raw_fields_reject_same_tag_siblings_and_truncated_close_aliases`
- `cargo test -p bacnet-services --locked non_trailing_raw_fields_preserve_delimiters_inside_values`
- `cargo test -p bacnet-services --locked public_notification_parameter_decode_accepts_event_values_close`
- `cargo test -p bacnet-services --locked notification_parameters_boundaries`

Residual risks:

- Raw fields are validated for structural TLV framing, not each field’s semantic BACnet datatype/cardinality; callers remain responsible for the newly documented raw-`Vec` contract.
- The AccessEvent/ChangeOfTimer/choice model corrections tracked in #351 and BACnetPropertyStates work tracked in #346 remain outside this target.
- Nested/mismatched/extended/depth-limit behavior is clear from the bounded iterative parser, but dedicated unit vectors for every one of those branches would further reduce regression risk.

jscott3201 (commented):
[Luna]
Reviewed head: 01f4a73aa78e07b34382682013269ea08b513cec
READY

No findings. The prior Luna blockers are resolved at this head:
- `crates/bacnet-services/src/alarm_event/property_states.rs:225-282` now traverses raw fields by decoded TLV lengths with a nesting stack, so delimiter-like content is preserved while mismatched/truncated nesting is rejected.
- `crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:11-18,40-47` preserves the public optional enclosing `[12]` close contract, while `event_notification.rs:160-191` supplies the strict bounded request boundary.
- Every affected structured branch routes its final cursor through `finish_variant` (`decode_helpers.rs:4-19`); the timer/discrete decoders also compare their cursor with `variant_body_end` (`decode_timer.rs:56-89,92-120`). This rejects same-choice siblings and trailing bytes rather than allowing them to be absorbed.
- The public model contract is now explicit in `notification_parameters/mod.rs` for raw `Vec<u8>` fields, and the encoder/tests use encoded BACnet TLV values (`notification_parameters/encode.rs` raw-field branches; `tests/notification_parameters_boundaries.rs` tests `raw_fields_reject_same_tag_siblings_and_truncated_close_aliases`, `non_trailing_raw_fields_preserve_delimiters_inside_values`, `public_notification_parameter_decode_accepts_event_values_close`, and `raw_fields_require_encoded_bacnet_values`). This is coherent with the chosen scope: malformed raw sequences may be rejected, while valid TLV content remains byte-preserving.

Gaps
- The raw `Vec<u8>` precondition is documented but not represented by a typed public wrapper; callers can still construct invalid bytes and receive an encode result that only fails on later decode.
- The separate model migrations called out in the PR description remain outside this review scope.

Non-blocking follow-ups
- Consider a validated `EncodedValue`/TLV-sequence type in a future API cleanup to make the raw-field contract enforceable at construction.
- Keep the focused malformed-nesting and enclosing-buffer fixtures when future decoder refactors move these helpers.

Evidence/sources
- GitHub PR #352 `get` and `get_diff`, with verified base `c1ac43261827f665fbdee8758aefdfd55ff47fe9` and reviewed head `01f4a73aa78e07b34382682013269ea08b513cec`.
- Head reads of the changed event decoder, notification decoder/helpers/timer decoder, public models, encoder, direct callers, tests, and BACnet tag decoder semantics.
- No local commands or repository modifications were used.

jscott3201 (commented):
[Dataflow] READY
Reviewed head: 01f4a73aa78e07b34382682013269ea08b513cec

No material findings.

Evidence:
- Target verification: PR #352 reports base `c1ac43261827f665fbdee8758aefdfd55ff47fe9` and head `01f4a73aa78e07b34382682013269ea08b513cec`; reviewed exactly `c1ac432...01f4a73` via the GitHub PR diff.
- EventNotification now owns the final `[12]` close and passes the exact inner bound (`crates/bacnet-services/src/alarm_event/event_notification.rs:159-191`). The public decoder recognizes an optional enclosing `[12]`, while the bounded path remains explicit (`crates/bacnet-services/src/alarm_event/notification_parameters/decode.rs:11-49`).
- `finish_variant` compares every structured cursor with the validated variant-body boundary (`crates/bacnet-services/src/alarm_event/notification_parameters/decode_helpers.rs:4-17`); ChangeOfDiscreteValue and ChangeOfTimer also enforce their final cursor (`decode_timer.rs:51-91`).
- Raw fields are now traversed as encoded BACnet TLV sequences, preserving content octets and rejecting malformed nesting, same-tag siblings, and close aliases (`crates/bacnet-services/src/alarm_event/property_states.rs:227-276`). The new boundary tests cover all trailing-raw variants, non-trailing delimiter bytes, public enclosing `[12]`, truncation, and same-tag siblings (`crates/bacnet-services/src/alarm_event/tests/notification_parameters_boundaries.rs`).
- Direct primitive fields now use exact context-tag decoding and strict four-bit StatusFlags across the affected variants (`notification_parameters/decode.rs:67-709`, `decode_helpers.rs:40-83`); the prior exact-tag finding is resolved. The remaining BACnetPropertyStates model permissiveness is explicitly tracked outside this change in #346.

Validation evidence supplied with the target: 440 service tests, server event tests, feature-enabled workspace tests, workspace Clippy, fmt, file-size, and secret scan passed.

Residual risks: raw `Vec<u8>` fields now intentionally require encoded BACnet TLV sequences; callers providing legacy arbitrary bytes may receive decode errors. BACnetPropertyStates and the separately tracked model migrations remain outside this PR.

PR #356, opened 2026-08-17
fix(types): restore no_std compilation; drop orphaned binary tests module

### LINKED ISSUE #263: binary/tests/mod.rs is an orphaned module file that nothing declares
### Affected revision

`dev` at `415eb32`. Found by the Event_Time_Stamps family grounding pass (2026-07-30).

### Behavior

`crates/bacnet-objects/src/binary/tests/mod.rs` declares `mod command_failure; mod input_output; mod value;` — but nothing declares *it*. `binary/mod.rs` has no `mod tests;`; every binary test file is attached directly via `#[path]` instead (`binary/input.rs:297`, `binary/output.rs:322`, `binary/value.rs:337`, `binary/mod.rs:23` for `generic_event_properties.rs`). The file is dead code the compiler never sees, and wiring it up naively would double-declare all three modules.

Contrast `multistate/mod.rs:71` and `analog/mod.rs`, which use the conventional `mod tests;` pattern.

### Suggested direction

Delete the orphan, or migrate the binary module to the `mod tests;` convention the other two object families use — one consistent pattern either way. Worth doing before the Event_Time_Stamps storage slice adds more binary tests.

### Relevant code

- `crates/bacnet-objects/src/binary/tests/mod.rs`
- `crates/bacnet-objects/src/binary/mod.rs`

### Related

#235

### LINKED ISSUE #354: Restore bacnet-types no_std compilation
### Affected revision

Reproduced on `dev` at `944601b`.

### Behavior

`cargo check -p bacnet-types --no-default-features --locked` fails before compiling consumers. The crate declares a `no_std` mode, but several existing items rely on standard-prelude imports:

- `crates/bacnet-types/src/error.rs` uses `format!` without `alloc::format`.
- `crates/bacnet-types/src/primitives.rs` uses `format!` without `alloc::format`.
- `crates/bacnet-types/src/constructed/mod.rs` uses `String` without `alloc::string::String`.

The same check also reports existing unused `alloc::vec` and `alloc::format` imports.

### Expected work

- restore the intended `no_std` imports without changing the default-feature API;
- add `cargo check -p bacnet-types --no-default-features --locked` as local verification for the fix;
- verify a representative dependent crate in its supported feature configuration before adding any CI coverage.

Discovered while validating #353; no files in that PR cause the failure.

jscott3201:
Two compile-hygiene defects in disjoint crates, no API surface between them.

## `#354` — `bacnet-types` does not compile without `std`

`cargo check -p bacnet-types --no-default-features --locked` failed with 8 errors before any consumer compiled. The crate declares `no_std` (`lib.rs:14`) and pulls in `alloc` (`:17`), but three modules used items that only the std prelude supplies:

| file | missing |
|---|---|
| `error.rs:25-28` | `format!` → `alloc::format` |
| `primitives.rs:155,223,277` | `format!` → `alloc::format` |
| `constructed/mod.rs:551` | `String` → `alloc::string::String` |

`enums/mod.rs:8-9` also imported `alloc::format` that nothing uses — every remaining `format!` under `enums/` is in test or serde-test code.

All four edits are inside `#[cfg(not(feature = "std"))]` blocks, so **the default-feature API is unchanged**.

**Not a regression.** `git show v0.10.1:crates/bacnet-types/src/error.rs` has the same missing import, so the *published* 0.10.1 is broken while `Cargo.toml:10` lists the crate under the crates.io `no-std` category. It rotted unnoticed because `.github/workflows/ci.yml` contains zero `--no-default-features` invocations — the configuration is declared and categorized but compiled nowhere.

`bacnet-types` is the only workspace member with a `std` feature, and no member depends on it with `default-features = false`, so nothing else is affected.

## `#263` — orphaned module file

`crates/bacnet-objects/src/binary/tests/mod.rs` was three `mod` declarations and no test code. `binary/mod.rs` has no `mod tests;`, and all four test files are already reached via `#[path]` attributes (`input.rs:297`, `output.rs:347`, `value.rs:357`, `mod.rs:24`). Deleting it loses nothing; wiring it up instead would have double-declared three modules and still missed `generic_event_properties.rs`, which the orphan never listed.

`binary` was the only orphan — the other nine `*/tests/mod.rs` files in the workspace all have a matching `mod tests;` in their parent.

## Verification

- `cargo check -p bacnet-types` across all four feature combinations (`--no-default-features`, `--no-default-features --features serde`, default, `--features serde`) — all compile, zero unused-import warnings
- `cargo test --workspace --exclude rusty-bacnet` — pass
- `cargo clippy --workspace --exclude rusty-bacnet --all-targets --locked` — clean

## Deliberately not in this PR

**No CI change.** `AGENTS.md:16` says not to add CI unsolicited, and `#354` itself asks to verify a dependent crate before adding coverage. Without a gate this will rot again, so it is worth an owner decision: the cheapest guard is adding `--no-default-features` to an existing job rather than a new one.

**Pre-existing warning left alone.** `server/mod.rs:773` re-exports `EXECUTED_CONFIRMED`/`EXECUTED_UNCONFIRMED`, which are only consumed by `pics/tests.rs:775-782` under `#[cfg(test)]`, so the non-test build warns on both. It dates to `ce0ef1e` (#192) and belongs to `bacnet-server`, not to either issue here.

Closes #354
Closes #263

### DIFF
--- a/crates/bacnet-objects/src/binary/tests/mod.rs
+++ b/crates/bacnet-objects/src/binary/tests/mod.rs
@@ -1,3 +1,0 @@
-mod command_failure;
-mod input_output;
-mod value;
--- a/crates/bacnet-types/src/constructed/mod.rs
+++ b/crates/bacnet-types/src/constructed/mod.rs
@@ -5,7 +5,7 @@
 //! All types follow the same `no_std`-compatible pattern used in `primitives.rs`.
 
 #[cfg(not(feature = "std"))]
-use alloc::{vec, vec::Vec};
+use alloc::{string::String, vec::Vec};
 
 use crate::error::Error;
 use crate::primitives::{Date, ObjectIdentifier, Time};
--- a/crates/bacnet-types/src/enums/mod.rs
+++ b/crates/bacnet-types/src/enums/mod.rs
@@ -5,9 +5,6 @@
 //! without panicking. Every type provides `from_raw` / `to_raw` for
 //! wire-level conversion and a human-readable `Display` impl.
 
-#[cfg(not(feature = "std"))]
-use alloc::format;
-
 // ---------------------------------------------------------------------------
 // Macro to reduce boilerplate for newtype enum wrappers
 // ---------------------------------------------------------------------------
--- a/crates/bacnet-types/src/error.rs
+++ b/crates/bacnet-types/src/error.rs
@@ -5,7 +5,7 @@
 //! and timeouts.
 
 #[cfg(not(feature = "std"))]
-use alloc::string::String;
+use alloc::{format, string::String};
 #[cfg(feature = "std")]
 use std::time::Duration;
 
--- a/crates/bacnet-types/src/primitives.rs
+++ b/crates/bacnet-types/src/primitives.rs
@@ -4,7 +4,7 @@
 //! [`Time`], and the [`PropertyValue`] sum type.
 
 #[cfg(not(feature = "std"))]
-use alloc::{string::String, vec::Vec};
+use alloc::{format, string::String, vec::Vec};
 
 use crate::enums::ObjectType;
 use crate::error::Error;


jscott3201:
Review correction to this PR body, from verification in a scratch worktree at head `85fab63`.

**My "No CI change" note suggested the cheapest guard would be "adding `--no-default-features` to an existing job". That will not work.** Both existing cargo jobs build test targets — `ci.yml:70` runs `cargo clippy --workspace --exclude rusty-bacnet --all-targets --locked` and `ci.yml:153/:179` run `cargo test --workspace --exclude rusty-bacnet --locked` — and `bacnet-types` still fails without `std` once test targets are included:

```
cargo check -p bacnet-types --no-default-features --all-targets --locked
  -> 72 errors (30x cannot find macro `format`, 11x cannot find macro `vec`,
                27x no method `to_string`, 4x cannot find type `Vec`)
```

Every one of those sits in a `#[cfg(test)]` or `#[cfg(all(test, feature = "serde"))]` module — `enums/tests.rs`, `bitstring.rs`, `enums/resolve.rs`, `constructed/tests.rs` — none in the modules this PR edits. The library itself is genuinely fixed; the test modules were never `no_std`-clean and are not made worse here.

So a future guard has to be **lib-only** (`cargo check -p bacnet-types --no-default-features` with no `--all-targets`), or the test modules need the same import treatment first. That is a bigger piece of work than this PR and stays out of scope, but the owner should have the accurate version when deciding on #354's follow-up.

Two smaller corrections for the record:

- "All four edits are inside `#[cfg(not(feature = "std"))]` blocks" is literally true for three of the four; the `enums/mod.rs` hunk deletes the `#[cfg(not(feature = "std"))]` attribute line itself along with the `use`. The conclusion it supports — default-feature API unchanged — was verified independently and holds.
- "clippy — clean" means **exit 0**, not warning-free. The workspace emits 906 warning lines on this head, essentially all `missing_docs`, and `ci.yml:18-21` documents that `-Dwarnings` is deliberately not set. Singling out the `EXECUTED_CONFIRMED` warning may have read as if it were the only one.

Verification that did not change: all four `bacnet-types` feature combinations compile with zero unused-import warnings; `cargo test --workspace --exclude rusty-bacnet --locked` exits 0; no test coverage was lost to the `#263` deletion — `cargo test -p bacnet-objects --lib -- --list` yields 999 tests on both base and head with byte-identical `binary::` lists (91 tests: 40 input_output + 32 value + 10 command_failure + 9 generic_event_properties).

Issue #186, opened 2026-07-27
Recipient BACnetAddress network-number is discarded, so notifications to a remote-network recipient are sent as local unicasts with no DNET/DADR


jscott3201:
### Affected revision

Observed on `dev` at commit `3990f3e`.

### Behavior

`crates/bacnet-objects/src/notification_class/mod.rs:568-581` correctly decodes and preserves a recipient's `network_number` — that is what #126 fixed. The notification sender then throws it away. The confirmed path keeps only `addr.mac_address` (`crates/bacnet-server/src/server/event_notifications.rs:199-204`), and the unconfirmed path passes only `&addr.mac_address` (`:317-321`).

`NetworkLayer::send_apdu` forwards to `send_apdu_with_data_attributes` (`crates/bacnet-network/src/layer.rs:148-171`), which hardcodes `destination: None` (`:160`). The encoded NPCI therefore has the destination specifier bit clear, so there is no DNET, no DLEN, no DADR and no Hop Count, and the frame goes out as a unicast on the local link.

Clause 21.6 (p884) defines `network-number` with "A value of 0 indicates the local network", so a non-zero value is by definition a recipient on a different BACnet network. Clause 6.2 (p65) defines control-octet bit 5 as "Destination specifier where: 0 = DNET, DLEN, DADR, and Hop Count absent". An APDU addressed to a remote recipient but sent with that bit clear is delivered to whichever device happens to hold that MAC on the local link — the wrong device, or nobody.

The correct API exists and is unused on this path. `NetworkLayer::send_apdu_routed` is at `crates/bacnet-network/src/layer.rs:315-334`, and `crates/bacnet-server/src/server/cov_notifications.rs:4-31` already demonstrates the routed-versus-local branch for COV. `event_notifications.rs` contains no reference to `send_apdu_routed` and no reference to the `RouterTable`, whose `lookup(network)` is at `crates/bacnet-network/src/router_table.rs:219`.

### Suggested direction

- carry the recipient's `network_number` alongside its MAC through the notification sender instead of dropping it at `:199-204` and `:317-321`;
- send through `send_apdu_routed` when `network_number` is non-zero, resolving the next-hop router MAC via `RouterTable::lookup`, and keep `send_apdu` for `network_number` 0 — the same shape as `send_cov_apdu`;
- return an explicit unresolved-route outcome when no route to the network is known, rather than falling back to a local unicast;
[...]

Issue #126, opened 2026-07-14
Preserve recipient BACnetAddress network numbers


jscott3201:
### Affected revision

Observed on rusty-bacnet v0.10.1 at commit `8b8c0cba3e92caa30e145c91abedfa003c64f2c9`.

### Behavior

`Recipient_List` projects an address recipient as MAC bytes only. The write and filtering paths reconstruct that value with network number zero. A recipient configured for a nonzero BACnet network therefore cannot round-trip through the object property representation.

### Suggested direction

- encode and decode the complete `BACnetAddress`;
- document a compatibility policy for existing MAC-only values;
- preserve local, remote, and broadcast address forms;
- add a nonzero-network round-trip test.

### Relevant code

- `crates/bacnet-types/src/constructed/mod.rs`
- `crates/bacnet-objects/src/notification_class/mod.rs`

Issue #125, opened 2026-07-14
Resolve Notification Class device recipients instead of broadcasting


jscott3201:
### Affected revision

Observed on rusty-bacnet v0.10.1 at commit `8b8c0cba3e92caa30e145c91abedfa003c64f2c9`.

### Behavior

Configured device recipients do not produce a target MAC in the EventNotification sender. Confirmed and unconfirmed notifications for those recipients are consequently broadcast.

A targeted device recipient is silently widened to network broadcast.

### Suggested direction

- resolve device recipients through a device-to-address source;
- return an explicit unresolved-recipient outcome when resolution fails;
- keep broadcast as a separately configured policy;
- test resolved and unresolved device recipients.

### Relevant code

- `crates/bacnet-objects/src/notification_class/mod.rs`
- `crates/bacnet-server/src/server/event_notifications.rs`

jscott3201:
**Advanced by PR #357, not closed.**

### Done

The half this issue called the actual defect — *"A targeted device recipient is silently widened to network broadcast"* — is fixed. `BACnetRecipient::Device` now resolves to an explicit unresolved outcome and the recipient is **skipped**, with `warn!(notification_class, %device, "Skipping recipient: no address binding for this device instance")`.

Two things that made the old behavior worse than "widened" are also gone:

- The confirmed path broadcast a **ConfirmedRequest** APDU for a device recipient (`event_notifications.rs:288-296` before the change). Clause 6.3 (extracted Standard line 4658) permits only Unconfirmed-Request-PDUs at a broadcast address, and lines 3075-3079 have every receiver silently discard a broadcast Confirmed-Request — so it was both non-conformant and undeliverable.
- The unconfirmed path broadcast the alarm to every device on the link.

Test: `device_recipient_is_skipped_not_broadcast`.

### Still owed — the resolution itself

Resolving a device instance to an address needs a device-to-address binding, and **the server maintains none**. `BACnetServer` has no address cache and no Who-Is/I-Am correlation for outbound addressing; a grep for `device_address` / `address_binding` across `bacnet-server` returns nothing.

What a fix needs:

- a device-instance → `BACnetAddress` store, populated from observed I-Am (and from Device_Address_Binding if that property is implemented);
- Who-Is on an unbound instance, with the notification queued pending the I-Am — which makes sending stateful and asynchronous, the same shape #186 needs for unresolved routes;
- an explicit unresolved outcome retained for when discovery fails, rather than falling back to a broadcast.

[...]

Issue #124, opened 2026-07-14
Separate empty recipient outcomes from EventNotification broadcast policy


jscott3201:
### Affected revision

Observed on rusty-bacnet v0.10.1 at commit `8b8c0cba3e92caa30e145c91abedfa003c64f2c9`.

### Behavior

Missing Notification Class, unreadable or malformed configuration, an empty recipient list, and no day/time/transition match all become an empty vector. The production sender treats every empty result as permission to broadcast.

These operationally different outcomes cannot be diagnosed or governed independently.

### Suggested direction

- return a typed lookup and filtering result;
- distinguish missing, invalid, empty, and no-match cases;
- make any broadcast fallback explicit and configurable;
- test each outcome separately.

### Relevant code

- `crates/bacnet-objects/src/notification_class/mod.rs`
- `crates/bacnet-server/src/server/event_notifications.rs`

jscott3201:
**Advanced by PR #357, not closed.**

### Done — the broadcast fallback is gone

The behavior this issue objected to — *"The production sender treats every empty result as permission to broadcast"* — no longer exists. An empty recipient set now distributes nothing.

Grounded rather than assumed: Clause 13.2.5 (extracted Standard line 41547) distributes "to the notification-clients specified by the Recipient_List input", and an exhaustive scan of Clause 13 event reporting and Clause 12.21 found **no** broadcast-on-empty rule. Two further passages make it stronger than I first argued — `Recipient_List` is defined as "a list of **zero or more** recipient destinations" (line 18819), and the History of Revisions records Addendum ax to 135-2012 item 1, *"Remove Incorrect Recipient_List Requirement to be Non-empty"* (line 91820). An empty list is explicitly legal and explicitly means no clients.

Destination outcomes are now typed at the send site (`RecipientRoute`), and the two undeliverable cases log **separately**, since they are different operator problems: an unreachable network is a routing gap, an unbound device instance is an unaddressable recipient.

Tests: `empty_recipient_list_distributes_nothing`, `event_notification_missing_class_distributes_nothing`, and `one_unresolvable_recipient_does_not_suppress_the_others`.

[...]

PR #357, opened 2026-08-17
fix(server): route notification recipients by their BACnetAddress

### LINKED ISSUE #185: A Notification Class recipient configured as a local broadcast (zero-length MAC) fails to send instead of broadcasting
### Affected revision

### Behavior

`BACnetAddress` recipients are decoded with `MacAddr::from_slice(mac)` and no length check — `crates/bacnet-objects/src/notification_class/mod.rs:568-581` on the notification path, and `:238` on the write path — so a spec-conformant zero-length mac-address survives as an empty `MacAddr`.

The unconfirmed sender then calls `network.send_apdu(&buf, &addr.mac_address, ...)` unconditionally (`crates/bacnet-server/src/server/event_notifications.rs:317-327`). `NetworkLayer::send_apdu` reaches `TransportPort::send_unicast` (`crates/bacnet-network/src/layer.rs:148-171`, and the data-attributes variant delegates at `crates/bacnet-transport/src/port.rs:98-105`), and the BACnet/IP implementation calls `decode_bip_mac` first (`crates/bacnet-transport/src/bip/mod.rs:571-574`), which rejects any length other than 6 (`crates/bacnet-transport/src/bvll.rs:150-156`, "BIP MAC must be 6 bytes, got 0"). The result is a `warn!` at `event_notifications.rs:323-326` and a dropped notification. The confirmed path has the same shape: `:199-204` keeps only the MAC, `:242-250` sends.

Clause 21.6 (p884) defines a zero-length MAC as the way to ask for a broadcast:

```
BACnetAddress ::= SEQUENCE {
      network-number   Unsigned16,     -- A value of 0 indicates the local network
      mac-address      OCTET STRING    -- A string of length 0 indicates a broadcast
      }
```

Clause 12.21 (p290) makes that exact configuration mandatory for a whole class of device: "If the Recipient_List is not writable and the device is not using local Notification Forwarder objects, then the Recipient_List shall have a length of 1 with the Recipient set to a local broadcast, Valid Days set to all days, From Time set to 00:00:00.0, To_Time set to 23:59:59.99, Process Identifier set to 0, Issue Confirmed Notifications set to FALSE, and all bits in Transitions set to TRUE." A device shipping the prescribed list sends nothing. The only way this stack currently broadcasts a notification is the untyped empty-recipient fallback described in #124 — the spec's deliberate way of requesting a broadcast is the one that fails.

### Suggested direction

- treat a zero-length mac-address as an explicit broadcast request rather than as a unicast MAC;
- with `network_number` 0, route it to `NetworkLayer::broadcast_apdu`; a non-zero `network_number` is a remote broadcast, which Clause 6.3 (p71) defines as "DNET shall specify the network number of the remote network and DLEN shall be set to zero" — that belongs with the routed-recipient work, not with a plain local broadcast;
- reject Issue Confirmed Notifications on a broadcast recipient, per Clause 6.3 (p71): "Of the BACnet APDUs, only the BACnet-Unconfirmed-Request-PDU may be transmitted using a multicast or broadcast network layer address";
- make this an explicit recipient outcome rather than a fallback, so it stays distinguishable from the empty and no-match cases in #124;
- test a zero-length MAC recipient with `network_number` 0 (broadcast sent), and a zero-length MAC recipient with Issue Confirmed Notifications TRUE (rejected, not silently unicast).

### Relevant code

- `crates/bacnet-server/src/server/event_notifications.rs`

jscott3201:
The EventNotification sender read a recipient's MAC and nothing else, so three of the four destinations Clause 21 can express came out wrong — including the one the standard *requires* a whole class of device to ship.

## The four destinations

Clause 21 (extracted Standard line 56354) gives `BACnetAddress` two independent fields:

```
BACnetAddress ::= SEQUENCE {
      network-number   Unsigned16,     -- A value of 0 indicates the local network
      mac-address      OCTET STRING    -- A string of length 0 indicates a broadcast
      }
```

Their combinations are four different sends, so this resolves them once up front instead of re-reading a MAC at each send site:

| `network-number` | `mac-address` | now | before |
|---|---|---|---|
| 0 | 6 bytes | `send_apdu` | same |
| 0 | **empty** | `broadcast_apdu` | unicast with an empty MAC |
| N | **empty** | `broadcast_to_network(N)` | unicast with an empty MAC |
| N | 6 bytes | skipped, logged | local unicast to the wrong device |
| **65535** | **empty** | `broadcast_global_apdu` | unicast with an empty MAC |

### The local-broadcast case is the severe one (`#185`)

Clause 12.21 (line 18861) requires that a device whose `Recipient_List` is not writable and which uses no local Notification Forwarder objects ship exactly one entry, "with the Recipient set to a local broadcast". That reached `send_apdu` with a zero-length MAC, which BACnet/IP rejects outright — `decode_bip_mac` raises *"BIP MAC must be 6 bytes, got 0"* — producing a `warn!` and a dropped notification.

**A device configured the way the standard prescribes distributed nothing at all.**

### Remote recipients (`#186`)

`network_number` was decoded correctly by `#126` and then discarded by the sender, so the NPDU went out with the destination specifier bit clear: no DNET, no DLEN, no DADR. A recipient on network 1000 was delivered to whichever device held that MAC on the *local* link.

Remote broadcast now uses `NetworkLayer::broadcast_to_network`, which is exactly the Clause 6.3 form (line 4679): *"DNET shall specify the network number of the remote network and DLEN shall be set to zero."*

Remote **unicast** needs the MAC of a next-hop router. `RouterTable` lives in `BACnetRouter`, which `BACnetServer` does not hold — there is no route resolution reachable from here. So it is skipped and logged rather than misdelivered. That is strictly better than the current behavior but is not a resolution; see the scope note below.

## Two invented broadcasts removed

**Empty `Recipient_List` no longer broadcasts.** Clause 13.2.5 (line 41547) distributes "to the notification-clients specified by the Recipient_List input", and an empty list specifies none. There is no broadcast-on-empty rule anywhere in the Standard.

⚠️ **This is a behavior change.** An object whose `Notification_Class` names a class that does not exist now stays silent instead of alarming every device on the link. `event_notification_missing_class_falls_back_to_defaults` asserted the old behavior and is replaced by `event_notification_missing_class_distributes_nothing`.

**Confirmed notifications to a broadcast address are skipped.** Clause 6.3 (line 4658): *"only the BACnet-Unconfirmed-Request-PDU may be transmitted using a multicast or broadcast network layer address."* Device recipients previously broadcast a **ConfirmedRequest** APDU, which that sentence forbids and which has nowhere to return its SimpleACK from. Skipped rather than downgraded to unconfirmed, since downgrading would silently drop the acknowledgment the recipient was configured to require.

## Scope — what this does and does not close

Only `#185` closes outright. I would rather be explicit than let the others look finished:

| issue | status |
|---|---|
| **#185** | **closed** — both broadcast forms work |
| #186 | remote broadcast works and remote unicast no longer misdelivers; **route resolution still owed** |
| #125 | device recipients are no longer silently widened to a broadcast; **address binding still owed** |
| #124 | the broadcast fallback is gone and destinations are typed at the send site; **the four-way taxonomy at the lookup boundary is still owed** — `get_notification_recipients_strict` still returns `Option<Vec<..>>`, so "class missing", "list empty", and "nothing matched today" remain indistinguishable to the caller. Separating them means changing that public `bacnet-objects` signature, which is a larger change than this one |

## Tests

New module `event_recipient_routing_tests.rs` — `event_notifications_tests.rs` is at **672 of the 700-LOC cap**, so this could not go there.

Eleven tests, one per destination shape plus the two unresolvable ones, the empty list, and a mixed list proving one bad recipient does not suppress its neighbours. The transport records broadcasts and unicasts separately and keeps each unicast's **target MAC**, which is what catches a remote recipient delivered locally.

**Verified they discriminate:** run against `dev` in a throwaway worktree, **9 of 11 fail**. The two that pass are `local_unicast_recipient_reaches_its_mac` and `confirmed_notification_to_a_local_unicast_recipient_is_sent` — the paths this change must not disturb.

One trap worth flagging: the confirmed path `tokio::spawn`s its send, so the first draft of the confirmed test passed **vacuously** on both branches — it sampled the transport before the spawned task ran. `distribute_to` now yields until the send lands, which is what turned that test into a real discriminator.

## Verification

- `cargo test --workspace --exclude rusty-bacnet` — 34 suites green (422 in `bacnet-server`)
- `cargo clippy --workspace --exclude rusty-bacnet --all-targets --locked` — no errors
- `bash .github/scripts/check-file-size.sh` — all files within the cap

Refs #124, #125, #186
Closes #185

### DIFF
--- a/crates/bacnet-objects/src/notification_class/mod.rs
+++ b/crates/bacnet-objects/src/notification_class/mod.rs
@@ -443,9 +443,8 @@ pub fn get_notification_recipients(
 ///   was found and its `Recipient_List` decoded; `list` may be empty after
 ///   filtering (no recipient viable right now).
 /// - `None` — a NotificationClass matched but its stored `Recipient_List`
-///   FAILED to decode: the configured recipients are unknown, and
-///   delivering to a decodable prefix (or falling back to the
-///   would notify the wrong set of devices.
+///   FAILED to decode: the configured recipients are unknown, and delivering
+///   to a decodable prefix would notify the wrong set of devices.
 pub fn get_notification_recipients_strict(
     db: &ObjectDatabase,
     notification_class: u32,
--- a/crates/bacnet-objects/src/notification_class/mod.rs
+++ b/crates/bacnet-objects/src/notification_class/mod.rs
@@ -363,9 +363,12 @@ fn find_notification_class(
 /// When no NotificationClass matches the given number (the object's
 /// `Notification_Class` was never configured or points at a missing class),
 /// the spec leaves the projection undefined; we fall back to the BACnet
-/// defaults — `Priority = 255` (lowest) and `Ack_Required = false` — so the
-/// notification is still delivered with a benign priority and no
-/// acknowledgement demand rather than dropped silently.
+/// defaults — `Priority = 255` (lowest) and `Ack_Required = false`.
+///
+/// Those defaults no longer reach the wire on the server's send path. A class
+/// that does not exist also names no recipients, and the sender distributes
+/// nothing when the recipient set is empty, so the fallback survives only for
+/// direct callers of this function.
 pub fn resolve_transition_priority_ack(
     db: &ObjectDatabase,
     notification_class: u32,
@@ -442,7 +445,7 @@ pub fn get_notification_recipients(
 /// - `None` — a NotificationClass matched but its stored `Recipient_List`
 ///   FAILED to decode: the configured recipients are unknown, and
 ///   delivering to a decodable prefix (or falling back to the
-///   no-recipients broadcast path) would notify the wrong set of devices.
+///   would notify the wrong set of devices.
 pub fn get_notification_recipients_strict(
     db: &ObjectDatabase,
     notification_class: u32,
--- a/crates/bacnet-server/src/server/dcc_event_detection_tests.rs
+++ b/crates/bacnet-server/src/server/dcc_event_detection_tests.rs
@@ -97,6 +97,12 @@ async fn dcc_states_preserve_delayed_detection_but_suppress_distribution() {
             .unwrap(),
         ))
         .unwrap();
+        // A recipient that WOULD be broadcast to, so the empty assertion can
+        // only be satisfied by the DCC gate rather than by an unnamed recipient.
+        db.add(Box::new(
+            super::event_notifications_tests::notification_class_0_broadcasting(),
+        ))
+        .unwrap();
 
         let mut server = BACnetServer::start(ServerConfig::default(), db, transport)
             .await
--- a/crates/bacnet-server/src/server/event_enable_distribution_tests.rs
+++ b/crates/bacnet-server/src/server/event_enable_distribution_tests.rs
@@ -126,6 +126,12 @@ impl Fixture {
             .unwrap(),
         ))
         .unwrap();
+        // These objects leave Notification_Class at its default of 0, so class
+        // 0 has to exist and name a recipient for anything to be distributed.
+        db.add(Box::new(
+            super::event_notifications_tests::notification_class_0_broadcasting(),
+        ))
+        .unwrap();
 
         let sent = StdArc::new(StdMutex::new(Vec::new()));
         Self {
--- a/crates/bacnet-server/src/server/event_enrollment_task_tests.rs
+++ b/crates/bacnet-server/src/server/event_enrollment_task_tests.rs
@@ -94,6 +94,14 @@ async fn spawned_task_advances_and_fires_the_time_delay_countdown() {
         .unwrap(),
     ))
     .unwrap();
+    // A recipient that WOULD be broadcast to. The Event Enrollment path has no
+    // send call today, so this changes nothing now — it makes the assertion
+    // below a live tripwire: when #127 gives that path a sender, this test goes
+    // red and asks to be updated instead of quietly staying green.
+    db.add(Box::new(
+        super::event_notifications_tests::notification_class_0_broadcasting(),
+    ))
+    .unwrap();
 
     let config = ServerConfig {
         event_enrollment_interval_secs: 1,
--- a/crates/bacnet-server/src/server/event_notifications.rs
+++ b/crates/bacnet-server/src/server/event_notifications.rs
@@ -20,10 +20,11 @@ const GLOBAL_BROADCAST_NETWORK: u16 = 0xFFFF;
 ///
 /// Clause 21's `BACnetAddress` gives a recipient two independent knobs —
 /// `network-number`, where "A value of 0 indicates the local network", and
-/// `mac-address`, where "A string of length 0 indicates a broadcast". Their
-/// four combinations are four different sends, not one unicast with edge
-/// cases, which is why this is resolved once up front rather than decided at
-/// each send site.
+/// `mac-address`, where "A string of length 0 indicates a broadcast" — and
+/// Clause 6.3 reserves network 65535 for the global broadcast. Their
+/// combinations are distinct sends rather than one unicast with edge cases,
+/// which is why this is resolved once up front rather than decided at each
+/// send site.
 enum RecipientRoute {
     /// Network 0 with an explicit MAC: unicast on the local network.
     LocalUnicast(MacAddr),
@@ -78,7 +79,7 @@ impl RecipientRoute {
     /// broadcast MAC — `X'FFFFFFFFFFFF'` on Ethernet, an all-ones host portion
     /// on BACnet/IP, `X'FF'` on MS/TP — reads as `LocalUnicast` here and is
     /// still sent confirmed. Recognizing those needs the transport's broadcast
-    /// address, which this layer does not have.
+    /// address, which this layer does not have. Tracked by #360.
     fn permits_confirmed(&self) -> bool {
         matches!(self, Self::LocalUnicast(_))
     }
--- a/crates/bacnet-server/src/server/event_notifications.rs
+++ b/crates/bacnet-server/src/server/event_notifications.rs
@@ -1,4 +1,5 @@
 use super::*;
+use bacnet_types::constructed::BACnetRecipient;
 use bacnet_types::enums::EventType;
 
 pub(super) struct NotificationTransition {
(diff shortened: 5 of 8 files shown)


jscott3201:
**Pushed `636bb34` — a defect in this PR's own first commit, found while probing the edges of `network-number`.**

Network **65535** with a zero-length MAC is a destination of its own, not a remote network that happens to be numbered 65535. Extracted Standard line 4681:

> A global broadcast, indicated by a DNET of X'FFFF', is sent to all networks through all routers.

Confirmed at lines 4236 and 77618 (`X'FFFF'  DNET = 65535 (Global broadcast)`).

My first commit resolved it to `RemoteBroadcast(65535)` → `broadcast_to_network(65535)`, which **rejects that DNET by design** (`layer.rs:284`: *"dest_network 0xFFFF is reserved for global broadcasts; use broadcast_global_apdu instead"*). So a recipient configured as a global broadcast produced a send error and reached nobody — a fresh way to drop a notification, introduced by the very change that fixes two others.

Now `GlobalBroadcast` is its own route and sends via `broadcast_global_apdu`. Network 65535 carrying a unicast MAC stays skipped, since a broadcast requires DLEN zero and that recipient contradicts itself.

Note this was reachable from configuration: `bacnet-types` documents the field as *"Network number (0 = local network, 1-65534 = remote, 65535 = broadcast)"*, and `notification_class/tests/recipient_list.rs:35` already round-trips a net-65535 recipient — so the value was expected to appear in a `Recipient_List`, and nothing downstream handled it.

**Two new tests** (`zero_length_mac_on_network_65535_is_a_global_broadcast`, `global_broadcast_network_with_a_mac_is_skipped`).

**Updated discrimination numbers.** The PR body says 7 of 9 fail against `dev`; with the two additions it is now **9 of 11**, re-verified in a throwaway worktree at `origin/dev`. The 2 that pass remain `local_unicast_recipient_reaches_its_mac` and `confirmed_notification_to_a_local_unicast_recipient_is_sent` — the paths this change must not disturb.

Gates re-run on the amended branch: `fmt` clean, file-size cap OK, clippy 0 errors, `cargo test --workspace --exclude rusty-bacnet` 34 suites green.

jscott3201:
## Review response — `bd13a30`

Two-lane review (spec accuracy + adversarial) both returned REQUEST_CHANGES. All findings addressed.

### The one that mattered: a coverage regression I introduced

Removing the empty-recipients broadcast **defanged two DCC suppression tests**. `dcc_suppresses_periodic_event_send` and `dcc_states_preserve_delayed_detection_but_suppress_distribution` both build a database holding only an AnalogInput and a Device, so once a missing Notification Class stopped producing a broadcast, their `sent.is_empty()` assertions passed for the wrong reason — "no recipient was named" rather than "the DCC gate held".

I updated four other fixtures for exactly this reason and missed these two.

Verified by mutation, patching the DCC early return to `if false && comm_state.load(...) >= 1`:

| | before `bd13a30` | after |
|---|---|---|
| `dcc_suppresses_periodic_event_send` | **ok** ❌ | FAILED ✅ |
| `dcc_states_preserve_delayed_detection_...` | **ok** ❌ | FAILED ✅ |
| `dcc_states_preserve_per_write_detection_...` | FAILED ✅ | FAILED ✅ |

With the gate entirely removed, two of the three tests guarding it stayed green. Both fixtures now hold a Notification Class that *would* be broadcast to, so the assertion can only be satisfied by the gate.

### Already fixed before the review landed

The spec lane's **major** finding — a global-broadcast recipient (network 65535) resolving to `RemoteBroadcast(0xFFFF)` and being dropped by `broadcast_to_network`'s own 0xFFFF rejection — was fixed in `636bb34`. Both lanes reviewed `0ab32ad`; the adversarial lane re-verified against `636bb34` and withdrew it. Two tests cover it now.

### Documentation that promised the removed fallback

- `fire_event_notifications` and `build_and_send_event_notification` both said "or broadcast if none" — the opposite of what they now do.
- `resolve_transition_priority_ack` justified its `(255, false)` fallback as keeping the notification "still delivered ... rather than dropped silently". No longer true: a missing class also names no recipients, so those values cannot reach the wire from the send path. The fallback stays for direct callers; its stated reason is gone.
- Two dangling references to the "no-recipients broadcast path".
- A comment naming `undeliverable_reason`, which I renamed to `is_deliverable`.

### Two overclaims corrected rather than papered over

**`permits_confirmed` covers one of two broadcast spellings.** It recognizes the network-layer form (zero-length `mac-address`) but not the data-link form — `X'FFFFFFFFFFFF'` on Ethernet, an all-ones host portion on BACnet/IP, `X'FF'` on MS/TP (line 4674). A recipient using the literal broadcast MAC still goes out confirmed. Pre-existing — `dev` does the same — but my new guard advertised coverage it lacks. Catching it needs the transport's broadcast address, which this layer does not have, so it is now **documented on the method** and filed as **#360** rather than silently implied.

**The missing-class docstring overstated its citation.** It read as though Clause 13.2.5 mandates silence. It does not — 13.2.5 governs distribution *given* a Recipient_List, and Clause 12 defines `Notification_Class` with no rule for a dangling reference. The docstring now presents silence as a deliberate choice for an undefined configuration. The load-bearing empty-list claim is unaffected and the spec lane found it **stronger** than I stated: `Recipient_List` is defined as "a list of zero or more recipient destinations" (line 18819), and the History of Revisions records Addendum ax to 135-2012 item 1, *"Remove Incorrect Recipient_List Requirement to be Non-empty"* (line 91820).

### Verification

`fmt` clean · file-size cap OK · clippy 0 errors · `cargo test --workspace --exclude rusty-bacnet` 34 suites green.

Also removed two stale git worktrees the review left registered under `scratchpad/`.

jscott3201:
## A third vacuously-passing test — `84c98bf`

The adversarial review found two tests that my removal of the broadcast-on-empty fallback had defanged. I swept the rest of `bacnet-server` for the same defect class and found **a third**.

`event_enable_cleared_suppresses_periodic_time_delay_send` builds its own database holding only an AnalogInput and a Device. With no Notification Class, both of its `sent.is_empty()` assertions were satisfied by "no recipient was named" rather than by the `Event_Enable` gate they exist to guard.

What makes this one worth calling out: **its own docstring claimed to be mutation-proven.**

> Proven to fail when the `outcome.distribute` check at that site is replaced with `if true`.

That was true when written and my PR silently made it false. Re-running exactly that mutation on `lifecycle.rs:521`:

| fixture | gate disabled |
|---|---|
| no Notification Class (before `84c98bf`) | **ok** — vacuous |
| Notification Class present (after) | **FAILED** — discriminating |

### Sweep of every remaining absence-assertion in `bacnet-server`

| site | verdict |
|---|---|
| `event_enable_distribution_tests` (5 sites) | covered — routes through `Fixture::from_object`, fixed earlier in this branch |
| `dcc_event_detection_tests` (2 sites) | covered — both fixtures fixed earlier |
| `event_notifications_tests:96` | covered — `dcc_suppresses_periodic_event_send`, fixed earlier |
| `event_notifications_tests:450` | **intentional** — `event_notification_missing_class_distributes_nothing`; the absent class *is* the behavior under test |
| `event_notifications_tests:602` | covered — uses `db_with_high_limit_transition`, fixed earlier |
| `event_notifications_tests:731,739` | **this fix** |
| `event_enrollment_task_tests:142` | not affected — asserts the EE path emits nothing, which #127 owns. That path has no send call at all, so the fixture is not what makes it pass. Adding a recipient would not strengthen it until #127 lands |
| `cov_notifications_tests:222` | not affected — COV works off the subscription table, which never consults Notification Class recipients |

### Verification

`cargo test --workspace --exclude rusty-bacnet` 34 suites green · fmt clean · file-size cap OK · clippy 0 errors.

Three fixtures in this branch now carry a Notification Class purely so their negative assertions mean something. That is the real cost of removing an output path, and it is worth stating plainly: **four sibling fixtures went red and got fixed immediately; these three stayed green and had to be hunted.**

jscott3201:
## Independent verification complete — `9a39e6e`

A follow-up verification agent re-derived the third vacuous test **independently**, by mutation testing, and arrived at byte-for-byte the same fix that landed in `84c98bf`. It also confirmed the sweep is exhaustive, and caught two things I had not.

### Its item-2 sweep, confirming mine

It enumerated all six files in `crates/bacnet-server` that `impl TransportPort`, then checked every emptiness assertion in each:

| site | verdict |
|---|---|
| `event_enable_distribution_tests` — `fresh_binary_input_active_...`, `remove_last_alarm_value_disarms_msi` | clean, shared `Fixture::from_object` adds the NC |
| `event_enable_cleared_suppresses_per_write_send` | clean via `db_with_high_limit_transition` — which is also why the third DCC test was never vacuous |
| `event_recipient_routing_tests` — all ten assertions | real; `distribute_to` builds NC 0 from the passed destinations |
| `cov_notifications_tests` | no emptiness assertion at all; its one `is_empty` is a positive poll loop |
| `segmentation_tests`, `server/tests.rs`, `handlers/` | no transport emptiness assertions |

It also confirmed the per-write gate is independently covered: mutating `event_notifications.rs:160` to `if true` fails 4 tests.

The tell it noted is the one I should have seen: the vacuous test's **positive twin** — `periodic_time_delay_carries_detector_event_type_to_wire`, thirty lines below — already added the Notification Class. A negative test whose positive twin needs a fixture the negative one lacks is exactly the shape of a vacuous assertion.

### Two things it caught that I had missed

**1. I shipped a mangled sentence.** My doc-correction pass in `bd13a30` deleted `no-recipients broadcast path)` from `get_notification_recipients_strict`'s rustdoc but left the opening paren and a dangling "the":

```
///   delivering to a decodable prefix (or falling back to the
///   would notify the wrong set of devices.
```

Unbalanced paren, ungrammatical, on a **public** API. A commit whose whole purpose was correcting documentation had no business shipping that. Fixed.

**2. A dead tripwire.** `event_enrollment_task_tests` asserts the EE path emits nothing (#127 owns giving it a sender). I had checked this and left it alone, reasoning correctly that it is not a live false guarantee — that path has no send call, so it is guarded twice over. But the agent's framing is better than mine: it is a **tripwire that will not trip**. When #127 lands, the test stays green instead of going red to ask for an update. Its fixture now holds a recipient, so it will trip.

### Verification at `9a39e6e`

`cargo test --workspace --exclude rusty-bacnet` 34 suites green · clippy 0 errors · fmt clean · file-size cap OK.

Independent confirmation at `bd13a30`: unpatched 9/9 routing tests pass; with the DCC gate patched out all three DCC distribution tests fail; deleting the `GlobalBroadcast` match arm makes the global-broadcast test fail 0-vs-1, proving `broadcast_to_network`'s `0xFFFF` guard would have dropped it. Spec anchors 4236 / 4681 / 77618-77619 verbatim. Full suites: `bacnet-server` 424 passed, `bacnet-objects` 999 passed.

Issue #184, opened 2026-07-27
Shipped rustdoc cites the wrong normative clauses for the Notification Class and event-parameter types


jscott3201:
### Affected revision

Observed on `dev` at commit `3990f3e`.

### Behavior

Doc comments across four crates anchor types and services to clauses covering unrelated subject matter. These comments are what an implementer follows when deciding what an encoding is supposed to be, and two of the wrong anchors have already been copied into the filed bodies of #152 and #154.

Notification Class family:

- `crates/bacnet-objects/src/notification_class/mod.rs:1` — "NotificationClass object per ASHRAE 135-2020 Clause 12.31". 12.31 is Access Point Object Type (p367). Notification Class is Clause 12.21 (p290).
- `crates/bacnet-objects/src/notification_class/mod.rs:5` and `crates/bacnet-types/src/constructed/mod.rs:309` — "Clause 12.15.5" for `BACnetDestination`. 12.15 is Life Safety Point Object Type (p247), and 135-2020 has no clause 12.15.5 at all — a grep of the standard text returns nothing. `BACnetDestination ::= SEQUENCE` is Clause 21.6 Base Types (p884); the `Recipient_List` property that uses it is Clause 12.21 (p290).
- `crates/bacnet-objects/src/event.rs:76` — "Clause 12.31.5 / 12.31.6" for the Notification Class `Priority` and `Ack_Required` array ordering. Same 12.31/12.21 transposition; both properties belong to Clause 12.21 (p290).

Event-parameter and alarm-service family:

- `crates/bacnet-types/src/constructed/mod.rs:658` and `crates/bacnet-types/src/constructed/event_parameter/mod.rs:2`, `:20`, `:43`, `:62` — "Clause 13.5" for `BACnetEventParameter`. 13.5 is the AcknowledgeAlarm Service (p685). `BACnetEventParameter ::= CHOICE` and `BACnetFaultParameter ::= CHOICE` are Clause 21.6 (p884); the `Event_Parameters` property is Clause 12.12 (p234), with algorithm semantics in Clause 13.3 (p649) and fault algorithms in Clause 13.4 (p678).
- `crates/bacnet-services/src/alarm_event/mod.rs:3` — "AcknowledgeAlarm (Clause 13.3)". 13.3 is Event Algorithms (p649); AcknowledgeAlarm is Clause 13.5 (p685).
[...]

PR #361, opened 2026-08-17
fix(client): refuse a peer max APDU below MinimumMessageSize

### LINKED ISSUE #319: Client can exceed tiny discovered maximum APDU limits for empty requests
### Behavior

A discovered device can advertise a maximum APDU length of 0 through 3. For a confirmed request with empty service data, the client sees the four-byte unsegmented APDU as too large and switches to segmentation. The segmentation path creates one empty segment even when the calculated segment payload capacity is zero, then sends a six-byte segmented Confirmed-Request. That frame exceeds the peer-advertised limit.

Non-empty payloads already fail when the segment payload capacity is zero. The empty-payload special case bypasses that check.

### Relevant path

- `crates/bacnet-client/src/client/requests.rs`: chooses segmentation from the discovered maximum APDU length
- `crates/bacnet-client/src/client/segmentation.rs`: calculates a zero-byte segment payload and sends the segment
- `crates/bacnet-encoding/src/segmentation.rs`: returns one empty segment before rejecting a zero maximum segment size
- `crates/bacnet-encoding/src/apdu/mod.rs`: encodes the segmented Confirmed-Request header

### Suggested regression

Seed a discovered device with maximum APDU lengths 0 through 3, send a confirmed request with empty service data, and assert a typed error with no outbound frame. Keep boundary coverage for the smallest limit that can carry an unsegmented empty confirmed request.

Found while reviewing #318; the behavior predates that change.

jscott3201:
A discovered device can advertise a maximum APDU length of 0..49. For a confirmed request with **empty service data** the client saw the four-octet unsegmented APDU as too large, switched to segmentation, computed a segment payload capacity of zero, and sent a **six-octet** segmented Confirmed-Request anyway — larger than the limit that caused the switch.

## Two holes, both closed

**1. The peer's value had no floor.** `requests.rs` took `d.max_apdu_length` from the device table verbatim. The only existing clamp — `cap_max_apdu_to_transport` against `VALID_MAX_APDU_LENGTHS` (`client/mod.rs:49,691`) — guards the client's **own** advertised value and never the peer's. That asymmetry is the bug: peer-controlled input reached a size calculation that local config could not.

Clause 5.2.1.2 (extracted Standard line 2054) derives the maximum transmittable length as the smallest of three values, including:

> (c) the maximum APDU size accepted by the remote peer device, **which must be at least 50 octets**.

Below that, no conformant APDU can be formed at all. Both target branches now check it.

**2. `split_payload` returned a segment at zero capacity.** The `payload.is_empty()` shortcut ran *before* the `max_segment_size == 0` check, so an empty payload yielded `Ok(vec![Bytes::new()])` regardless. A segment still costs its PDU header, so zero capacity cannot carry one. The checks are now in the other order.

Non-empty payloads were already rejected there — which is exactly why the empty-payload shortcut above them was the hole.

## A floor, not the six-value set

This deliberately does **not** reuse `is_valid_max_apdu_length`. I-Am carries `Max APDU Length Accepted` as an **Unsigned octet count**, not the four-bit code of Clause 20.1.2.5 — and Clause 20.1.2.5 notes the true value "may be larger than indicated in this parameter" (line 52653). So 600 and 1500 are legitimate advertisements and must not be rejected. Only the 50-octet floor from Clause 12.11.18 ("shall be greater than or equal to 50", line 14693) applies.

Reusing the six-value set here would have been a plausible-looking regression.

## Why refuse rather than clamp to 50

Both are spec-defensible, since a sub-50 advertisement is already non-conformant. Refusing keeps the client from inventing a capability the peer never claimed: a typed error names the misconfiguration, where a silent clamp would send the device frames it said it cannot hold. Flagging the alternative explicitly in case you prefer reachability over strictness.

## Server is unaffected

Its `client_max_apdu` comes from the four-bit header field via `decode_max_apdu`, which maps through `MAX_APDU_DECODE` and whose lowest value is 50. Reserved codes already error. The hole is reachable only through the client's discovered-device path.

## Tests

New module `peer_max_apdu_tests.rs` — `client/tests.rs` is at **698 of the 700-LOC cap**.

Verified against `dev` in a throwaway worktree:

| test | on `dev` | what it pins |
|---|---|---|
| `peer_max_apdu_too_small_for_an_unsegmented_apdu_sends_nothing` | **FAILED** | the defect — 0..3 emitted a six-octet frame on `dev`; now a typed error and **no outbound frame** |
| `peer_max_apdu_below_the_floor_but_above_the_apdu_size_is_also_refused` | ok | 4, 6, 49 were never defective; a deliberate new refusal under the 50-octet floor |
| `nonempty_request_to_peer_below_minimum_message_size_sends_nothing` | ok | never broken; pins that the same case is now refused *earlier*, before segmentation |
| `confirmed_request_at_minimum_message_size_is_sent_unsegmented` | ok | boundary — exactly 50 still transacts unsegmented |

One of three discriminates, which is the honest number: the other two guard behavior that must not change. A guard stuck permanently closed would satisfy the refusal tests alone, so the boundary test is deliberately paired with them.

Also `split_empty_payload_zero_segment_size_errors` at the encoding layer.

## Verification

`cargo test --workspace --exclude rusty-bacnet` 34 suites green · clippy 0 errors · fmt clean · file-size cap OK.

Closes #319

### DIFF
--- a/crates/bacnet-client/src/client/mod.rs
+++ b/crates/bacnet-client/src/client/mod.rs
@@ -755,6 +755,8 @@ mod cov_renewal_tests;
 mod cov_tests;
 #[cfg(test)]
 mod device_events_tests;
+#[cfg(test)]
+mod peer_max_apdu_tests;
 #[cfg(all(test, feature = "sc-tls"))]
 mod sc_builder_tests;
 #[cfg(test)]
--- a/crates/bacnet-client/src/client/peer_max_apdu_tests.rs
+++ b/crates/bacnet-client/src/client/peer_max_apdu_tests.rs
@@ -1,0 +1,181 @@
+//! What the client does with a peer that advertises an impossible APDU size.
+//!
+//! I-Am carries `Max APDU Length Accepted` as an Unsigned octet count rather
+//! than the four-bit code of Clause 20.1.2.5, so a peer can put any value on
+//! the wire — including values below the 50-octet MinimumMessageSize that
+//! Clause 12.11.18 and Clause 5.2.1.2(c) both require.
+//!
+//! Split from `tests.rs`, which is at the 700-LOC cap.
+
+use std::time::Instant;
+
+use bacnet_encoding::apdu::{self, encode_apdu, Apdu, SimpleAck};
+use bacnet_network::layer::{NetworkLayer, ReceivedApdu};
+use bacnet_transport::loopback::LoopbackTransport;
+use bacnet_types::enums::{ConfirmedServiceChoice, NetworkPriority, ObjectType, Segmentation};
+use bacnet_types::primitives::ObjectIdentifier;
+use bacnet_types::MacAddr;
+use bytes::BytesMut;
+use tokio::time::{timeout, Duration};
+
+use crate::discovery::DiscoveredDevice;
+
+use super::BACnetClient;
+
+/// Build a client whose device table holds one peer advertising
+/// `max_apdu_length`, plus the remote end of the loopback pair.
+async fn client_with_peer_advertising(
+    max_apdu_length: u32,
+) -> (
+    BACnetClient<LoopbackTransport>,
+    Vec<u8>,
+    NetworkLayer<LoopbackTransport>,
+    tokio::sync::mpsc::Receiver<ReceivedApdu>,
+) {
+    let client_mac = vec![0x01];
+    let remote_mac = vec![0x02];
+    let (client_transport, remote_transport) =
+        LoopbackTransport::pair(client_mac, remote_mac.clone());
+    let mut remote_network = NetworkLayer::new(remote_transport);
+    let remote_rx = remote_network.start().await.unwrap();
+    let client = BACnetClient::generic_builder()
+        .transport(client_transport)
+        .build()
+        .await
+        .unwrap();
+
+    client.device_table.lock().await.upsert(DiscoveredDevice {
+        object_identifier: ObjectIdentifier::new(ObjectType::DEVICE, 2003).unwrap(),
+        mac_address: MacAddr::from_slice(&remote_mac),
+        max_apdu_length,
+        segmentation_supported: Segmentation::NONE,
+        max_segments_accepted: None,
+        vendor_id: 42,
+        last_seen: Instant::now(),
+        source_network: None,
+        source_address: None,
+    });
+
+    (client, remote_mac, remote_network, remote_rx)
+}
+
+/// A peer advertising less than MinimumMessageSize gets no frame at all.
+///
+/// The empty-payload case is the one that used to escape. A four-octet
+/// unsegmented APDU looks too large for a limit of 0..3, so the client fell
+/// through to segmentation, where `max_segment_payload` yields a capacity of
+/// zero and `split_payload` returned one empty segment anyway — producing a
+/// six-octet segmented Confirmed-Request, larger still than the limit that
+/// triggered segmentation in the first place.
+#[tokio::test]
+async fn confirmed_request_to_peer_below_minimum_message_size_sends_nothing() {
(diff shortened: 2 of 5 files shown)


jscott3201:
## Review response — `8a92d91`

Adversarial review returned **APPROVE_WITH_NITS**. The one item that could have been blocking is cleared, and four findings are fixed.

### Not blocking: no transport can legitimately report under 50

The concern was that a small-MTU transport would land under the floor after the routed-NPDU header is subtracted, breaking something that worked. Verified against the actual code: BIP / BIP6 / Ethernet return 1476, MS/TP 480, and SC negotiates from the hub (default 1476). Table 6-1 (extracted Standard lines 4160-4174) gives **228** as the smallest data-link NPDU length in BACnet at all, and the routed subtraction removes only `dest_mac.len() + 4`.

### Fixed

**1. The error blamed the peer unconditionally.** The checked value is a minimum over several terms, and the peer is only sometimes the binding one — BACnet/SC recomputes its limit from the hub's Connect-Accept (`sc/mod.rs:226-232`), so a transport can fall below the floor while the peer is entirely conformant. The reviewer built exactly that case: this branch said *"the peer's advertised Max APDU Length Accepted is non-conformant"* where `dev` had correctly said *"transport max-APDU-length 48 leaves no valid BACnet APDU bucket"*. That is a diagnostic regression, and a real one. `check_transmittable_length` now takes both terms and names whichever bound the minimum.

**2. I misattributed a clause.** *"may be larger than indicated in this parameter"* is **Clause 20.1.2.5**, not 5.2.1.2. The phrase occurs exactly once in the Standard, at line 52653, inside 20.1.2.5 (which spans from 52650); 5.2.1.2 runs 2046-2080 and contains no such sentence. The substantive claim was right and only the attribution was wrong — which is precisely the failure mode #184 tracks, so I would rather correct it loudly than quietly. The PR body above has been corrected too.

**3. `Error::Segmentation` → `Error::Encoding`.** The check runs *before* any segmentation decision, and the sibling refusal in `cap_max_apdu_to_transport` already uses `Encoding`. A caller whose empty request was refused was seeing "segmentation error" for a request that was never segmented.

**4. Three literal `50`s → one.** `MINIMUM_MESSAGE_SIZE` is now exported from `bacnet-encoding` beside `MAX_APDU_DECODE`, whose first element it is.

### Test corrections — the more important half

**The loop test was failing for the wrong reason.** It asserted the error text before the no-frame check, so against `dev` it panicked at the wording assert (*"request timed out after 6s"*) and never looked at the wire. Assertion order is reversed.

**And it overclaimed.** The reviewer probed `dev` directly: advertised **0 and 3** each delivered a six-octet `[0a 05 00 00 01 0c]` — the defect. Advertised **4, 6 and 49** delivered a conforming four-octet `[02 05 00 0c]` — no defect at all. My single test presented all seven values as "the defect". They are now two tests:

- `peer_max_apdu_too_small_for_an_unsegmented_apdu_sends_nothing` — 0..3, the actual bug
- `peer_max_apdu_below_the_floor_but_above_the_apdu_size_is_also_refused` — 4, 6, 49, an honest **deliberate behavior change** under Clause 12.11.18's floor

Plus unit tests pinning which term the error blames, that the six-value set is **not** the check, and that 51 / 600 / 1500 / `u16::MAX` remain acceptable.

### Filed, not fixed

**#362** — the `Routed` branch never consults the device table, so a routed peer's advertised `Max APDU Length Accepted` is still unenforced and can be overrun. Pre-existing and out of scope for #319, but this PR adds a 5.2.1.2 guard to that exact branch, so leaving the clause reference implying compliance would be misleading. The branch now passes `LengthBoundedBy::LocalConfig` with a comment saying so.

### Verification

`cargo test --workspace --exclude rusty-bacnet` 34 suites green (74 in `bacnet-client`) · clippy 0 errors · fmt clean · file-size cap OK.

Issue #401, opened 2026-08-21
Decide and document in-service write policy for Escalator status properties


jscott3201:
## Context

Follow-up from PR #396 review. This is a behavior-policy question, not a confirmed conformance defect.

Clause 12.60 explicitly requires Escalator status properties such as `Operation_Direction` and `Escalator_Mode` to be writable while `Out_Of_Service = TRUE` for simulation/testing. The general Clause 12 object rules also allow an `R` property to be writable at the implementor's option unless its property text prohibits that.

The current `EscalatorObject` accepts `Operation_Direction` writes regardless of `Out_Of_Service`. PR #396 intentionally preserved that behavior and only exercised the normative OOS path.

## Decision to make

Choose and document one policy:

1. Preserve in-service writes as an intentional implementation option; or
2. Gate these status-property writes to `Out_Of_Service = TRUE` as a stricter ownership model.

Whichever policy is chosen, apply it consistently to the relevant Escalator status properties rather than treating `Operation_Direction` in isolation.

## Acceptance criteria

- Re-read Clause 12.60 and the Clause 12 general writability rules and record the rationale.
- Define which Escalator status properties are application/device-owned in service.
- Add tests that pin both OOS behavior and the chosen in-service behavior.
- Do not describe the current unconditional write as non-conformant unless a prohibiting normative requirement is identified.

## Relevant code

- `crates/bacnet-objects/src/elevator.rs`
- `crates/bacnet-objects/src/elevator/tests.rs`

Related: #284 / PR #396 and the Escalator_Mode follow-up.

Issue #397, opened 2026-08-21
Atomic File handlers bypass FileObject data/record storage, so reads return empty payloads and record writes do not persist


jscott3201:
## Summary

Follow-up from PR #395 review.

`FileObject` already owns private stream (`data`) and record (`records`) storage, with `set_data` / `data` and `set_records` / `records` helpers. The server AtomicReadFile/AtomicWriteFile handlers do not use that storage. Instead, the stream path attempts a synthetic `PropertyIdentifier::from_raw(PROP_FILE_DATA)` read/write and falls back to empty data; the record read path synthesizes empty records from `Record_Count`; the record write path ACKs without persisting record payloads.

This is not a missing standardized `File_Data` property: Table 12-16 does not define one. File payload transfer belongs to the Clause 14 service data. The implementation needs an internal object/service boundary that reaches the File object's actual backing storage without inventing a network-visible property.

## Impact

- AtomicReadFile stream responses can ACK with empty octets despite populated `FileObject::data()`.
- AtomicWriteFile stream requests can fail downstream because the synthetic property is not writable.
- AtomicReadFile record responses report empty record bodies.
- AtomicWriteFile record requests can ACK while discarding the supplied records.

## Acceptance criteria

- Route AtomicReadFile/AtomicWriteFile through File-specific internal storage hooks or another narrow typed internal interface; do not add a non-standard File_Data property to the BACnet property model.
- Stream reads return the stored octets for the requested range.
- Successful stream writes persist data and keep `File_Size` coherent.
- Record reads return stored record payloads.
- Successful record writes persist payloads and keep `Record_Count`/`File_Size` coherent.
- Add end-to-end handler tests proving write→read persistence for both access methods and proving existing preloaded data is returned.
- Keep append (`-1`), modification timestamps/archive policy, locking/concurrency, and broader storage redesign separate unless required to make the bounded persistence path correct.

## Standard anchors

- Clause 12.13 / Table 12-16 — File object property model (no File_Data property).
- Clauses 14.1 and 14.2 — AtomicReadFile / AtomicWriteFile service payload semantics.

## Relevant code

- `crates/bacnet-objects/src/file.rs`
- `crates/bacnet-server/src/handlers/file.rs`
- `crates/bacnet-server/src/handlers/tests/file_access_method.rs`

Found during immutable-head review of PR #395.

Issue #384, opened 2026-08-17
Server-side reassembly SegKey uses the immediate-hop MAC, so a routed peer via redundant routers splits its own session


jscott3201:
### Affected revision

`dev` after PR #382. Noted during that PR's review.

### Behavior

The server's reassembly `SegKey` is `(source_mac, source_network, invoke_id)` where `source_mac` is the **immediate-hop** MAC. The client keys the same state through `response_tsm_mac`, which for a routed peer uses the NPDU SNET/SADR and is therefore route-independent.

On a network with redundant routers, a routed peer's PDUs can arrive via different immediate MACs mid-transaction. Server-side, a segment or Abort arriving via the second router produces a different `SegKey`: the segment draws a spurious no-session INVALID_APDU_IN_THIS_STATE Abort, and an Abort fails to clear the session (it lives until the reaper).

### Suggested direction

Key server-side reassembly the way the client keys its TSM: by the routed source when SNET/SADR is present, by the immediate MAC otherwise. The `(mac, network)` pair inside `SegKey` already carries both halves; the change is to stop letting the immediate MAC vary the key when the routed half pins the peer. Same consideration applies to `seg_ack_senders` (segmented response sends).

Related: #366 (the client-side counterpart work), #375.

PR #402, opened 2026-08-21
fix(client): make device-table address identity route-safe

### LINKED ISSUE #372: DeviceTable lookups conflate router and routed-peer identities and resolve duplicate matches nondeterministically
Surfaced by the code sweep for #362 / PR #369. Four facets of one root cause: the table is keyed by device instance, and its secondary lookups are linear scans whose match predicates encode identity assumptions that hold only partially.

### Facets

1. **`get_by_mac` matches routed entries by the router's MAC.** A routed peer's `mac_address` holds the router it was heard through (`dispatch.rs` I-Am arm). `get_by_mac` (`crates/bacnet-client/src/discovery.rs`) does not exclude routed entries, so a plain `confirmed_request(router_mac, ..)` — a local request to the router's *own* Device object — can pick up a routed peer's `max_apdu_length` and be bounded (or segmented) by a limit the router never advertised. PR #369's `get_by_network_address` guards the opposite direction (local entries can't satisfy a routed lookup); the mirror guard is missing.

2. **Duplicate matches resolve nondeterministically.** Both lookups are `values().find(..)` over a HashMap, so two entries sharing a MAC or a (SNET, SADR) pair — e.g. a stale entry for a re-commissioned device instance surviving up to DEVICE_MAX_AGE = 600 s (`client/lifecycle.rs`), or a gateway exposing several instances at one address — return an arbitrary one, and the segmentation decision can differ run to run.

3. **`add_device` cannot register a routed peer.** `crates/bacnet-client/src/client/discovery.rs` hardcodes `source_network: None`, `source_address: None`, `max_apdu_length: 1476`, so a caller who knows a routed peer's limit cannot seed it; the request path silently falls back to local config with no diagnostic.

4. **Routing decision and size bound can come from different rows.** `resolve_device` reads the table, releases the lock, and `confirmed_request_inner` re-locks and re-finds by (net, addr). A table update between the two (or facet 2) means the address the request goes to and the limit applied to it may come from different entries. Not a deadlock; a coherence gap.

### Expected work (directional)

- `get_by_mac` should skip entries with a `source_network` (they are not local devices at that MAC), with a test mirroring `get_by_network_address_ignores_local_devices`.
- Define duplicate resolution: most-recent `last_seen` wins is the natural rule for both lookups.
- `add_device` grows routed-identity and limit parameters, or a routed variant.
- Facet 4 is acceptable to document rather than fix (single lock hold across resolve+request would serialize all requests); a doc comment on `resolve_device` stating the coherence contract would do.

### Related

#362, #369.

jscott3201:
Closes #372

## Summary

`DeviceTable`'s secondary lookups encoded partial identity assumptions. This PR gives the client one deterministic interpretation of local and routed peer identity, ahead of #371 consuming `segmentation_supported`. No wire-format change.

## Standard anchors (kept light)

- **Clause 6.2.2** — SNET/SADR identify the original source; for a routed peer the router's local MAC is transport next-hop information only.
- **Clause 5.4** — transaction identity binds client/server BACnet addresses and Invoke ID; identity must not depend on which immediate route delivered traffic.

## Behavior changes

| Lookup / operation | Before | After |
|---|---|---|
| `get_by_mac(router_mac)` with a routed peer behind that router | returned the routed peer (wrong limits) | returns None unless a local row matches |
| duplicate local rows sharing a MAC | arbitrary HashMap-order pick | freshest `last_seen` wins |
| duplicate routed rows sharing SNET/SADR | fixed in #369 (freshest wins) — preserved | preserved |
| partial routing metadata (one of SNET/SADR set) | counted as local by `get_by_mac` | deliberately neither local nor routable until refreshed |
| manual routed registration | impossible; silent local-config fallback | new `add_routed_device(RoutedDeviceConfig)` |
| legacy `add_device(instance, mac)` | unchanged behavior | unchanged behavior; docs now state its defaults are manual registration values, not advertised capabilities |

`resolve_device`'s two-lock snapshot/coherence boundary is documented on the method per #372's explicit allowance: address snapshot under one lock, later capability lookup may observe a newer entry; no lock is held across network I/O.

## Files changed

- `crates/bacnet-client/src/discovery.rs` — address-space invariant docs on `DeviceTable`, `DiscoveredDevice::is_local`, local-only + freshest-wins `get_by_mac`, 7 focused table tests
- `crates/bacnet-client/src/client/discovery.rs` — coherence contract rustdoc on `resolve_device`, `add_routed_device`, corrected `add_device` docs
- `crates/bacnet-client/src/client/mod.rs` — import of `RoutedDeviceConfig`
- `crates/bacnet-client/src/client/routed_max_apdu_tests.rs` — end-to-end loopback test proving a manually seeded peer limit (128, Segmentation::BOTH) forces segmentation instead of local-config fallback, plus registration round-trip/resolve assertions and a legacy-add_device-stays-local test
- `CHANGELOG.md` — Fixed + Added entries

Python surface intentionally unchanged (`add_routed_device` is Rust-only for now); no `.pyi`/python-docs changes needed.

## Red evidence (base 34c6dbb00c183e8c46bbd5fd2f7a2a0195355176)

Command: `RUSTFLAGS=-Awarnings cargo test -q -p bacnet-client get_by_mac --locked`

- `get_by_mac_ignores_routed_rows` FAILED — `assertion failed: table.get_by_mac(router_mac).is_none()` (the routed peer was returned)
- `get_by_mac_ignores_multiple_routed_rows_behind_one_router` FAILED — same assertion

This deterministically discriminates facet 1 of #372 (router MAC conflated with routed-peer identity). Facet 2's duplicate-nondeterminism is covered as green coverage via the deterministic freshest-row test, per brief (no hash-order-dependent timing test). The manual-registration API tests are compile-time-absent at base, as allowed.

## Green verification (final head 97c26e4e5c14046dec552e781ac818e7cd314fe8)

- `cargo fmt --all -- --check` — pass
- `RUSTFLAGS=-Awarnings cargo test -q -p bacnet-client --locked` — 131 lib + 3 integration passed
- workspace suite w/ features — 3164 passed, 0 failed, 2 ignored (pre-existing loopback-gated)
- `RUSTFLAGS=-Awarnings cargo clippy -q --workspace --exclude rusty-bacnet --all-targets --locked` — pass
- `RUSTFLAGS=-Awarnings cargo check -q -p rusty-bacnet --tests --locked` — pass
- `check-file-size.sh`, `check-no-secrets.sh`, `generate-conformance-docs.py --check`, `git diff --check` — pass

Note: the counts below were produced at `ef40366`; the only later commit (`97c26e4`) is a CHANGELOG.md section restructure with no code effect. GitHub Actions runs green at the final head.

## Compatibility

- Additive public Rust API: `add_routed_device` method + `RoutedDeviceConfig` struct (public fields, `Clone`); existing `add_device` callers unaffected. `add_routed_device` deliberately rejects empty `router_mac`/`remote_mac` with `Error::Encoding` before any table mutation.
- No public struct fields added to `DiscoveredDevice`; only a new inherent method `is_local()`.
- Python: unchanged by design (documented above).
- Wire behavior: none; only which table rows lookups select and what callers may seed.
- Changelog describes identity/registration fixes; no new BACnet feature-support or conformance claim.

## Non-goals honored

#371 segmentation enforcement, #370 internetwork length discovery, #379/#380 reassembly timeouts, #383 duplicate-window, #384 server reassembly keying, DeviceTable replacement, lock-across-I/O, broad discovery redesign — all untouched. Tracked adjacent backlog (#397–#401) untouched.

No benchmark: lookup scans are small linear passes over an in-memory map on an unmeasured path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   detection-disabled reset clears all three, and restart semantics match
   the intrinsic detectors (in-memory only).
 
+### Added
+
+- `BACnetClient::add_routed_device` with a `RoutedDeviceConfig` (#372):
+  callers that already know a routed peer can register its device instance,
+  immediate-hop router MAC, remote SNET/SADR, advertised Max APDU Length
+  Accepted, segmentation capability, and optional Max Segments Accepted,
+  so requests are sized by the peer's advertised limits instead of silently
+  falling back to local config. Vendor ID defaults to 0 when unknown and
+  empty `router_mac`/`remote_mac` registrations are refused. Rust-only for
+  now; the Python surface is unchanged.
+
 ### Fixed
 
 - DeviceTable address identity no longer conflates routers with routed
@@ -100,20 +111,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   them. Both secondary lookups now select the freshest `last_seen` when
   several rows share an address, replacing nondeterministic hash-order
   picks. `resolve_device`'s two-lock snapshot/coherence boundary is now
-  documented on the method.
-
-### Added
-
-- `BACnetClient::add_routed_device` with a `RoutedDeviceConfig` (#372):
-  callers that already know a routed peer can register its device instance,
-  immediate-hop router MAC, remote SNET/SADR, advertised Max APDU Length
-  Accepted, segmentation capability, and optional Max Segments Accepted,
-  so requests are sized by the peer's advertised limits instead of silently
-  falling back to local config. Vendor ID defaults to 0 when unknown.
-  Rust-only for now; the Python surface is unchanged. Legacy
-  `add_device(instance, mac)` still creates local rows and its metadata
-  defaults are now documented as manual registration values rather than
-  advertised peer capabilities.
+  documented on the method. Legacy `add_device(instance, mac)` still
+  creates local rows and its metadata defaults are documented as manual
+  registration values rather than advertised peer capabilities.
 
 - The Escalator object's `operation_direction` (477) is now stored as the
   typed `EscalatorOperationDirection` (default UNKNOWN) instead of a raw
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- DeviceTable address identity no longer conflates routers with routed
+  peers, and duplicate matches resolve deterministically (#372). A local
+  `get_by_mac` lookup now considers only unambiguously local rows: a routed
+  row's `mac_address` is the immediate-hop router (Clause 6.2.2), never the
+  remote device's identity, so requesting the router's own Device object can
+  no longer pick up a peer's limits from behind it. Rows with partial
+  routing metadata match neither lookup until a complete I-Am refreshes
+  them. Both secondary lookups now select the freshest `last_seen` when
+  several rows share an address, replacing nondeterministic hash-order
+  picks. `resolve_device`'s two-lock snapshot/coherence boundary is now
+  documented on the method.
+
+### Added
+
+- `BACnetClient::add_routed_device` with a `RoutedDeviceConfig` (#372):
+  callers that already know a routed peer can register its device instance,
+  immediate-hop router MAC, remote SNET/SADR, advertised Max APDU Length
+  Accepted, segmentation capability, and optional Max Segments Accepted,
+  so requests are sized by the peer's advertised limits instead of silently
+  falling back to local config. Vendor ID defaults to 0 when unknown.
+  Rust-only for now; the Python surface is unchanged. Legacy
+  `add_device(instance, mac)` still creates local rows and its metadata
+  defaults are now documented as manual registration values rather than
+  advertised peer capabilities.
+
 - The Escalator object's `operation_direction` (477) is now stored as the
   typed `EscalatorOperationDirection` (default UNKNOWN) instead of a raw
   `u32` with an invented 0=unknown/1=up/2=down/3=stopped comment mapping
--- a/crates/bacnet-client/src/client/discovery.rs
+++ b/crates/bacnet-client/src/client/discovery.rs
@@ -1,7 +1,19 @@
 use super::*;
+use bacnet_types::enums::Segmentation;
 
 impl<T: TransportPort + 'static> BACnetClient<T> {
     /// Resolve a device instance to its MAC address and optional routing info.
+    ///
+    /// # Snapshot/coherence contract
+    ///
+    /// This returns an address/routing snapshot taken under one
+    /// [`DeviceTable`] lock; the lock is released before any network I/O.
+    /// The request's later capability lookup (Max APDU/segmentation) takes a
+    /// second lock, so a table update between the two reads can make the
+    /// capability decision observe a newer entry than the address snapshot.
+    /// This is an intentional non-atomic snapshot boundary, not a promise
+    /// that both reads use the same row forever; holding the lock across the
+    /// request would serialize all in-flight requests.
     pub(super) async fn resolve_device(
         &self,
         device_instance: u32,
@@ -122,11 +134,14 @@ impl<T: TransportPort + 'static> BACnetClient<T> {
         self.device_table.lock().await.clear();
     }
 
-    /// Manually register a device in the device table.
+    /// Manually register a local device in the device table.
     ///
     /// Useful for adding known devices without requiring WhoIs/IAm exchange.
-    /// Sets default values for max_apdu_length (1476), segmentation (NONE),
-    /// and vendor_id (0) since these are unknown without IAm.
+    /// The metadata defaults below (max_apdu_length 1476, segmentation NONE,
+    /// vendor_id 0) are local/manual registration defaults, **not** values
+    /// learned from an I-Am; requests to this device are bounded by them
+    /// until a real I-Am refreshes the row. The row is unambiguously local:
+    /// routed lookups by SNET/SADR will not match it.
     pub async fn add_device(&self, instance: u32, mac: &[u8]) -> Result<(), Error> {
         let oid = bacnet_types::primitives::ObjectIdentifier::new(
             bacnet_types::enums::ObjectType::DEVICE,
@@ -142,6 +157,38 @@ impl<T: TransportPort + 'static> BACnetClient<T> {
             last_seen: std::time::Instant::now(),
             source_network: None,
             source_address: None,
+        };
+        self.device_table.lock().await.upsert(device);
+        Ok(())
+    }
+
+    /// Manually register a routed peer whose address and capabilities are
+    /// already known, without fabricating an I-Am exchange.
+    ///
+    /// `router_mac` is only the immediate transport next hop (Clause 6.2.2);
+    /// the peer's identity is the (`remote_network`, `remote_mac`) SNET/SADR
+    /// pair, which is what [`BACnetClient::confirmed_request_routed`] and the
+    /// device-table routed lookup match against. The supplied Max APDU
+    /// Length Accepted, segmentation capability, and optional Max Segments
+    /// Accepted are treated as advertised peer limits for request sizing.
+    ///
+    /// Vendor ID is recorded as 0 when unknown; this does not affect
+    /// routing or request-size decisions.
+    pub async fn add_routed_device(&self, config: RoutedDeviceConfig) -> Result<(), Error> {
+        let oid = bacnet_types::primitives::ObjectIdentifier::new(
+            bacnet_types::enums::ObjectType::DEVICE,
+            config.instance,
+        )?;
+        let device = DiscoveredDevice {
+            object_identifier: oid,
+            mac_address: MacAddr::from_slice(&config.router_mac),
+            max_apdu_length: config.max_apdu_length,
+            segmentation_supported: config.segmentation_supported,
+            max_segments_accepted: config.max_segments_accepted,
+            vendor_id: 0,
+            last_seen: std::time::Instant::now(),
+            source_network: Some(config.remote_network),
+            source_address: Some(MacAddr::from_slice(&config.remote_mac)),
         };
         self.device_table.lock().await.upsert(device);
         Ok(())
--- a/crates/bacnet-client/src/client/discovery.rs
+++ b/crates/bacnet-client/src/client/discovery.rs
@@ -174,7 +174,22 @@ impl<T: TransportPort + 'static> BACnetClient<T> {
     ///
     /// Vendor ID is recorded as 0 when unknown; this does not affect
     /// routing or request-size decisions.
+    ///
+    /// Malformed routing metadata is rejected deliberately: an empty
+    /// `router_mac` or `remote_mac` cannot identify a next hop or a peer, so
+    /// such a registration returns [`Error::Encoding`] and the table is left
+    /// unchanged (an empty SADR could never satisfy a routed lookup).
     pub async fn add_routed_device(&self, config: RoutedDeviceConfig) -> Result<(), Error> {
+        if config.router_mac.is_empty() {
+            return Err(Error::Encoding(
+                "router_mac must not be empty: it is the immediate transport next hop".into(),
+            ));
+        }
+        if config.remote_mac.is_empty() {
+            return Err(Error::Encoding(
+                "remote_mac must not be empty: it is half of the routed peer's identity".into(),
+            ));
+        }
         let oid = bacnet_types::primitives::ObjectIdentifier::new(
             bacnet_types::enums::ObjectType::DEVICE,
             config.instance,
--- a/crates/bacnet-client/src/client/mod.rs
+++ b/crates/bacnet-client/src/client/mod.rs
@@ -33,7 +33,7 @@ use bacnet_types::enums::{
 use bacnet_types::error::Error;
 use bacnet_types::MacAddr;
 
-use crate::discovery::{DeviceTable, DeviceUpsertResult, DiscoveredDevice};
+use crate::discovery::{DeviceTable, DeviceUpsertResult, DiscoveredDevice, RoutedDeviceConfig};
 use crate::segmentation::{max_segment_payload, split_payload, SegmentReceiver, SegmentedPduType};
 use crate::tsm::{Tsm, TsmConfig, TsmResponse};
 
--- a/crates/bacnet-client/src/client/routed_max_apdu_tests.rs
+++ b/crates/bacnet-client/src/client/routed_max_apdu_tests.rs
@@ -19,7 +19,7 @@ use bacnet_types::primitives::ObjectIdentifier;
 use bacnet_types::MacAddr;
 use tokio::time::{timeout, Duration};
 
-use crate::discovery::DiscoveredDevice;
+use crate::discovery::{DiscoveredDevice, RoutedDeviceConfig};
 
 use super::tests::send_routed_response;
 use super::BACnetClient;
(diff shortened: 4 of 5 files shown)


jscott3201 (commented):
[Adversary] Reviewed head: 8579fd7cc2a5e60171feace6a5436920f474143c

**Target verification:** Packet base_sha/head_sha match the observed PR exactly (base dev@34c6dbb0, head fix/device-table-address-identity@8579fd7c, single commit). All evidence taken from GitHub refs at these exact SHAs, not the local checkout. **Tooling limitation:** this session exposes no shell tool, so `git rev-parse HEAD` could not be run and item 9 (`cargo test -q -p bacnet-client --locked`) could NOT be executed by this reviewer; the PR body's claimed counts (130 lib + 3 integration green at head; both red tests failing at base) remain unverified by me. Everything else below is verified by direct read of base/head sources.

**Attack-lens results (all eight probes):**

1. Multiple routed peers sharing one router MAC — covered: `get_by_mac_ignores_multiple_routed_rows_behind_one_router` (discovery.rs:383-395). Passes only because `is_local()` filters all three rows out.
2. Address-space cross-match — covered both directions: local MAC byte-equal to a routed SADR (`get_by_mac_returns_local_row_not_routed_sadr_namesake`, discovery.rs:400-409) plus pre-existing `get_by_network_address_ignores_local_devices`. `is_local()` + the `Some(network)` requirement make the spaces disjoint by construction.
3. Stale-vs-fresh symmetry — `get_by_mac_prefers_freshest_local_row` (discovery.rs:412-425) uses distinct `last_seen` values; `max_by_key` over distinct keys is insertion-order-independent, so it passes with either insert order. Not secretly HashMap-order dependent. Caveat at L1 below for exact ties.
4. Partial metadata — tested variant is **source_network-only** (`partial_routing_metadata_is_excluded_from_local_lookup`, discovery.rs:427-437); the source_address-only variant is untested. By inspection the guards are symmetric (`is_local()` requires neither set; `get_by_network_address` requires `Some(network)`), so behavior is identical, but only one half has red/green coverage. See L2.
5. `manual_routed_registration_honors_seeded_peer_limit` — the seeded limit provably binds: local config is 1476, service data 204 octets; the test asserts every received APDU ≤ 128 (`payload.len() <= 128`) AND `req.segmented` (routed_max_apdu_tests.rs:297-310). A local-config fallback would emit one unsegmented 204-octet APDU and fail both assertions. Misroute/wrong-frame regressions are caught structurally: any failure to deliver to `router_rx` hits the 2 s timeout. An extra no-frame guard (as in `routed_confirmed_request_rejects_subminimum_peer_limit`, routed_max_apdu_tests.rs:186-198) is NOT needed here — that test guards a *rejection* path; this test has no rejection path. Informational only.
6. `legacy_add_device_creates_local_row_only` (routed_max_apdu_tests.rs:343-362) asserts BOTH directions: `get_by_mac` finds the row and `get_by_network_address(100, &mac)` returns None. Complete.
7. Red evidence integrity — VERIFIED. Base discovery.rs@34c6dbb0 `get_by_mac` is `self.devices.values().find(|d| d.mac_address.as_slice() == mac)` with no routing-field filter, so both `get_by_mac_ignores_*` tests deterministically fail at base (the routed row matches first). Genuinely discriminating.
8. Panic paths — none found. `MacAddr = smallvec::SmallVec<[u8; 6]>`; `from_slice` copies any length (empty → empty SmallVec, oversized → heap spill), no panic/truncation. `ObjectIdentifier::new(...)?` handles bad instance ranges via `Result`. `u32→u16` max_apdu consumption saturates (`unwrap_or(u16::MAX)`). No unwrap/expect added outside `#[cfg(test)]`.

**Regression sweep:** sole production consumer of `get_by_mac` is the `ConfirmedTarget::Local` arm of `confirmed_request_inner` (requests.rs:150-166); the change makes a local unicast to a router's MAC fall back to local-config sizing instead of a behind-router peer's limits — exactly the intended fix. Routed sizing still resolves via `get_by_network_address`. I-Am ingestion (dispatch.rs) sets routing fields strictly both-or-neither, so partial rows cannot arise from wire input.

**Findings (ranked):**

- **BLOCKER:** none.
- **HIGH:** none.
- **MEDIUM [M1] — `add_routed_device` silently accepts malformed addresses** (crates/bacnet-client/src/client/discovery.rs:164-191). Empty `router_mac` or `remote_mac` register `Ok(())`: an empty `mac_address` produces requests sent to a zero-length destination MAC, and `source_address = Some(empty)` is a routable-looking row that `get_by_network_address(net, &[])` will happily match. Given the method returns `Result<(), Error>`, rejecting zero-length MACs costs nothing; today hostile/host input yields a permanently broken registration with no diagnostic. (Related: `max_apdu_length < 50` is also accepted at registration and only surfaces later as a request-time error — acceptable, but worth documenting.)
- **LOW [L1] — exact `last_seen` ties stay nondeterministic**: `max_by_key` returns the last maximum in HashMap iteration order (discovery.rs:133-140, 155-163). Two registrations inside one clock tick on coarse-resolution platforms resolve arbitrarily. Inherent and mirrored on the routed side; noting for honesty about "deterministic".
- **LOW [L2] — half-set-metadata test covers only `source_network`-set** (discovery.rs:427-437). Add the mirror variant or accept inspection-only assurance.
- **LOW [L3] — doc overstatement**: the invariant doc says a partial row "counts as neither local nor routable", but `resolve_device` maps `(Some, _)/(_, Some)` to `routing = None` and unicasts to its `mac_address` as if local (client/discovery.rs:14-23). Unreachable from wire input today, reachable via public fields.
- **LOW [L4] — stale-purge doc fiction**: "a re-commissioned device instance leaves two rows" (discovery.rs get_by_network_address/get_by_mac docs) cannot occur — upsert keys by instance and replaces; duplicates require two *distinct* instances sharing one address.
- **LOW [L5] — shared instance key flips address space silently**: `add_device(i, …)` over an existing routed row (or vice versa) wholesale replaces it, discarding learned capabilities/vendor_id with no result signal. Pre-existing upsert semantics made more consequential by the second entry point.

**Verified facts vs inference:** findings M1, L1–L4 and all eight probe results are verified against exact head/base file reads; L5's exploitability depends on caller patterns (inference). Test-suite greenness at head is claimed by the PR body but not independently executed (no shell available) — flagged above.

**Verdict: READY** — no blocker/high survives; M1 is a hardening gap in brand-new API surface, reasonable to fix in this PR or immediately after.

jscott3201 (commented):
[Adversary] — re-review of delta 8579fd7cc2a5e60171feace6a5436920f474143c...ef4036673a898ffce06b65f7ee84c8ec6e37d5d4

Reviewed head: ef4036673a898ffce06b65f7ee84c8ec6e37d5d4

Verdict: READY

Target check: PR #402 base is 34c6dbb00c183e8c46bbd5fd2f7a2a0195355176 and head is ef4036673a898ffce06b65f7ee84c8ec6e37d5d4 — both match the review packet; the PR contains exactly the two expected commits. Reviewed via the GitHub PR diff and file reads pinned at head SHA.

Verification caveat: this reviewer's session has no shell-execution capability, so the requested local commands (`git rev-parse HEAD`, `RUSTFLAGS=-Awarnings cargo test -q -p bacnet-client --locked`) could not be run by me; no local counts to report. All evidence below is from GitHub file contents at head `ef40366`. Combined status at head was still `pending` (0 checks reported) at review time.

## Prior Medium — RESOLVED

The prior finding (add_routed_device silently accepted empty router_mac/remote_mac, creating a row no routed lookup could use) is fixed:

- `crates/bacnet-client/src/client/discovery.rs:182-192` (head): both guards run first, returning `Error::Encoding` with a message naming the offending field, before `ObjectIdentifier::new` (which can also fail before any mutation) and before `device_table.lock().await.upsert(device)`. A refused registration provably cannot mutate the table — the lock is not even acquired on the refusal paths.
- Bypass hunt: no bypass found. MACs are `Vec<u8>`/`MacAddr` byte strings, so "whitespace-only" is not a meaningful input class; single-byte MACs are legitimate (MS/TP/loopback-style) and correctly remain accepted. The `Error::Encoding(String)` variant matches this crate's established convention for client-side validation refusals (same variant and style used in `client/mod.rs` `ClientOptions::validate`, `validate_identity`, `cap_max_apdu_to_transport`, and `resolve_device` in `discovery.rs`).
- Doc claim at `discovery.rs:178-181` matches the implemented behavior.

## Test quality — pins both variants and the no-row guarantee

`crates/bacnet-client/src/client/routed_max_apdu_tests.rs:350-381`: `manual_routed_registration_rejects_empty_addresses` iterates `(empty router_mac, non-empty remote_mac)` and `(non-empty router_mac, empty remote_mac)`, asserting `is_err` on each. A regression accepting variant 1 fails the in-loop assert; a regression accepting variant 2 leaves a row at instance 3003 and fails the closing `get(3003).is_none()` assertion. Both variants and the no-row guarantee are genuinely pinned.

## Findings

LOW — test does not pin the error variant or per-field message. `routed_max_apdu_tests.rs:372-373` asserts only `result.is_err()`. A regression that rejects both variants for an unrelated reason, or that swaps the two error strings (router_mac message on the remote_mac check and vice versa), still passes. Suggest `matches!(err, Error::Encoding(_))` plus asserting the message names `router_mac` vs `remote_mac`. Non-blocking.

LOW (scope note, not a defect) — non-empty all-zero addresses still register: `router_mac: vec![0;6]` / `remote_mac: vec![0;6]` (the BACnet/SC reserved UNKNOWN_VMAC spelling) are accepted. Defensible as deliberate scope (an all-zero link MAC is medium-dependent and rejecting it risks false refusals); noting for the record.

NIT — `discovery.rs:181` claims "an empty SADR could never satisfy a routed lookup". Strictly, `DeviceTable::get_by_network_address` compares stored SADR equality, so an empty-SADR row would match a lookup that itself passes empty MAC bytes (e.g., `confirmed_request_routed` with an empty `dest_mac`). The new guard makes such a row unconstructible through this API, so the claim holds in practice; wording only.

## Regression check

The delta adds two early-return guards and one test; no existing path changed, so no earlier finding can regress. No in-repo caller passes empty addresses to `add_routed_device` (the API is new in this PR; only tests invoke it).

## Test run

Could not be executed in this session (no shell tool available). Statically, the new test compiles against imports already present in the test file (`Segmentation`, `RoutedDeviceConfig`, `MacAddr`). Recommend relying on the PR's CI green run for `RUSTFLAGS=-Awarnings cargo test -q -p bacnet-client --locked` before merge.

READY for head ef4036673a898ffce06b65f7ee84c8ec6e37d5d4.

jscott3201 (commented):
[Dataflow] Reviewed head: ef4036673a898ffce06b65f7ee84c8ec6e37d5d4 (base 34c6dbb00c183e8c46bbd5fd2f7a2a0195355176, packet SHAs match PR head/base via API; local `git rev-parse` not available in this environment — no shell tool — so the authoritative PR head/base SHAs from the GitHub API were used; both match the packet).

Verdict: READY (no code/dataflow defects; one docs-structure finding and two low doc nits below).

## Checklist verification (all against head ef403667)

1. **Local/routed spaces cannot cross-match — confirmed.** `get_by_mac` filters `DiscoveredDevice::is_local()` (both routing fields `None`, crates/bacnet-client/src/discovery.rs:33-38) before the MAC comparison and takes `max_by_key(last_seen)` (discovery.rs:139-145). A routed row always has at least one routing field set, so no input MAC can ever select it. `get_by_network_address` requires `source_network == Some(network)` **and** `source_address` byte-match (discovery.rs:166-177), so local rows (both `None`) and partial rows (one `None`) can never satisfy it. Cross-space namesakes (local MAC byte-equal to a routed SADR, and vice versa) are pinned by tests `get_by_mac_returns_local_row_not_routed_sadr_namesake` and `get_by_network_address_ignores_local_devices`. Partial-metadata rows match neither lookup (test `partial_routing_metadata_is_excluded_from_local_lookup`); such rows are unreachable via public constructors (I-Am sets both together; `add_routed_device` sets both or errors).

2. **SNET/SADR identity preserved — confirmed.** `add_routed_device` stores `router_mac` in `mac_address` (next hop only) and `remote_network`/`remote_mac` in `source_network`/`source_address` (client/discovery.rs:178-208). `git`-equivalent check via commit file lists: neither PR commit (8579fd7, ef40366) touches `crates/bacnet-client/src/client/dispatch.rs` — the dispatch/I-Am path is byte-unchanged, so routed rows continue to record router MAC + SNET/SADR exactly as the lookup now assumes.

3. **Freshest-row selection — applied in both lookups; tie stance is the one residual gap (low, below).** Duplicate creation requires two *distinct* instances (same-instance `upsert` replaces the row, discovery.rs:96-110), and `last_seen` is `Instant::now()` taken under the table lock (monotonic), so distinct-timestamp duplicates resolve deterministically to the freshest row.

4. **Capability lookup uses the intended row — confirmed.** `confirmed_request_inner` (crates/bacnet-client/src/client/requests.rs:181-199) resolves Local via the now-local-only `get_by_mac` and Routed via `get_by_network_address`; any found row classifies as `LengthBoundedBy::DiscoveredPeer`. A manually seeded routed peer is found by its SNET/SADR and its seeded 128-octet limit forces segmentation (end-to-end test `manual_routed_registration_honors_seeded_peer_limit`, client/routed_max_apdu_tests.rs:216-296) instead of local-config fallback — exactly the #372 facet-3 fix. Legacy `add_device` rows stay local-only (test `legacy_add_device_creates_local_row_only`). Classification remains correct: a caller-supplied `RoutedDeviceConfig.max_apdu_length` genuinely represents the peer's advertised limit.

5. **`resolve_device` coherence rustdoc — present and accurate.** client/discovery.rs:4-17 documents the two-lock snapshot boundary (address under one lock released before I/O; capability lookup re-locks and may observe a newer row), matching the code and issue #372's explicit "document rather than fix" allowance for facet 4.

6. **Scope — clean.** Diff touches exactly CHANGELOG.md, crates/bacnet-client/src/discovery.rs, crates/bacnet-client/src/client/discovery.rs, crates/bacnet-client/src/client/mod.rs, crates/bacnet-client/src/client/routed_max_apdu_tests.rs. No unrelated changes.

## Findings

**MEDIUM (docs-only, no runtime impact) — CHANGELOG heading placement recategorizes prior Fixed entries as Added.**
CHANGELOG.md:107 inserts a new `### Added` heading between the new #372 Fixed entry (CHANGELOG.md:91-105) and the pre-existing Fixed entries. At base 34c6dbb, the Escalator entry (#284) and everything after it (File_Access_Method #287, Audit #345, COV Multiple #342, GetEnrollmentSummary #358, #374, #365, #375, #364, #367, #368, #381, #187, #186, #360, #366, #362, #205, #209/#289, #163, #166, #137, #182, #252, #235, #192, the breaking bit-string fix #203) sat under `### Fixed`; at head they now all fall under this new `### Added` heading (starting with the Escalator entry at CHANGELOG.md:120). Observable consequence: release notes misrepresent ~25 fixes — several of them breaking changes — as additions. Fix is mechanical: move the `add_routed_device` entry into the existing top-level `### Added` section (CHANGELOG.md:7) or relocate the new heading below the Fixed list. Does not block the code change.

**LOW — `max_by_key` tie behavior is undocumented and remains hash-order dependent.**
`Iterator::max_by_key` returns the *last* maximum in iteration order (discovery.rs:144 and :176), and `HashMap` iteration order is randomized per process, so two rows sharing both the address and a byte-equal `last_seen` would still resolve nondeterministically — the exact symptom #372 facet 2 describes. I could not construct a reachable concrete trigger: `last_seen` is `Instant::now()` under the lock (monotonic, sub-microsecond on all supported platforms), duplicates require distinct instances, and same-instance upsert replaces the row — so exact ties require two distinct instances learned within one clock tick. Per #372's "define duplicate resolution" the rule is defined (freshest wins) but the tie case has no documented stance. Suggest one sentence ("ties resolve arbitrarily; timestamps are monotonic so ties are not produced in practice") or a deterministic tiebreak on instance number.

**LOW (docs) — `LengthBoundedBy::DiscoveredPeer` wording vs manually seeded rows.**
The variant doc says "A length advertised by a peer, from I-Am" (requests.rs:11-12) and the floor error says "the peer's advertised Max APDU Length Accepted" (requests.rs:73-75), but manually registered rows now also classify as `DiscoveredPeer` — including `add_device`'s 1476 default, which the new doc at client/discovery.rs:141-146 explicitly says is *not* I-Am-learned. Behavior is correct and unchanged (pre-existing classification); only the error/enum wording can misattribute a manual default as an advertisement in the sub-50-octet error path, which `add_device`'s hardcoded 1476 cannot reach. Cosmetic.

No blocker or high findings: the identity invariants hold for every reachable input, the routed peer's limits govern request sizing in both discovery and manual-registration paths, dispatch is untouched, and the diff is exactly the declared scope.

Issue #383, opened 2026-08-17
Segmented receive paths NAK the first duplicate; DuplicateInWindow/DuplicateCount/Ndup (Clauses 5.4.2.2, 5.4.5.2, 5.4.4.4) are unimplemented


jscott3201:
### Affected revision

`dev` after PR #382. Scoped out of that PR by its review.

### Behavior

The server's segmented-request receive path answers the **first** duplicate or out-of-order segment with a negative SegmentAck (`lifecycle.rs`, the `seq != state.expected_seq` arm). Clause 5.4.5.2 distinguishes the two cases via `DuplicateInWindow` (Clause 5.4.2.2):

- `DuplicateSegmentReceived` (in the duplicate window, `DuplicateCount < Ndup`): "discard the ... segment; restart SegmentTimer; increment DuplicateCount" — **silently**.
- `TooManyDuplicateSegmentsReceived` (`DuplicateCount == Ndup`): NAK, reset the count.
- `SegmentReceivedOutOfOrder` (outside the duplicate window): NAK immediately.

The implementation has no `DuplicateCount`/`Ndup` and no `DuplicateInWindow`, so every non-in-order segment draws an immediate NAK. The client-side receive path (`segmentation.rs`) has the same shape: duplicates are acked immediately rather than counted.

### Why this is only chatty, not wrong, today

Since PR #382 the NAK carries the last accepted sequence number and peers advance past it (Clause 5.4.4.2 semantics), so an early NAK moves the transfer forward correctly — it is more traffic than the Standard wants, not a stall. At window size 1 the spec itself NAKs every duplicate (`DuplicateInWindow` is FALSE there), so the deviation is observable only at window sizes above 1.

### Suggested direction

Implement `DuplicateInWindow(seqA, first, last)` per Clause 5.4.2.2 in `bacnet-encoding`'s segmentation module, add `DuplicateCount` to both receive states with a shared `Ndup` (the Standard's device constant), and route the duplicate/out-of-order split through it on both server and client receive paths.

Related: #364, #367 (PR #382).

Issue #379, opened 2026-08-17
Device Max_Segments_Accepted defaults to 65 under SEGMENTED_TRANSMIT where Clause 12.11 requires 1


jscott3201:
### Affected revision

`dev` at `3a0bc93`. Found during grounding for #367.

### Behavior

The client-side reassembly reaper (`crates/bacnet-client/src/client/lifecycle.rs`, ~95-129) deviates from Clause 5.4.4.4 `Timeout` in four ways:

1. **Interval**: `SEG_RECEIVER_TIMEOUT` is a hardcoded 4 s (`client/mod.rs` ~359). The spec's SEGMENTED_CONF bound is `Tseg × 4`, where Tseg is APDU_Segment_Timeout — 24 s at common defaults. The reaper is ~6× too aggressive and config-blind, so it can evict a healthy slow transfer whose inter-segment gaps exceed 4 s.
2. **Wrong action**: 5.4.4.4 `Timeout` (extraction ~3047-3051) prescribes ONLY "send ABORT.indication with 'server' = FALSE and 'abort-reason' = TSM_TIMEOUT to the local application program" — no PDU transmission. The reaper instead transmits an Abort PDU to the peer.
3. **Missing action**: the reaper never completes or cancels the TSM transaction, so the caller is left to time out on its own instead of receiving the TSM_TIMEOUT abort indication.
4. **Trigger**: the reap loop runs only inside the `apdu_rx.recv()` arm, so a peer that simply goes quiet (and no other inbound traffic arrives) leaves the session in place indefinitely — `routed_reply_tests.rs` documents this ("The reaper runs on inbound traffic").

### Suggested direction

Derive the interval from configuration (an `apdu_segment_timeout_ms` if introduced, else `apdu_timeout_ms`), complete the TSM transaction with `TsmResponse::Abort { TSM_TIMEOUT }` on eviction, drop the peer-directed Abort transmission (or keep it as a documented local-matter courtesy), and drive reaping from a timer as well as inbound traffic.

Related: #367 (sessions outliving their transaction — the opposite defect).

Issue #370, opened 2026-08-17
Clause 5.2.1.2(b) is bounded by the local egress link only: intervening-network NPDU limits are never enforced


jscott3201:
Surfaced by spec grounding for #362 / PR #369; pre-existing and unchanged by that fix.

### Behavior

`crates/bacnet-client/src/client/requests.rs`, `target_transport_max_apdu_length` derives the transport term of the Clause 5.2.1.2 minimum from `self.network.transport().max_apdu_length()` minus the routed NPDU header. That is the **local egress** data link only. For a routed target, no term accounts for the data links of the destination network or any intervening network.

A client on BACnet/IP (1476) sending to a peer behind a router whose far side is LonTalk (Table 6-1: 228-octet NPDU) will emit up to 1476 octets in one APDU; the router can only drop it or reject it.

### Spec (live-extracted text, `pdftotext -layout`, 92,279 lines)

- Term (b), lines 2051–2053: the maximum APDU size conveyable "is constrained by the maximum NPDU length permitted by the data links used by the local, remote, and **any intervening networks**, as specified in Clause 6".
- Clause 6, lines 4156–4160 (normative): "The maximum NPDU length **shall not exceed** the capability of any data link technology encountered along the path from source to destination." Table 6-1 (lines 4162–4185) lists per-technology maxima; the interesting rows for this client are LonTalk 228, ARCNET/PTP/ZigBee 501.
- Clause 19.4 (lines 51586–51683) is the sanctioned discovery procedure (probe with segmented writes, back off on Reject-Message-To-Network reason 4 — line 4743: "The message is too long to be routed to this DNET."), with a **normative caching requirement** at lines 51682–51683.
- Units caveat: Table 6-1 is NPDU octets; term (b) is an APDU bound. The spec's own conversion example is line 52662 ("1476 octets ... fits in a 1497 octet NPDU"), i.e. NPCI headroom must be subtracted before comparing.

### Expected work (directional)

Not necessarily full Clause 19.4 probing. A defensible minimum:

- react to Reject-Message-To-Network reason 4 by failing the request with a typed error naming the DNET rather than timing out;
- optionally, a per-DNET conveyable-length cache (fed by reason-4 rejects and/or an explicit configuration knob) folded into the same `check_transmittable_length` minimum as a fourth `LengthBoundedBy` term, so the error names the path when the path binds.

If Clause 19.4 probing is ever implemented, its cache is mandatory, not optional.

### Related

[...]

PR #403, opened 2026-08-21
fix(client): honor peer receive-segmentation capability

### LINKED ISSUE #371: Client segments confirmed requests to peers whose I-Am advertised NO_SEGMENTATION or SEGMENTED_TRANSMIT
Surfaced by the code sweep for #362 / PR #369; pre-existing on the local path and now equally reachable on the routed path.

### Behavior

`crates/bacnet-client/src/client/dispatch.rs` records `segmentation_supported` from every I-Am into the device table, and `confirmed_request_inner` (`requests.rs`) now consults that same table row for `max_apdu_length` on both paths — but nothing ever reads `segmentation_supported`. A request larger than the peer's Max APDU is segmented and sent regardless of whether the peer said it can receive segments.

For a peer that advertised `NO_SEGMENTATION` or `SEGMENTED_TRANSMIT`, the outcome is predetermined: the peer aborts. Extracted Standard lines 50937–50938: "SEGMENTATION_NOT_SUPPORTED - Generated in response to an APDU that has its segmentation bit set to TRUE when the receiving device does not support segmentation." The client learns nothing it did not already have in its own table, and the failure surfaces one round-trip later than it needs to.

### Spec

Clause 12.11 (Device object), extracted lines 14700–14712:

- `Segmentation_Supported` enumerates `{SEGMENTED_BOTH, SEGMENTED_TRANSMIT, SEGMENTED_RECEIVE, NO_SEGMENTATION}`.
- "If the Segmentation_Supported property of the Device object has a value of SEGMENTED_TRANSMIT or NO_SEGMENTATION, this [Max_Segments_Accepted] property shall have a value of 1."

So the spec's own coupling says a peer in either of those states accepts exactly one segment — i.e. only unsegmented requests. Since I-Am carries `Segmentation Supported` but not `Max_Segments_Accepted`, the capability enum is the only advance signal the client has, and it is currently discarded at the decision point.

### Expected work

- In the unified length block of `confirmed_request_inner`: when `unsegmented_apdu_size > remote_max_apdu` and the discovered peer's `segmentation_supported` is `NO_SEGMENTATION` or `SEGMENTED_TRANSMIT`, return a typed error naming the peer's advertised capability instead of entering `segmented_confirmed_request`.
- Unknown peers (no table row) keep today's behavior — there is no advertised capability to honor, and the peer's Abort remains the discovery mechanism.
- Test: seed a peer with `max_apdu_length: 50, segmentation_supported: NONE`, send an oversized request, assert the typed error and that no frame leaves (50 ms recv-timeout idiom), on both the local and routed paths.

### Related

#362, #369 (the lookup this decision would ride on), #319 (tiny-limit empty-request segmentation).

jscott3201:
Closes #371

## Summary

`confirmed_request_inner` consulted the peer's Max APDU Length Accepted but discarded `Segmentation_Supported`, so a request larger than the peer's limit was segmented and sent regardless of capability. For a peer that advertised `NO_SEGMENTATION` or `SEGMENTED_TRANSMIT`, the Clause 18 SEGMENTATION_NOT_SUPPORTED abort was predetermined one round-trip later. The client now refuses locally — before transaction allocation or transmission — when the capability is authoritative.

## Standard anchors (kept light)

- **Clause 12.11** — Segmentation_Supported distinguishes BOTH/TRANSMIT/RECEIVE/NONE; TRANSMIT and NONE peers accept exactly one segment (only unsegmented requests).
- **Clause 18** — SEGMENTATION_NOT_SUPPORTED is the certain outcome of sending a segmented APDU to such a peer; refusing locally surfaces the known outcome immediately.
- **Clause 5.2.1.2** — the existing maximum-transmittable-APDU calculation is unchanged; this PR only gates entry into the segmented-send path.

## Capability provenance (Round 3 review constraint)

| Source | Authoritative? | Behavior |
|---|---|---|
| I-Am ingestion (`upsert_with_result`) | yes | honored |
| `add_routed_device(RoutedDeviceConfig)` | yes | honored |
| explicit `DeviceTable::upsert` | yes | honored |
| legacy `add_device(instance, mac)` | no — placeholder | never preflight-rejects |
| no matching row | no | current behavior preserved |

Provenance is a private `PeerSegmentation` marker stored in the same internal table entry as the device data — no public struct fields added, no inference from field values (NONE/1476/vendor 0 are all legitimately producible by a real I-Am). Any replacement of an instance replaces its provenance; purge and clear cannot orphan it. Address-space identity rules from #402 are untouched.

## Refusal rule

Only when segmentation is already required by the existing sizing calculation AND the selected row's capability is authoritative NONE or TRANSMIT: `Error::Segmentation` naming the advertised capability, before TSM invoke-ID allocation, before any NPDU/APDU, before the segmented loop. RECEIVE/BOTH proceed; unknown raw capability values proceed (nothing invents receive capability or a prohibition); `max_segments_accepted` is untouched.

## Files changed

- `crates/bacnet-client/src/discovery.rs` — private `TableEntry` + `PeerSegmentation` provenance, `upsert_placeholder`, `local_peer_segmentation`/`routed_peer_segmentation`, provenance-lifecycle test
- `crates/bacnet-client/src/client/requests.rs` — preflight in the unified sizing block
- `crates/bacnet-client/src/client/discovery.rs` — `add_device` uses placeholder provenance
- `crates/bacnet-client/src/client/peer_segmentation_tests.rs` (new) — 8 request-path tests (matrix below)
- `crates/bacnet-client/src/client/mod.rs` — test module registration
- `CHANGELOG.md` — Fixed entry

## Red evidence (base 758ee906924b7211cc94002e96535e9d6ab31b2b)

Command: `RUSTFLAGS=-Awarnings cargo test -q -p bacnet-client peer_segmentation --locked` → 4 passed, 4 failed.

Failed (the discriminators — base segmented and transmitted instead of refusing locally):

- `authoritative_local_no_segmentation_refuses_segmented_request`
- `authoritative_local_transmit_only_refuses_segmented_request`
- `authoritative_routed_no_segmentation_refuses_segmented_request`
- `explicit_upsert_refresh_makes_placeholder_authoritative`

The failures prove #371 specifically: each asserts the typed `Error::Segmentation` naming the capability plus wire silence; base instead entered the segmented-send path. The four passing tests (RECEIVE segments, fitting request sends, legacy placeholder does not refuse, unknown raw value does not masquerade) already passed at base — they are regression guards, not discriminators. The provenance primitive tests are compile-time-absent at base (allowed).

## Test matrix coverage

A. authoritative local NONE → typed refusal, no frame ✓ (red)
B. authoritative local TRANSMIT → typed refusal, no frame ✓ (red)
C. authoritative routed NONE via `add_routed_device` → typed refusal, no frame to router ✓ (red)
D. authoritative RECEIVE → full segmented exchange completes ✓ (BOTH already covered by the preserved Round 3 test)
E. fitting request to NONE peer → unsegmented, completes ✓
F. legacy `add_device` placeholder → 1600-octet request enters and completes the segmented path, no local refusal ✓
G/H. placeholder refreshed by explicit authoritative upsert → oversized now refused locally ✓ (red)
I. authoritative unknown raw capability (from_raw(200)) → not treated as a prohibition; segmented path proceeds ✓

## Verification (final head 5e1953cd07011441017833544c29c865b8f05df0)

- `cargo fmt --all -- --check` — pass
- `RUSTFLAGS=-Awarnings cargo test -q -p bacnet-client --locked` — 140 lib + 3 integration passed
- workspace suite w/ features — 3173 passed, 0 failed, 2 ignored (pre-existing loopback-gated)
- `RUSTFLAGS=-Awarnings cargo clippy -q --workspace --exclude rusty-bacnet --all-targets --locked` — pass
- `RUSTFLAGS=-Awarnings cargo check -q -p rusty-bacnet --tests --locked` — pass
- file-size / no-secret / conformance `--check` / `git diff --check` — pass

Note: test/clippy/fmt counts were produced at `3c2728a`; the only later commit (`5e1953c`) removes an unused test import and fixes a private intra-doc link — no behavioral change. GitHub Actions runs green at the final head.

## Compatibility

- Rust API: no public API change. Private `DeviceTable` storage refactor keeps `upsert`, `upsert_with_result`, `all`, `get`, secondary lookups, `purge_stale*`, `clear` contracts intact (existing tests unchanged and passing). `DiscoveredDevice`/`RoutedDeviceConfig` fields untouched.
- Python: unchanged; no stub or docs changes.
- Wire behavior: only that a known no-receive peer now gets a local refusal instead of guaranteed remote abort. Unknown peers, placeholder peers, receive-capable peers: unchanged.
- Changelog: describes refusal only when receive inability is known; explicitly states placeholder/unknown peers are not refused; no conformance/PICS/BIBB claim.

## Non-goals honored

#370 path limits, #379/#380 timers, #383 duplicate-window, #384 server reassembly, equal-last_seen tie-break, #397–#401 backlog, server segmentation, discovery-table redesign, Python exposure, new CI — all untouched.

No benchmark: constant-time capability check on the existing sizing path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- The client no longer segments confirmed requests to peers whose
+  authoritative `Segmentation_Supported` says they cannot receive segments
+  (#371). When request sizing requires segmentation and the selected
+  device-table row's capability is authoritative — learned from the peer's
+  I-Am or explicitly supplied via `add_routed_device` or an explicit
+  `DeviceTable::upsert` — a `NO_SEGMENTATION` or `SEGMENTED_TRANSMIT` peer
+  now produces a local `Error::Segmentation` naming the advertised
+  capability before any frame is sent or transaction allocated, instead of
+  the guaranteed Clause 18 SEGMENTATION_NOT_SUPPORTED abort. Legacy manual
+  `add_device` rows keep their previous behavior: their capability fields
+  are documented placeholders, so unknown capability never causes a local
+  refusal, and a later I-Am refresh makes the row authoritative. Receive-
+  capable (`SEGMENTED_RECEIVE`/`SEGMENTED_BOTH`) peers segment exactly as
+  before.
+
 - DeviceTable address identity no longer conflates routers with routed
   peers, and duplicate matches resolve deterministically (#372). A local
   `get_by_mac` lookup now considers only unambiguously local rows: a routed
--- a/crates/bacnet-client/src/client/discovery.rs
+++ b/crates/bacnet-client/src/client/discovery.rs
@@ -158,7 +158,9 @@ impl<T: TransportPort + 'static> BACnetClient<T> {
             source_network: None,
             source_address: None,
         };
-        self.device_table.lock().await.upsert(device);
+        // Placeholder capability defaults (NONE, 1476) are not I-Am evidence;
+        // they must never drive a segmentation-capability decision (#371).
+        self.device_table.lock().await.upsert_placeholder(device);
         Ok(())
     }
 
--- a/crates/bacnet-client/src/client/mod.rs
+++ b/crates/bacnet-client/src/client/mod.rs
@@ -797,6 +797,8 @@ mod device_events_tests;
 #[cfg(test)]
 mod peer_max_apdu_tests;
 #[cfg(test)]
+mod peer_segmentation_tests;
+#[cfg(test)]
 mod response_correlation_tests;
 #[cfg(test)]
 mod routed_max_apdu_tests;
--- a/crates/bacnet-client/src/client/peer_segmentation_tests.rs
+++ b/crates/bacnet-client/src/client/peer_segmentation_tests.rs
@@ -1,0 +1,602 @@
+//! The client must not knowingly segment a confirmed request to a peer whose
+//! authoritative Segmentation_Supported says it cannot receive segments.
+//!
+//! Clause 12.11: NO_SEGMENTATION and SEGMENTED_TRANSMIT peers accept exactly
+//! one segment — only unsegmented requests. Clause 18: sending anyway yields
+//! the peer's SEGMENTATION_NOT_SUPPORTED abort, so the client refuses locally
+//! before allocating a transaction or transmitting (issue #371).
+//!
+//! Capability provenance matters: only I-Am-learned or explicitly supplied
+//! capability is authoritative. Legacy `add_device` rows store placeholder
+//! defaults and must not cause a local refusal (#372 follow-up constraint).
+//!
+//! New file: `tests.rs` is over the 700-LOC cap.
+
+use std::time::Instant;
+
+use bacnet_encoding::apdu::{self, Apdu, SegmentAck as SegmentAckPdu, SimpleAck};
+use bacnet_encoding::npdu::{decode_npdu, Npdu};
+use bacnet_transport::loopback::LoopbackTransport;
+use bacnet_transport::port::TransportPort;
+use bacnet_types::enums::{ConfirmedServiceChoice, ObjectType, Segmentation};
+use bacnet_types::error::Error;
+use bacnet_types::primitives::ObjectIdentifier;
+use bacnet_types::MacAddr;
+use bytes::{Bytes, BytesMut};
+use tokio::time::{timeout, Duration};
+
+use crate::discovery::{DiscoveredDevice, RoutedDeviceConfig};
+
+use super::BACnetClient;
+
+const OVERSIZED: usize = 200;
+const PEER_MAX_APDU: u32 = 128;
+
+fn local_peer(
+    instance: u32,
+    mac: &[u8],
+    max_apdu_length: u32,
+    segmentation_supported: Segmentation,
+) -> DiscoveredDevice {
+    DiscoveredDevice {
+        object_identifier: ObjectIdentifier::new(ObjectType::DEVICE, instance).unwrap(),
+        mac_address: MacAddr::from_slice(mac),
+        max_apdu_length,
+        segmentation_supported,
+        max_segments_accepted: None,
+        vendor_id: 42,
+        last_seen: Instant::now(),
+        source_network: None,
+        source_address: None,
+    }
+}
+
+fn oversized_data(len: usize) -> Vec<u8> {
+    (0..len as u16).map(|i| i as u8).collect()
+}
+
+/// Assert the error is the typed local segmentation refusal naming the
+/// peer's advertised capability.
+fn assert_segmentation_unsupported(err: Error, capability: &str) {
+    let text = err.to_string();
+    assert!(
+        matches!(err, Error::Segmentation(_)),
+        "expected Error::Segmentation, got: {text}"
+    );
+    assert!(
+        text.contains(capability),
+        "refusal must name the advertised capability {capability}, got: {text}"
+    );
+}
+
+/// A refusal must leave the wire silent: no frame reaches the peer.
+async fn assert_no_frame_reaches(
+    peer_rx: &mut tokio::sync::mpsc::Receiver<bacnet_transport::port::ReceivedNpdu>,
+) {
+    // A closed channel (the request task dropped its transport end) is not a
+    // transmission; only a delivered NPDU fails this assertion.
+    match timeout(Duration::from_millis(50), peer_rx.recv()).await {
+        Err(_) | Ok(None) => {}
+        Ok(Some(received)) => panic!(
+            "no frame may be transmitted for a locally refused request; got npdu {:x?}",
+            received.npdu
+        ),
+    }
+}
+
+async fn send_local_reply(peer_transport: &LoopbackTransport, client_mac: &[u8], apdu: Apdu) {
+    let mut apdu_buf = BytesMut::new();
+    bacnet_encoding::apdu::encode_apdu(&mut apdu_buf, &apdu).expect("valid APDU encoding");
+    let npdu = Npdu {
+        payload: apdu_buf.freeze(),
+        ..Npdu::default()
+    };
+    let mut npdu_buf = BytesMut::new();
+    bacnet_encoding::npdu::encode_npdu(&mut npdu_buf, &npdu).unwrap();
+    peer_transport
+        .send_unicast(&npdu_buf, client_mac)
+        .await
+        .unwrap();
+}
+
+struct ScriptedExchange {
+    all_service_data: Vec<u8>,
+    every_apdu_segmented: bool,
+}
+
+/// Drive one segmented send to completion against a scripted peer that
+/// acknowledges each segment (window 1) and finally SimpleAcks.
+async fn script_segmented_exchange(
+    peer_rx: &mut tokio::sync::mpsc::Receiver<bacnet_transport::port::ReceivedNpdu>,
+    peer_transport: &LoopbackTransport,
+    client_mac: &[u8],
+    service_choice: ConfirmedServiceChoice,
+) -> ScriptedExchange {
+    let mut all_service_data = Vec::new();
+    let mut every_apdu_segmented = true;
+    let invoke_id = loop {
+        let received = timeout(Duration::from_secs(2), peer_rx.recv())
+            .await
+            .expect("timed out waiting for a segment")
+            .expect("peer channel closed");
+        let npdu = decode_npdu(received.npdu).unwrap();
+        let decoded = apdu::decode_apdu(npdu.payload).unwrap();
+        let Apdu::ConfirmedRequest(req) = decoded else {
+            panic!("Expected ConfirmedRequest, got {decoded:?}");
+        };
+        if !req.segmented {
+            every_apdu_segmented = false;
+        }
+        let seq = req.sequence_number.unwrap();
+        all_service_data.extend_from_slice(&req.service_request);
+
+        let seg_ack = Apdu::SegmentAck(SegmentAckPdu {
+            negative_ack: false,
+            sent_by_server: true,
+            invoke_id: req.invoke_id,
+            sequence_number: seq,
+            actual_window_size: 1,
+        });
+        send_local_reply(peer_transport, client_mac, seg_ack).await;
+
+        if !req.more_follows {
+            break req.invoke_id;
+        }
+    };
+
+    send_local_reply(
+        peer_transport,
+        client_mac,
+        Apdu::SimpleAck(SimpleAck {
+            invoke_id,
+            service_choice,
+        }),
+    )
+    .await;
+
+    ScriptedExchange {
+        all_service_data,
+        every_apdu_segmented,
+    }
+}
+
+// ──────────────────────────────────────────────────────────────────────────
+// Authoritative refusals (red discriminators against #371's base)
+// ──────────────────────────────────────────────────────────────────────────
+
+#[tokio::test]
+async fn authoritative_local_no_segmentation_refuses_segmented_request() {
+    let client_mac = vec![0x01];
+    let peer_mac = vec![0x02];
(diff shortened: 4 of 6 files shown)


jscott3201 (commented):
[Adversary]
Reviewed head: 3c2728a9a756d033967e148a7f61e4ea9fc0730b

VERDICT: READY (no blocker or high findings; low-severity notes below, plus one verification caveat).

Target integrity: PR head/base match the packet exactly (head `3c2728a9…`, base `758ee906…` on `dev`). Review performed against GitHub content pinned to those SHAs.

Verification caveat (must-read): this reviewer session had no shell tool, so `git rev-parse HEAD` and `RUSTFLAGS=-Awarnings cargo test -q -p bacnet-client --locked` could NOT be executed; the PR-reported counts (140 lib + 3 integration passed) are unverified by me. All findings below are from static inspection of exact-SHA file contents. Everything else checked out.

Attack results, per requested vector:

1. Legacy placeholder false-positive — REFUTED as a defect. `legacy_add_device_placeholder_does_not_refuse_segmented_request` genuinely completes end-to-end: `script_segmented_exchange` loops until `more_follows == false`, asserts `all_service_data == expected` (1600 octets) AND `every_apdu_segmented` (so a single unsegmented frame cannot satisfy it), sends SimpleAck, and then asserts the client task returned `Ok(empty)` — proving the TSM actually consumed the ack, not just that segment 0 left the wire (`peer_segmentation_tests.rs`, fn `script_segmented_exchange` + fn `legacy_add_device_placeholder_…`). 1604 > min(1476, transport) guarantees the segmented branch is entered.

2. Authoritative NONE on both paths — holds. Local refusal test seeds only via explicit `DeviceTable::upsert`; routed test seeds only via public `add_routed_device(RoutedDeviceConfig{segmentation_supported: NONE})`; neither depends on the other's seeding. The I-Am ingestion call site still routes through `upsert_with_result` → `PeerSegmentation::Authoritative` (`client/dispatch.rs`, I_AM arm: `device_table.lock().await.upsert_with_result(device.clone())`); `discovery.rs::segmentation_provenance_tracks_row_lifecycle` covers that primitive directly (Placeholder→Authoritative transition), though it does not drive a real I-Am frame through dispatch (see Low-2).

3. TRANSMIT vs RECEIVE — correct directions both ways. `authoritative_local_transmit_only_refuses_segmented_request` refuses with "TRANSMIT" in the message; `authoritative_receive_peer_still_segments_successfully` completes a full segmented exchange against a RECEIVE peer. The BOTH claim resolves to the preserved Round 3 tests in `routed_max_apdu_tests.rs` (`routed_confirmed_request_honors_discovered_peer_max_apdu`, `manual_routed_registration_honors_seeded_peer_limit`) — both seed authoritative BOTH rows and assert every APDU segmented plus byte-exact reassembly and SimpleAck completion.

4. Local/routed symmetry — both refuse independently; the shared preflight sits in `confirmed_request_inner` after sizing and before TSM allocation/invoke-ID (`requests.rs:212-233`, diff hunk @@ -179). Coherence holds: `local_peer_segmentation`/`routed_peer_segmentation` run under the same `dt` lock guard as the max-APDU lookup, and re-resolve the same instance the secondary lookup selected, so provenance cannot diverge from the row used for sizing.

5. Unknown raw value — `Segmentation::from_raw(200)` cannot be refused: `bacnet_enum!` generates a newtype where `matches!(cap, NONE | TRANSMIT)` value-compares against consts 3/1, so 200 falls through; the test proves the segmented path actually executes (`every_apdu_segmented` + full payload equality). No panic risk (`from_raw` is a const passthrough, `enums/mod.rs`).

6. Placeholder→authoritative refresh — request-level covered by `explicit_upsert_refresh_makes_placeholder_authoritative` (add_device → oversized segments; upsert NONE/128 → refused). Primitive lifecycle: purge is real — row re-inserted via `upsert_placeholder` with `last_seen = now − 120s`, purged at >60s, `local_peer_segmentation` returns `None`; clear also returns `None`. Provenance lives inside `TableEntry`, so it cannot orphan.

7. Red evidence — verified: `git show 758ee906:crates/bacnet-client/src/client/requests.rs` equivalent (fetched at `refs/heads/dev`) contains no capability check anywhere between `check_transmittable_length` and `segmented_confirmed_request`. PR body lists exactly the four discriminators (A, B, C, G/H-refresh) as failing at base; consistent with base behavior (base would enter the segmented path for all four).

8. Error precision — sound. Refusal formats `{capability:?}`; `bacnet_enum!`'s Debug writes type-qualified names ("Segmentation::NONE" / "Segmentation::TRANSMIT"), which contain the asserted substrings; `Error::Segmentation(String)` Displays as "segmentation error: {0}" (`error.rs`), and `assert_segmentation_unsupported` pins the variant via `matches!`.

9. `assert_no_frame_reaches` soundness — acceptable. In every refusal test the request task is awaited to its error BEFORE the no-frame check, so any send would already be buffered in the peer's mpsc and would surface from `recv()` even after sender drop; the operative branch is the 50 ms timeout (channel stays open until `stop()` after the assertion). The residual weakness — an early-closed channel making the check vacuously pass — is not reachable here without also failing the preceding timeout-bounded assertions, and the refusal path provably returns before any send call and allocates no TSM timer, so no late frame exists. Not material.

Regression sweep: all pre-existing tests that seed authoritative rows were checked for interaction with the new refusal — `peer_max_apdu_tests.rs` seeds NONE but only issues sub-limit unsegmented requests; `tests.rs::confirmed_request_does_not_wrap_discovered_max_apdu_length` likewise unsegmented; ALL segmented-send tests (`tests.rs`, `segmentation_retransmit_tests.rs`, routed) use unknown peers or BOTH rows. No seeded-NONE/TRANSMIT test expects segmentation, so nothing silently regresses. Public API surface unchanged (`TableEntry`/`upsert_placeholder` are private/crate-private; `DeviceTable` field refactor invisible outside `discovery.rs`).

Findings:
- BLOCKER: none.
- HIGH: none.
- MEDIUM: none.
- LOW-1: Routed-path refusal is exercised only with NONE; no `add_routed_device` + TRANSMIT case exists, so the routed arm of the preflight (`routed_peer_segmentation`) is never proven against the second prohibition constant. Shared code makes misbehavior unlikely, but the matrix in the PR body slightly oversells symmetry. (`peer_segmentation_tests.rs`, fn `authoritative_routed_no_segmentation_refuses_segmented_request`.)
- LOW-2: I-Am provenance is verified at the primitive level only (`upsert_with_result` called directly in `discovery.rs::segmentation_provenance_tracks_row_lifecycle`); no test drives a real inbound I-Am through `dispatch.rs` and then issues an oversized request. Call site inspected and correct; coverage gap, not a defect.
- LOW-3: Test-run verification gap noted above — counts cited from the PR body, not independently executed by this reviewer.

READY.

jscott3201 (commented):
[Dataflow]
Reviewed head: 3c2728a9a756d033967e148a7f61e4ea9fc0730b
Base reviewed: 758ee906924b7211cc94002e96535e9d6ab31b2b

**READY**

Client/TSM/network dataflow lane for #371. Target identity confirmed via the PR API (base/head SHAs match the packet exactly; local shell unavailable in this toolset, so verification was done against the authoritative GitHub PR object). Findings below; none block shipping.

## Verified facts

1. **Refusal ordering** — In `confirmed_request_inner` (crates/bacnet-client/src/client/requests.rs:172-233), everything before the capability gate is pure computation: `unsegmented_apdu_size`, `target_transport_max_apdu_length`, and the device-table sizing block. The only fallible call before the gate is `check_transmittable_length` (requests.rs:206), which transmits nothing. The refusal sits inside `if unsegmented_apdu_size > remote_max_apdu` (requests.rs:207-228) strictly before `segmented_confirmed_request` (requests.rs:229-236), before `tsm.lock().await`/`allocate_invoke_id`/`register_transaction` (requests.rs:243-250), before `TsmGuard::new` (requests.rs:253), and before any `send_apdu`. On refusal the wire is silent and no transaction state is created — confirmed by the red-discriminator tests asserting immediate typed error plus frame silence (peer_segmentation_tests.rs:130-141).

2. **Target identity preserved (#402)** — Local → `dt.get_by_mac(mac)` + `dt.local_peer_segmentation(mac)`; Routed → `dt.get_by_network_address(...)` + `dt.routed_peer_segmentation(...)` (requests.rs:189-199). Both members of each pair select the same row: the lookups run back-to-back under one held `tokio::sync::Mutex` guard `dt` (requests.rs:185), so no mutation can interleave on `&self`; and `HashMap` iteration order is fixed per map instance absent mutation, so the `max_by_key(last_seen)` tie-break resolves identically in both calls. `local_peer_segmentation` re-derives via `get_by_mac` then instance-key fetch under the same borrow (discovery.rs:232-237) — coherent. Routed rows are excluded from `get_by_mac` by `is_local()` (discovery.rs:203-210), preserving #402.

3. **Provenance correctness** — `upsert_with_result` marks `Authoritative(device.segmentation_supported)` (discovery.rs:135-138); `upsert` delegates to it; `add_device` uses `upsert_placeholder` (client/discovery.rs:161); `add_routed_device` uses authoritative `upsert` (client/discovery.rs:225). The I-Am dispatch path (client/dispatch.rs, unmodified in this diff) still calls `upsert_with_result` — authoritative, unchanged. Provenance lives inside `TableEntry` in the same `HashMap<u32, TableEntry>` slot as the device data (discovery.rs:76-80, 148-158), so any replacement of an instance atomically replaces provenance, and `purge_stale_at`/`clear` remove entry+provenance together (discovery.rs:294-302) — orphaning is structurally impossible. Lifecycle covered by `segmentation_provenance_tracks_row_lifecycle` (discovery.rs:455-496).

4. **Receive-capable peers unchanged** — The gate matches only `NONE | TRANSMIT` (requests.rs:214-220); RECEIVE/BOTH/unknown raw values fall through to `segmented_confirmed_request` with `remote_max_segments` passed through untouched. `segmentation.rs` is not in the diff; max-segments handling is byte-for-byte base behavior.

5. **No lock across await** — The `dt` guard scope ends at the closing brace of the sizing block (requests.rs:205) before `check_transmittable_length` and all subsequent awaits. The TSM lock is confined to its own expression block (requests.rs:243-250) and released before encoding/sending. No new lock is held across `.await`; lock-acquisition awaits themselves are the pre-existing pattern.

6. **Public contracts intact** — `DeviceTable::upsert/all/get/get_by_mac/get_by_network_address/purge_stale/purge_stale_collect/clear/len/is_empty` keep identical signatures and observable behavior (discovery.rs:127-302; `all` now clones `e.device`, `purge_stale_at` maps removed entries back to `DiscoveredDevice`). New items are correctly scoped: `PeerSegmentation` pub(crate) (discovery.rs:70), `upsert_placeholder`/`local_peer_segmentation`/`routed_peer_segmentation` pub(crate), `TableEntry` private.

7. **Integration boundary** — All service wrappers route through `confirmed_request` or `confirmed_request_inner` (verified cov.rs send_subscribe_cov_request/send_subscribe_cov_property_request; file_list.rs read_range/atomic_read_file/etc.), so no path reaches `segmented_confirmed_request` without passing the new preflight. `Error::Segmentation(String)` pre-exists in bacnet-types/src/error.rs:73-75 (untouched crate).

## Findings

**Blocker:** none.
**High:** none.
**Medium:** none.

**Low (docs nit):** Public-doc intra-doc links from the public `DeviceTable` docs to pub(crate) items — `[PeerSegmentation]` and `[DeviceTable::upsert_with_result]` at crates/bacnet-client/src/discovery.rs:72-77 and :128-129 — can trigger rustdoc's `private_intra_doc_links` warning when building public docs. Cosmetic; does not affect compilation or clippy runs reported in the PR.

**Low (message formatting):** The refusal message formats the capability with `{capability:?}` (requests.rs:222-226), yielding e.g. `Segmentation::NONE` rather than the Display form `NONE`. Tests assert containment so they pass either way; purely cosmetic.

**Observation (pre-existing, not introduced here):** the capability decision is a separate lock acquisition from `resolve_device`'s address snapshot (documented non-atomic boundary, client/discovery.rs:12-20). The refusal evaluates the freshest row at request time, consistent with the documented freshest-wins contract; noted only so future readers know the two reads are intentionally not one snapshot.

PR #405, opened 2026-08-23
fix(server): correlate routed segmented transactions by source

### LINKED ISSUE #384: Server-side reassembly SegKey uses the immediate-hop MAC, so a routed peer via redundant routers splits its own session

jscott3201:
## Summary

- key valid routed server segmentation transactions by NPDU SNET/SADR and Invoke ID instead of immediate router MAC
- apply one canonical private key to request reassembly, peer Abort cleanup, segmented ComplexACK registration, and SegmentACK/Abort dispatch
- preserve local MAC identity and physical reply routing while adding alternate-router regressions and conservative conformance evidence

Closes #384

## Validation

- `cargo test -p bacnet-server --locked` (483 passed)
- `cargo deny check`
- conformance generator, file-size, no-secret, and diff-hygiene checks
- exact-ref segmentation A/B: three 15-second runs, zero errors, all `ok`; variable deltas showed no repeatable regression

## Scope boundary

Segmented ComplexACK frames continue to use the router MAC captured when the response starts. Dynamic outbound route migration is not part of this change.

## Known baseline

`cargo audit` continues to report unchanged `RUSTSEC-2026-0258` for `h2 0.4.15`; `Cargo.lock` is unchanged.


jscott3201 (commented):
Cycle 1 immutable review: PASS

Reviewed target: `440740f18e25cd2be35e3bc92aca416a13818216..56cd8afa59d2fc90d09bd8701f2e0ddfbc9fb3c0`.

- `review-luna`: PASS; no blocker/major or deferred finding.
- `review-dataflow-inkling`: PASS; no blocker/major. Its documentation-anchor observation is accurate, non-behavioral, and requires no change.
- Required CI is green on the exact reviewed head.
- Local package/workspace/feature/conformance gates and the exact-ref segmentation A/B guard passed.

No repair cycle is required.

PR #406, opened 2026-08-23
fix(objects): derive max segments from segmentation mode

### LINKED ISSUE #379: Device Max_Segments_Accepted defaults to 65 under SEGMENTED_TRANSMIT where Clause 12.11 requires 1

jscott3201:
## Summary

- Advertise `Max_Segments_Accepted = 1` for transmit-only Device objects, as required by Clause 12.11.
- Preserve omission for `NO_SEGMENTATION` and the existing value 65 for receive-capable and unknown non-`NONE` modes.
- Add mode-matrix regression coverage and a focused conformance-ledger child row.

## Scope boundaries

This changes Device-object construction only. It does not synchronize `DeviceConfig` with `ServerConfig`, change the server's 256-segment sequence-space guard, or implement client reassembly timers.

No benchmark was added because the changed branch runs only during Device-object construction.

## Validation

- Red-first regression: transmit-only mode returned 65 before the production change.
- `cargo test -p bacnet-objects --locked mode_derived_max_segments_accepted`
- workspace, all-feature, root-package, formatting, and clippy gates
- repository file-size and no-secret checks
- `cargo deny check`

`cargo audit` still reports the repository's existing `h2` advisory; `Cargo.lock` is unchanged.

Closes #379


jscott3201 (commented):
Cycle 1 consolidated review — **PASS**

Reviewed immutable head `bbb5f65a9e7d73d4e846132190a2c863ee708596` against base `b7214fc62907aaae9b377c1a42de0cb4ebe4a404` with the holistic and dataflow lanes.

- Confirmed blocker/major findings: none
- Scope: contained to the Clause 12.11 Device-property correction, regression matrix, and conservative conformance evidence
- Exact-head checks: Rustfmt, Clippy, Ubuntu tests, file-size, and no-secret checks passed; policy-gated jobs were skipped as expected
- Deferred non-blocking observations: the conservative conformance status remains appropriate; no wire-level test or changelog detail for unknown raw modes is required for this constructor-only slice

No repair cycle is required.

PR #407, opened 2026-08-23
fix(client): stop RequestTimer during segmented response reassembly

### LINKED ISSUE #380: Client reassembly reaper: 4s hardcoded vs Tseg×4, transmits an Abort PDU where 5.4.4.4 Timeout is indication-only, never completes the TSM
### Affected revision

### Behavior

While a segmented ComplexACK is being reassembled, the caller in `confirmed_request` (`crates/bacnet-client/src/client/requests.rs` ~284-314) keeps its `timeout(apdu_timeout_ms, &mut rx)` running, retransmits the original ConfirmedRequest on each expiry, and calls `cancel_transaction` after `apdu_retries`. A segmented response that takes longer than `apdu_timeout_ms × (apdu_retries + 1)` — 24 s at defaults — is cancelled while it is progressing normally, and the retransmitted requests are protocol noise the peer must dedupe.

### What the Standard says

Clause 5.4.4.3 `SegmentedComplexACK_Received` (extraction ~2812-2824): on receiving the first segment of the reply, "stop RequestTimer" — the request-level timer does not run during reassembly. SEGMENTED_CONF is then governed by SegmentTimer (Tseg × 4), not RequestTimer.

### Consequence after #367's fix

The reassembly-session lifecycle is now tied to the pending transaction, so when the caller cancels, the client actively Aborts the in-flight transfer at the next segment. That is the correct consequence of an incorrect cancellation: fixing this issue (stopping the request timer once reassembly starts) removes the incorrect cancellation itself. Before #367's fix the same cancellation merely orphaned the session and the result was thrown away regardless, so no delivered outcome has changed.

### Suggested direction

Signal the requester when its transaction enters reassembly (e.g. a marker on the pending TSM entry, checked by the timeout loop), and while reassembly is live suppress request retransmission and extend the wait to the SegmentTimer bound.

Related: #367, #319.

jscott3201:
## Summary

- Stop the confirmed-request `RequestTimer` after accepting segment zero of a segmented ComplexACK, as required by Clause 5.4.4.3.
- Move the transaction into a TSM-owned segmented-response phase with activity generations and a receive timeout of four APDU segment timeouts.
- Serialize retry/cancel decisions against segment admission without holding the TSM mutex across async transport I/O.
- Apply the handoff to both unsegmented and outgoing-segmented request response paths.

This PR addresses the caller-timer defect described in **#380's body**. The issue title describes a separate reassembly-reaper defect and is intentionally not part of this slice.

## Behavior and evidence

- A progressing segmented response can cross the original request timeout without retransmitting the ConfirmedRequest or losing its invoke ID.
- New, duplicate, and out-of-order segment activity invalidates stale receive-timer expirations.
- No-progress receive expiry returns local `Error::Timeout` and adds no peer Abort; later segments retain the existing `INVALID_APDU_IN_THIS_STATE` transaction gate.
- Retry accounting terminates after exactly the configured number of retransmissions, including `apdu_retries = 255`.
- A blocked or failed retry send cannot block segment admission or cancel a response admitted while transport I/O was pending.

## Scope boundaries

The hardcoded, inbound-driven reassembly reaper remains unchanged: its interval, peer `TSM_TIMEOUT` Abort, and missing TSM completion are separate follow-up work. This PR also does not add public configuration, change server/transport/encoding behavior, solve stale-stream invoke-ID aliasing, add CI, or promote public support claims.

No benchmark was added because this is state-machine correctness rather than a measured hotspot.

## Validation

- Red-first: `segmented_response_stops_request_timer_and_completes` observed a retransmitted ConfirmedRequest before the production change.
- Focused request-timer, lifecycle, TSM arbitration, response-correlation, duplicate-window, and segmented-request regressions
- `cargo test --workspace --exclude rusty-bacnet --locked --features bacnet-transport/ipv6,bacnet-transport/sc-tls`
- conformance generator/check, file-size, no-secret, `cargo deny check`, and diff checks

`Cargo.lock` is unchanged; the repository's existing audit baseline is unaffected.

Closes #380


jscott3201 (commented):
Cycle 1 consolidated review — **NEEDS_REPAIR**

Reviewed immutable head `d48a3c35dd574e19cb3b05c6cc4ffcecd7e1c358` against base `6377f154833f79a066a6de4d8763c1ba75816234`. Exact-head CI is green.

### MAJOR-407-01 — stale segmented-request cleanup can cancel a newer transaction

`wait_for_confirmed_response` now removes the timed-out TSM entry and releases its invoke ID. In the outgoing-segmented request path, `seg_ack_senders` cleanup runs afterward, followed by an unconditional `cancel_transaction` on the old `(peer, invoke_id)` when the waiter returned an error (`client/segmentation.rs:670-680`). A concurrent request can allocate the released invoke ID during that cleanup window, after which the stale cancellation removes the newer transaction.

**Bounded correction:** make post-wait cleanup ownership-safe—do not cancel by reusable key after the phase-aware waiter has already ended that transaction, or carry a transaction ownership token and cancel only the same registration. Add a deterministic concurrent invoke-ID reuse regression.

The dataflow lane found no additional blocker/major issue. One batched repair and final review cycle remain.

jscott3201 (commented):
Cycle 2 consolidated review — **REPLAN_REQUIRED**

Final immutable review targeted head `0045519a71abd902bc3526a2e13370b809d72280`; exact-head CI is green.

- `MAJOR-407-01`: **fixed**. Post-wait TSM cancellation no longer removes a newer transaction reusing the invoke ID, and the deterministic regression passes.
- `MAJOR-407-02`: **unresolved**. The same delayed cleanup still removes `seg_ack_senders[(peer, invoke_id)]` by reusable key. If the replacement request is also segmented, it inserts its own SegmentACK sender before the old request's delayed cleanup; the old cleanup removes the replacement sender. Subsequent SegmentACKs are discarded by dispatch and the replacement transfer times out. The repair test uses an unsegmented replacement, so it does not exercise this map-ownership collision.

A bounded future plan should make SegmentACK-sender cleanup owner/generation-checked or remove the old sender before the response-wait window, then prove a segmented replacement with deterministic ordering.

Per the two-cycle limit, no further repair or review cycle will run in this delivery attempt. PR #407 is not merge-eligible.

PR #408, opened 2026-08-23
fix(client): stop request timer during segmented response reassembly

### LINKED ISSUE #380: Client reassembly reaper: 4s hardcoded vs Tseg×4, transmits an Abort PDU where 5.4.4.4 Timeout is indication-only, never completes the TSM

jscott3201:
## Summary

- stop the confirmed-request `RequestTimer` when a valid sequence-zero segmented `ComplexACK` starts response reassembly
- serialize request and receive-segment timer expiry with TSM-owned phase and activity generations while preserving exact retry behavior and keeping transport I/O outside TSM locks
- make timeout cleanup ownership-safe when an invoke ID is immediately reused, including the outgoing segmented request's SegmentACK sender registration
- add deterministic lifecycle, retry, correlation, and segmented-replacement regressions and retain conservative conformance status

## Scope

This replacement supersedes closed PR #407 after its final review exposed sender-map cleanup that could remove a newer segmented request's registration. It addresses the RequestTimer/reassembly work described in the body of #380. The mismatched issue-title work for the separate inbound reassembly reaper remains out of scope.

Closes #380.

## Validation

- `cargo test --workspace --exclude rusty-bacnet --locked --features bacnet-transport/ipv6,bacnet-transport/sc-tls`
- `cargo deny check`
- `git diff --check`

The replacement segmented-request regression was first run against the seeded pre-fix behavior and failed because stale cleanup removed the replacement sender; it passes with channel-identity ownership checking.


jscott3201 (commented):
## Review cycle 1 — PASS

Reviewed immutable range `6377f154833f79a066a6de4d8763c1ba75816234..4c89fca17b67d78a24c2dc1e64b753118d91598d` with the holistic and dataflow lanes.

- Blocker/major findings: none
- Minor/nit findings: none
- Prior #407 regressions: stale TSM cancellation and stale SegmentACK sender removal are both corrected
- Exact-head checks: required GitHub checks green; local focused, package, workspace, feature, formatting, lint, conformance, policy, and repository hygiene gates green

The replacement segmented-request regression confirms that old cleanup cannot remove the reused invoke ID's current sender registration.

PR #410, opened 2026-08-23
fix(client): implement outgoing segmented request state

jscott3201:
## Summary

- model Clause 5.4.4.2 `SEGMENTED_REQUEST` and `SentAllSegments` in the client TSM
- reject premature SimpleACK, ComplexACK, and Error PDUs with the required client/local Abort while preserving Reject and server-Abort transitions
- accept terminal replies after the final segment is issued, including before the final SegmentACK
- qualify SegmentACK routes by immutable transaction owner and authoritative TSM phase

## Ordering and safety

The final send is polled without the TSM lock. Terminal admission observed during that first poll waits for the owner-qualified issuance token to resolve, so a response caused by the final send is accepted without allowing an already-ready response to skip the final segment. Cleanup and delayed transitions remain owner-qualified for immediate invoke-ID reuse.

## Validation

- `cargo check -p rusty-bacnet --locked`
- `cargo fmt --all --check`
- `cargo deny check`

### DIFF
--- a/crates/bacnet-client/src/client/dispatch.rs
+++ b/crates/bacnet-client/src/client/dispatch.rs
@@ -40,6 +40,27 @@ async fn take_current_reassembly(
     }
 }
 
+async fn admit_terminal_response(
+    tsm: &Arc<Mutex<Tsm>>,
+    tsm_mac: &MacAddr,
+    invoke_id: u8,
+) -> TerminalResponseAdmission {
+    let mut expected_owner = None;
+    loop {
+        let admission =
+            tsm.lock()
+                .await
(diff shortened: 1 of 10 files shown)


jscott3201 (commented):
Cycle 1 consolidated review — PASS at `899a1e8a124e96772e8a570105f5b720a2dcd49d`.

Both independent review lanes are blocker/major-clean. Exact-head lean CI is green. No repair cycle is required.

Deferred non-blocking observation: the single-poll `poll_fn` synchronization is unconventional but intentional and covered by focused ordering regressions.

PR #423, opened 2026-08-24
fix(services): enforce AtomicWriteFile record count

### LINKED ISSUE #418: AtomicWriteFile decoder accepts a record list shorter than Record Count
## Summary

Follow-up from PR #421 (#397). `AtomicWriteFileRequest::decode` (`crates/bacnet-services/src/file.rs`, record-access arm) reads `record_count`, then loops over the octet strings and `break`s when the content runs out, so a message whose `file-record-data` has fewer elements than `record-count` decodes successfully with a short list.

The server now refuses such a request in the handler with a Reject `MISSING_REQUIRED_PARAMETER` before any mutation (Clause 14.2 Service Procedure writes "the number of records specified by 'Record Count'"), but the defect belongs in the decoder: a truncated SEQUENCE OF is a syntax error and every consumer of the request type (client, CLI, Python) sees the inconsistent value today.

## Acceptance criteria

- The decoder returns a decoding error when fewer than `record_count` octet strings are present (and, if reachable, when more are present before the closing tag).
- Round-trip tests for the boundary; the handler-level guard in `handle_atomic_write_file` can then become defensive-only.
- Consider whether `MAX_DECODED_ITEMS` interacts (it already bounds `record_count`).

## Relevant code

- `crates/bacnet-services/src/file.rs` (record-access decode loop)
- `crates/bacnet-server/src/handlers/file.rs` (handler guard)
- Related: #309 (decoders silently truncate over-wide Unsigned fields)

jscott3201:
## Summary

- require AtomicWriteFile record payloads to contain exactly the declared `Record Count`
- reject short lists with `MISSING_REQUIRED_PARAMETER` and surplus inner elements with `TOO_MANY_ARGUMENTS`
- keep the server's pre-mutation guard as defense in depth and update the conformance evidence

## Testing

- `cargo test -p bacnet-services atomic_write_record --locked`
- `cargo test -p bacnet-services bounded_record_counts_accept_limit_with_leading_zero --locked`
- `cargo test -p bacnet-server record_write_with_short_payload_list_is_rejected_without_mutation --locked`
- `cargo fmt --all --check`

Closes #418


jscott3201 (commented):
Cycle 1 immutable review — `8ca261e296fc70462a3322a1f561db95f46afbe6`

Both independent review lanes passed with 0 blocker / 0 major findings. The decoder now enforces exact record cardinality while preserving zero-count and `MAX_DECODED_ITEMS` boundaries, the handler keeps its pre-mutation defense, and the conformance evidence matches the implementation. Required exact-head CI checks are green; policy-inapplicable jobs are skipped.

**PASS**

PR #429, opened 2026-08-25
fix(objects): encode complete active COV subscriptions

### LINKED ISSUE #183: Active_COV_Subscriptions omits the BACnetRecipient and the monitored property identifier, both mandatory in BACnetCOVSubscription
### Affected revision

### Behavior

`DeviceObject::read_property` encodes each entry of `Active_COV_Subscriptions` as `List([ObjectIdentifier(monitored_property_reference.object_identifier), Unsigned(recipient.process_identifier), Boolean(issue_confirmed_notifications), Unsigned(time_remaining)])` plus an optional `Real(cov_increment)` (`crates/bacnet-objects/src/device/mod.rs:422-441`). This is the only encoder for the property — the sole other reference outside tests, `crates/bacnet-server/src/handlers/read_property.rs:40`, merely classifies it as an array property.

The source data carries what is dropped. `BACnetCOVSubscription` (`crates/bacnet-types/src/constructed/mod.rs:637-643`) holds a `BACnetRecipientProcess` (`:623-626`) whose `recipient` field — the subscriber's device identifier or address — is never read, and a `BACnetObjectPropertyReference` (`:174-181`) whose `property_identifier` and `property_array_index` are never read.

Clause 21.6 (p884):

```
BACnetCOVSubscription ::= SEQUENCE {
      recipient                     [0] BACnetRecipientProcess,
      monitored-property-reference  [1] BACnetObjectPropertyReference,
      issue-confirmed-notifications [2] BOOLEAN,
      time-remaining                [3] Unsigned,
      cov-increment                 [4] REAL OPTIONAL
      }
```

In the same clause, `BACnetRecipientProcess ::= SEQUENCE { recipient [0] BACnetRecipient, process-identifier [1] Unsigned32 }` and `BACnetObjectPropertyReference ::= SEQUENCE { object-identifier [0] BACnetObjectIdentifier, property-identifier [1] BACnetPropertyIdentifier, property-array-index [2] Unsigned OPTIONAL }`. Neither `recipient` nor `property-identifier` is optional, and neither reaches the wire: a client reading `Active_COV_Subscriptions` cannot determine who is subscribed or to which property. That is data loss, not a framing deviation.

The framing is also wrong. All five emitted fields carry application tags where `[0]`..`[4]` are assigned, and the two nested SEQUENCEs contain tagged elements, so Clause 20.2.1.3 (p843) applies: "If the production being encoded contains tagged elements, then the encoding is called 'constructed' and shall consist of (a) an 'opening' tag ... (b) the complete encoding, with tags, of the zero, one, or more elements that comprise the data; (c) a 'closing' tag ...".

Distinct from #152, which reworks framing for `Recipient_List` while preserving every field's value; here two mandatory values are absent before framing is even considered.

### Suggested direction

- encode `recipient.recipient` (the `BACnetRecipient` CHOICE) and `monitored_property_reference.property_identifier`, plus `property_array_index` when present;
- emit the context tags and SEQUENCE framing of the Clause 21.6 production rather than flat application tags, sharing whatever context-tag primitives #152 introduces;
- test a subscription with a device recipient and one with an address recipient, and a subscription to a property other than `Present_Value`, asserting on the encoded bytes rather than on the `PropertyValue` tree.

### Relevant code

- `crates/bacnet-objects/src/device/mod.rs`
- `crates/bacnet-types/src/constructed/mod.rs`

jscott3201:
## Summary
- add an encode-only `BACnetCOVSubscription` codec with bare, ordered BACnetLIST framing
- return complete Device `Active_COV_Subscriptions` values as `ApplicationData`
- prove ReadProperty and ReadPropertyMultiple preserve the constructed bytes
- add qualified conformance evidence and document the Rust compatibility change

## Standard grounding
Paraphrased from Standard 135-2020 Clause 12.11/Table 12-13/§12.11.31, Clause 12.1.5.2, the Clause 20 sequence/list rules, and the Clause 21 `BACnetCOVSubscription`, recipient, object-property-reference, ReadProperty-ACK, and ReadAccessResult productions.

## Validation
- `cargo test -p bacnet-encoding cov_subscription --locked`
- `cargo test -p bacnet-objects active_cov_subscriptions --locked`
- focused `bacnet-server` ReadProperty and ReadPropertyMultiple regressions
- file-size, no-secret, and `git diff --check` gates

## Compatibility and remaining scope
Direct Rust Device reads now return `ApplicationData` instead of the previous lossy nested `List`. Runtime server COV-table synchronization remains excluded, so the ledger row stays `in-progress`. This does not add a decoder, Python model, public support claim, or array-index behavior.

Closes #183


jscott3201 (commented):
Cycle 1 consolidated immutable review: **PASS** at `aba4a6cc792305cc9c14c722dc5ceb1eec95e72e`.

Both required review lenses found no blocker or major issue. The exact-head lean CI checks are green. Runtime COV-table synchronization and decoding remain explicitly outside this PR's scope; no minor/nit findings require follow-up.

PR #430, opened 2026-08-25
feat: seserialize EngineeringUnits from symbolic name

alvskar:
Add `Deserialize` to units, similar to `ObjectType` and `PropertyIdentifier`, to enable using the newtypes directly instead of a `u16` and for readability.

`Deserialize` can in theory be added to all the enums here by putting it in the macro but I assume this was decided against for a reason.

Ran `cargo test -p bacnet-types --features serde` as well as the checks listed under Development.

### DIFF
--- a/crates/bacnet-types/src/enums/units.rs
+++ b/crates/bacnet-types/src/enums/units.rs
@@ -2,6 +2,12 @@
 // EngineeringUnits (Clause 21) — large enum, grouped by category
 // ===========================================================================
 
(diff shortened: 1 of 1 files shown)


jscott3201:
I'm going to hold this open for a few days until I get this next release fully put together. I hope to have that by end of week and that should clear the issue log as well. Once all those are rolled in I'll look at this and the overall serialization and deserialization work on the library. I have been changing some types under the hood lately so it will be easier to do this off the 1.0.0 base.

Issue #431, opened 2026-08-25
BACnetClient and BACnetServer cannot share a transport, so one device cannot both initiate and execute


alvskar:
`BACnetServer` and `BACnetClient` both own their transport so they are unable to share one, but a BACnet device that does both is an expectation of certain device profiles. Closest I can find to an example of that is `examples/rust/bip_client_server.rs`, but the client there has no Device object, and both take `.port(0)`.

Putting the two on separate ports has its own consequences: BACnet is widely expected to stay on 47808, different ports imply different networks, and the reading device ends up not being the answering device. For MS/TP, where there's only one serial port, it's unclear how this would work.

Is this division intended in its current form? Giving `BACnetServer` client functions would make a lot of this cleaner, though it would need some structure for code reuse between the two. `BACnetServer` already sends requests and handles replies; it just can't handle replies with data at the moment.

jscott3201:
Thanks for the detailed report. We confirmed the ownership/lifecycle gap: the current client and server each start their own `NetworkLayer`, so the example’s two ephemeral B/IP ports are separate BACnet/IP endpoints, and the same split cannot represent one MS/TP station over one serial owner.

We also checked this against the repository’s local Standard 135-2020 source:

- §5.4.4 and §5.4.5 define requesting and responding transaction roles; they do not require separate physical endpoints.
- §12.1.1.3, §12.11, and §22.1.5 establish one Device identity and the configured Network Port objects for a BACnet device.
- Annex J §J.1.2 makes the UDP port part of the B/IP address and permits independent port groups, so two unrelated `.port(0)` bindings are not one shared device endpoint.
- Annex K distinguishes initiate/execute roles, while Annex L §L.4/§L.4.1 combines both sides in profiles such as B-BC.

So the reported dual-role use case is spec-aligned. The Standard does not prescribe where those methods belong in this Rust API, and adding thin requester methods to `BACnetServer` alone would not solve ingress ownership, ComplexAck/segmentation routing, invoke-ID coordination, shutdown, or MS/TP’s single-use reply path.

We have added this to the compliance roadmap as a separate M01A lane:

1. **PR-0113:** one-owner endpoint ingress and lifecycle;
2. **PR-0114:** coordinated confirmed transactions and owner-qualified response routing;
3. **PR-0115:** an opt-in Rust combined-device API over one B/IP socket and one Device identity;
4. **PR-0116:** one-link MS/TP proof and final issue closure.

[...]

PR #433, opened 2026-08-25
feat(network): introduce single-owner endpoint ingress

jscott3201:
## Summary

- add a crate-private endpoint ingress owner around one `NetworkLayer`
- route inbound requests separately from terminal and segment traffic without cloning `ReceivedApdu`
- preserve transport metadata and the single-use reply channel through routing
- make malformed, unsupported, full, and closed-route outcomes explicit
- cover lifecycle ownership, bounded queues, cancellation, and exact-once classification

## Scope

This is the private ingress and lifecycle foundation for the shared-device endpoint work in #431. Transaction correlation, invoke-ID coordination, public combined-device APIs, and B/IP or MS/TP composition remain in later slices. Existing standalone client and server lifecycle code is unchanged.

## Verification

- `cargo test -p bacnet-client lifecycle --locked`
- `git diff --check`

Refs #431

### DIFF
--- a/crates/bacnet-network/src/endpoint_ingress.rs
+++ b/crates/bacnet-network/src/endpoint_ingress.rs
@@ -1,0 +1,230 @@
+use bacnet_encoding::apdu::{decode_apdu, Apdu};
+use bacnet_transport::port::TransportPort;
+use bacnet_types::error::Error;
+use tokio::sync::{mpsc, oneshot};
+use tokio::task::JoinHandle;
+
+use crate::layer::{NetworkLayer, ReceivedApdu};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


jscott3201:
PR gate cycle 1 reviewed exact head `2075dd4f534ba665f8354974978608f7cae406ff`: **PASS**.

Both holistic and dataflow lanes are blocker/major-clean. Exact-head Test, Clippy, Rustfmt, file-size, and no-secret checks passed; release-only jobs were skipped as expected. No repair cycle is required.

Deferred non-blocking follow-ups for the transaction-routing slice: measure or avoid the `Bytes` payload clone used only for APDU decoding, add an explicit terminal-route `reply_tx` assertion, and add concurrent bounded-queue stress when PR-0114 introduces correlation.

PR #434, opened 2026-08-25
feat(endpoint-core): establish shared private endpoint core

jscott3201:
## Summary

- add an unpublished, cycle-free `bacnet-endpoint-core` workspace crate
- move the single-owner endpoint ingress into the shared private boundary without changing its tested behavior
- add a thread-safe coordinator with one device-wide pool of 256 invoke IDs
- retain role, peer, service, terminal-policy, and generation metadata for later role adapters
- enforce exact-token cleanup and bounded terminal-response admission

## Contract

Standard 135-2020 requires invoke IDs to be unique across all outstanding confirmed requests generated by a device. The coordinator therefore allocates globally rather than per peer. Generation tokens fence stale local cleanup work; they are not presented as a way to distinguish identical delayed wire responses after invoke-ID reuse.

This is the owner-approved bounded foundation split for PR-0114. Client and server TSM/dispatch integration remains the next slice. No existing published crate API, standalone lifecycle, service set, or transport behavior changes here.

## Verification

- `cargo check -p bacnet-endpoint-core --all-targets --locked`
- `cargo test -p bacnet-client tsm --locked`
- `cargo test -p bacnet-client lifecycle --locked`
- `git diff --check`

Refs #431


jscott3201:
PR gate cycle 1 reviewed exact head `a737b82e45b9b20f45c41dcfc5ca1fb530796413`: **repair required**.

Confirmed major finding: Abort direction must depend on the outbound lease owner. A requester lease accepts an Abort sent by the remote server (`server=true`), while a server-notification lease accepts an Abort sent by the remote client (`server=false`). The current universal `server=true` check would leave an aborted notification lease active until fallback cleanup.

The request to wire client/server adapters or coordinator-owned segment timers in this foundation PR is dismissed as out of the owner-approved split: the prior ingress had no production caller, and role integration/timers remain the next bounded slice. One batched repair will add owner-aware Abort admission and regression coverage.

jscott3201:
Final PR gate cycle reviewed repaired head `751756de86a5b9023d15620c8ed07cbdec6ee6a2`: **PASS**.

`HOL-434-001` is fixed: Abort direction is validated against the lease owner, mismatch paths leave state untouched, and regression coverage proves exact completion, generation-changed reuse, and stale-token safety. Both final review lanes are blocker/major-clean, and exact-head Test, Clippy, Rustfmt, file-size, and no-secret checks passed. No further repair cycle was used or is available.

PR #435, opened 2026-08-25
feat(endpoint-core): prepare crate for publication

jscott3201:
## Summary

- make `bacnet-endpoint-core` a publishable workspace package
- add it to the existing crates.io release sequence before server and client
- add an explicit requester policy that accepts either SimpleACK or ComplexACK
- preserve lenient Error correlation and defer segmented ComplexACK service validation to reassembly
- keep strict ACK policies and server-notification restrictions available

## Why

Published role crates cannot depend on endpoint-core while it is marked `publish = false` and absent from crates.io. This is the owner-approved packaging and compatibility prerequisite for the client/server transaction-adapter slices.

No client or server dependency, TSM integration, lifecycle change, combined-device API, or transport composition is included here.

## Verification

- `cargo package -p bacnet-endpoint-core --locked --allow-dirty`
- `cargo publish -p bacnet-endpoint-core --dry-run --locked --allow-dirty`
- `cargo check -p bacnet-endpoint-core --all-targets --locked`
- `git diff --check`

The package and publish dry-run both completed verification; no crate was uploaded.

Refs #431

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -283,6 +283,12 @@ jobs:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
       - run: sleep 30
 
+      - name: Publish bacnet-endpoint-core
+        run: cargo publish -p bacnet-endpoint-core --no-verify
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+      - run: sleep 30
+
       - name: Publish bacnet-server
         run: cargo publish -p bacnet-server --no-verify
         env:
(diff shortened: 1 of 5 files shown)


jscott3201:
PR gate cycle 1 reviewed exact head `dfe45480450e296cef1086ffa5e158048edfd3ed`: **PASS**.

Both holistic and dataflow lanes are blocker/major-clean. Package metadata and release ordering are dependency-safe; requester ACK compatibility, lenient Error admission, deferred segmented ComplexACK validation, exact-token cleanup, and server-notification restrictions are covered. Exact-head Test, Clippy, Rustfmt, file-size, and no-secret checks passed.

Deferred non-blocking note: the policy query helpers remain private; the future adapter can use `admit` outcomes and only needs public query methods if its implementation proves otherwise. No repair cycle is required.

PR #437, opened 2026-08-25
feat(server): coordinate confirmed notification transactions

jscott3201:
## Summary

- make confirmed COV and event notifications use one device-wide endpoint-core invoke-ID pool
- retain exact lease tokens across retries and reclaim them on every terminal or cleanup path
- correlate direct and routed terminals through canonical peer identity
- admit SimpleACK, Error, Reject, and client-originated Abort without consuming segmented-response traffic
- learn routed reply paths only after successful notification admission
- close and drain notification transactions during server stop and drop

## Scope

This completes the standalone server-notification side of PR-0114. Existing `BACnetServer<T>` APIs, legacy `ServerTsm` behavior, COV concurrency throttling, unconfirmed notifications, and segmented server responses remain intact. Shared ingress, public endpoint composition, B/IP composition, and MS/TP proof remain deferred.

Endpoint-core is not yet present on crates.io, so server package/publish dry-runs are intentionally deferred to the tag release that publishes endpoint-core first.

## Verification

- `cargo test -p bacnet-server notification --locked`
- `cargo test -p bacnet-server cov --locked`
- `cargo test -p bacnet-server event --locked`
- `cargo test -p bacnet-server segmentation --locked`
- `cargo test -p bacnet-server --features sc-tls --locked`
- `git diff --check`

Refs #431


jscott3201:
PR gate cycle 1 reviewed exact head `733c73f898bb0482c7ebbc62f16a2ec51e541c30`: **PASS**.

Both holistic and dataflow lanes are blocker/major-clean. Exact-head Test, Clippy, Rustfmt, file-size, and no-secret checks passed. The review verified the global COV/event lease pool, retry token retention, terminal filtering, canonical routed admission, admitted-only router learning, exact cleanup, and stop/drop drain behavior.

Deferred non-blocking note: a terminal arriving exactly at the timeout/rearm boundary can make the worker report `Closed` after the terminal was already delivered and the lease safely completed. This does not leak or corrupt state and can be refined later if result diagnostics require it. No repair cycle is required.

PR #438, opened 2026-08-26
feat(endpoint): add shared loopback runtime seam

jscott3201:
## Summary

- add a bounded direct egress handle to the single-owner endpoint session
- add a hidden direct ReadProperty requester using an injected coordinator
- add a private ReadProperty responder that reuses standalone server response construction
- support exact completion from centrally pre-admitted requester and notification terminals
- prove concurrent requester/responder traffic over one Loopback ingress with both wire transactions using invoke ID 0
- preserve exact `reply_tx` movement and endpoint-only shutdown authority

## Scope

This is the internal runtime prerequisite for PR-0115. It does not add a public B/IP device API, real UDP, routed/broadcast egress, segmentation, discovery, service expansion, Python bindings, or production server-to-client dependency. Existing standalone client and server lifecycles remain unchanged.

`bacnet-server` uses `bacnet-client` only as a dev-dependency for the private composition proof.

## Verification

- `git diff --check`

Before implementation, `cargo clean` reclaimed 25.2 GiB. No package, publication, or B/IP gate was run for this internal Loopback-only slice.

Refs #431

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -359,6 +359,7 @@ dependencies = [
 name = "bacnet-server"
 version = "0.10.1"
 dependencies = [
+ "bacnet-client",
  "bacnet-encoding",
  "bacnet-endpoint-core",
  "bacnet-network",
--- a/crates/bacnet-client/src/client/lifecycle.rs
+++ b/crates/bacnet-client/src/client/lifecycle.rs
@@ -51,13 +51,8 @@ impl<T: TransportPort + 'static> BACnetClient<T> {
 
         let network = Arc::new(network);
 
-        let tsm_config = TsmConfig {
-            apdu_timeout_ms: config.apdu_timeout_ms,
-            apdu_segment_timeout_ms: config.apdu_timeout_ms,
-            apdu_retries: config.apdu_retries,
-        };
         let coordinator = Arc::new(OutboundTransactionCoordinator::new());
-        let tsm = Arc::new(Mutex::new(Tsm::new_coordinated(tsm_config, coordinator)));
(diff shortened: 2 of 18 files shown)


jscott3201:
PR gate cycle 1 reviewed exact head `995c363da6d3071ecead2a15f19694a53dc6ad80`: **one batched repair required**.

Confirmed majors:
1. A pending direct egress send currently blocks APDU classification, and biased scheduling can starve inbound terminal traffic. The session must continue servicing ingress fairly while one bounded egress operation is pending.
2. A centrally admitted segmented ComplexACK for the direct requester is currently discarded, leaving the request pending until timeout. The unsupported segment must trigger one client Abort, exact lease cleanup, and the existing local Abort error without a second coordinator admission.

The low-level `#[doc(hidden)]` public seam is accepted by owner decision as the minimum cross-crate contract needed by the future `bacnet-server` facade; it is not treated as a repair finding in this cycle. One repair will address only the two runtime defects and their focused regressions.

jscott3201:
Final PR gate cycle reviewed repaired head `7ac67122d727727e3c8d453e88211a93f91640f6`: **PASS**.

`H-438-01` is fixed: inbound APDUs continue to be classified fairly while one bounded egress send is pending, and cancellation drains accepted commands before the single network stop. `H-438-02` is fixed: one pre-admitted unsupported segmented ComplexACK now produces one client Abort, resolves the local request, releases the exact lease, and rejects duplicate/stale attempts without re-admission.

Both final review lanes are blocker/major-clean, and exact-head Test, Clippy, Rustfmt, file-size, and no-secret checks passed. The owner-approved hidden low-level seam remains unchanged.

PR #453, opened 2026-08-28
feat(server): resolve device recipient bindings

jscott3201:
## Summary
- add a bounded server-owned Device recipient binding authority with explicit builder configuration and passive inbound I-Am learning
- route local and remote Device recipients by unicast while preserving final routed identity and the configured next hop
- keep configured entries authoritative, expire observed entries deterministically, and fail closed for unknown, stale, invalid, or capacity-rejected bindings
- preserve existing address-recipient behavior and public start/configuration compatibility

## Scope
This is an intermediate notification-distribution slice. Active discovery, pending delivery, persistence, and retry/backoff policy for unresolved recipients remain deferred.

## Validation
- focused `bacnet-server` binding and Device-recipient tests
- workspace feature test matrix
- root crate test check
- conformance-doc generation check
- file-size, secret, and diff checks

Refs #125

### DIFF
--- a/crates/bacnet-server/src/server/clock.rs
+++ b/crates/bacnet-server/src/server/clock.rs
@@ -55,6 +55,16 @@ impl Default for ClockConfig {
 }
 
 impl<T: TransportPort + 'static> BACnetServer<T> {
+    pub(super) async fn start_with_clock_mode(
+        config: ServerConfig,
+        db: ObjectDatabase,
+        transport: T,
+        clock_config: Option<ClockConfig>,
+    ) -> Result<Self, Error> {
+        Self::start_with_clock_mode_and_bindings(config, db, transport, clock_config, Vec::new())
+            .await
+    }
+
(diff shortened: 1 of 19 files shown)


jscott3201:
Cycle 1 review found two major, bounded issues on `adee2e3d1d70e232f837d3fe22725c1a4916f4ad`:

1. **SC startup validation:** SC-specific builder setup can begin its TLS/WebSocket connection before transport-specific broadcast validation rejects an invalid configured binding. Preflight SC binding addresses before any connection attempt and add a regression proving invalid configuration wins before connection setup.
2. **Observed-binding freshness during confirmed retries:** the route is resolved once, so later attempts can emit after the observed binding has expired. Carry observed freshness/provenance into the selected route and suppress every retry whose binding is no longer fresh; add a deterministic expiry-during-retry regression.

The repair must preserve configured-route retry behavior, final-peer correlation, existing address-recipient behavior, and the generalized source-comment policy.

PR #464, opened 2026-08-30
fix(mstp): Clause 9 CRCs + USB stream reassembly

bbartling:
## Summary
- Correct MS/TP header/data CRC polynomials to BACnet Clause 9.6 (`0x81` / `0x8408`); prior `0xE0` / `0xA001` passed self-round-trip but rejected live trunk frames (e.g. Token `55 FF 00 00 07 00 00 37`).
- Preserve partial frames across USB host read chunk gaps (do not treat async `read()` gaps as Clause 9 `T_frame_abort`).
- Add literal golden vectors and residual checks (`0x55` / `0xF0B8`).

## Commits
1. `a9912b8` fix(mstp): USB chunk gaps must not trigger T_frame_abort
2. `6a70b85` fix(mstp): use BACnet Clause 9 header and data CRCs

## Test plan
- [x] `cargo test -p bacnet-transport` (343 unit + 2 integration)
- [ ] upstream CI green
- [ ] Vibe13 passive sniff against live BASRT/FEC trunk (follow-on)

Made with [Cursor](https://cursor.com)

PR #465, opened 2026-08-30
fix(mstp): conform master token and PFM transitions to Clause 9.5.6

bbartling:
## Summary
Fixes multi-master coexistence: Rust MAC 3 must not PFM/Token in ways that rebuild the ring as `0→3→0` and exclude FEC MAC 7.

**Depends on #464** (Clause 9 CRC + USB stream). This branch is stacked on that tip. Prefer review/merge after #464, or treat as follow-up commits on the same train.

## Design (Clause 9.5.6)
- `done_with_token()`; no unconditional `pass_token()` on DoneWithToken
- NextStationUnknown → PFM to TS+1 (never Token TS→TS)
- Maintenance PS advances from TS toward NS only
- PassToken failure → PFM after Nretry_token (no blind Tokens)

## Tests
Mandatory A–E regressions (`clause956_tests.rs`), including 2000-rotation ring `0→3→7→0`.

Made with [Cursor](https://cursor.com)


bbartling:
Closing: keeping this work on the bbartling fork only (`vibe13-mstp`). Not pursuing merge into upstream.

PR #467, opened 2026-08-30
fix(mstp): Clause 9 CRC, USB stream assembly, and 9.5.6 token/PFM

bbartling:
## Summary
- **Clause 9.6 CRCs** — replace incorrect reflected polys (`0xE0` / Modbus `0xA001`) with BACnet header `0x81` and data `0x8408`, plus live golden vectors (Token `0←7` CRC `0x37`).
- **USB host gaps ≠ `T_frame_abort`** — async serial reads arrive in multi-ms chunks; treating inter-read gaps as wire `T_frame_abort` (~1.56 ms @ 38400) discarded partial frames. Stream decoder (`Complete` / `NeedMore` / `Invalid`) + host stale-partial timeout.
- **Clause 9.5.6 token/PFM** — `DONE_WITH_TOKEN` no longer reverse-scans from `NS+1` or emits Token `TS→TS`; maintenance `PS` advances `TS→NS` only; FindNewSuccessor via PollForMaster. Regressions A–E.
- Split frame tests into `mstp_frame_tests.rs` for the 700 LOC file-size gate.

## Lab evidence (not claimed as formal conformance)
Validated on a live MS/TP trunk (BASRT MAC 0 + JCI FEC MAC 7 + Waveshare USB RS-485 @ 38400) with JENEsys Workbench: passive token decode, FEC ReadProperty coexistence, and a Rust mini-device (MAC 3) discovered with points Polled `{ok}`. Details in bbartling playground vibe13 docs.

## Limitations (current)
- **Not a conformance claim** — Clause 9 checklist items beyond CRC / stream assembly / 9.5.6 master token+PFM are still open.
- **No extended frames** — types 32/33, COBS, CRC-32K not implemented; do not claim oversized-frame or router conformance.
- **Gates still open in lab** — shared endpoint + soak (vibe13 Gates 5–6) not closed.
- **Host USB timing** — stale-partial timeout is host-scheduling aware; wire `T_frame_abort` is not emulated on chunk boundaries.
- **Windows CI flake** — unrelated `max_retry_budget` 3s wall-clock flake is **not** in this PR (tracked on fork `main` only).

## Test plan
- [ ] `cargo test -p bacnet-transport --features serial`
- [ ] Confirm CI green (fmt/clippy/audit/deny/ubuntu/macos/windows)
- [ ] Review regressions in `clause956_tests.rs` and CRC goldens in `mstp_frame_tests.rs`

Made with [Cursor](https://cursor.com)

PR #468, opened 2026-08-30
feat(python): MS/TP transport + mini-device example

bbartling:
## Summary
- add `transport="mstp"` to Python `BACnetClient` and `BACnetServer`
- accept MS/TP unicast peer addresses as `"7"` or `"mstp:7"`, including valid slave MACs through 254 while rejecting broadcast 255
- preserve existing positional Python arguments and make the new MS/TP options keyword-only
- validate local master/Max_Master/Max_Info_Frames and the six Clause 9 baud rates before serial I/O
- preserve server registrations when serial open fails and add MS/TP drop/abort cleanup
- add a standalone USB-RS485 mini-device example and installed-wheel compatibility tests

## Dependency
PR #467 is merged. This branch is based on its Clause 9 CRC, host stream assembly, standard-frame boundary, and token/PFM corrections.

## Validation
- `cargo test -p bacnet-transport --features serial --locked` (365 tests)
- package and CI-native workspace Clippy/tests
- conformance-doc, 700-LOC, no-secret, and diff checks
- isolated `uvx maturin` CPython 3.14t wheel build/install/import
- 11 installed-artifact tests covering runtime/stub signatures, lazy construction, baud/config error ordering, registration retention, and peer-address boundaries
- installed `.pyi`/`py.typed` validation, Pyright, Python syntax, and example `--help`

## Limitations
- not a full Clause 9 conformance claim
- no COBS/CRC-32K frame types 32–127, segmentation claim, combined endpoint, or GATE-0007 closure
- real adapter/platform interoperability and soak remain optional hardware validation
- requires an RS-485 adapter/driver that provides appropriate transmit-direction handling

Hardware is not required for the automated validation in this PR.


jscott3201:
### Review cycle 1 — one bounded repair batch required

1. **PR468-001 (major): baud contract.** Owner decision: accept only the ANSI/ASHRAE 135-2020 Clause 9 standard rates (`9600`, `19200`, `38400`, `57600`, `76800`, `115200`) and reject other values before serial I/O.
2. **PR468-002 (major): registration retry.** Server startup currently drains pending objects before the newly reachable MS/TP serial open; an absent/busy/denied port loses registrations before retry. Open/build the MS/TP transport before destructive drain, or restore registrations on failure, with regression evidence.
3. **PR468-003 (major): drop cleanup.** `AnyTransport::abort()` delegates to MS/TP, but `MstpTransport` uses the no-op default. If shared client operations defer final drop, the receive task can retain the serial handle. Add bounded synchronous abort cleanup and prove the serial/task is released.

Preserve the existing keyword-only API repair, peer address range, artifact tests, and all explicit hardware/conformance/GATE-0007 exclusions. This is the sole cycle-1 repair batch.

PR #469, opened 2026-08-30
feat(client): enforce routed path APDU limits

### LINKED ISSUE #370: Clause 5.2.1.2(b) is bounded by the local egress link only: intervening-network NPDU limits are never enforced

jscott3201:
## Summary
- size routed confirmed APDUs from peer, live local transport, and configured/learned/conservative path NPDU bounds
- serialize one active confirmed request per immediate-router+DNET path so reason-4 feedback completes the exact TSM owner
- bound path state at 256 entries with safe evidence-preserving reclamation and fail-closed capacity errors
- quarantine ambiguous generations and use ingress sequencing so stale reason-4 controls cannot claim later requests
- surface typed `RoutedPathTooLong` and `RoutedPathCapacityExceeded` errors while preserving direct/unconfirmed behavior
- add opt-in decoded network-control ingress without changing `NetworkLayer::start()`

## Standards and policy
Grounded against Standard 135-2020 §5.2.1.2, Clause 6/Table 6-1, §6.4.4, and §19.4. Unknown paths use the 228-octet routed NPDU baseline minus the actual forwarded destination+source NPCI envelope. Full active probing and persistent cache are deferred.

The new public `Error` variants can require downstream exhaustive matches to add arms; this is documented in `docs/rust-api.md`.

Closes #370

## Validation
- `cargo test -p bacnet-client --all-features --locked`
- `git diff --check`

Cycle 1 findings RP469-M01 and DF-001 were repaired in one batch and both independent final review lanes passed at `acdfafa98eb3f014718f66d689c594a29d2c2187`.

### DIFF
--- a/crates/bacnet-client/src/client/lifecycle.rs
+++ b/crates/bacnet-client/src/client/lifecycle.rs
@@ -73,7 +73,9 @@ impl<T: TransportPort + 'static> BACnetClient<T> {
         #[cfg(test)]
         let segmented_cleanup_dispatch = Arc::clone(&segmented_cleanup);
         let response_limits = ResponseLimits::from_config(&config);
-        let routed_path_limits = Arc::new(RoutedPathLimits::default());
+        let routed_path_limits = Arc::new(RoutedPathLimits::new(routed_path_quarantine_horizon(
+            &config,
+        )));
         let routed_path_limits_dispatch = Arc::clone(&routed_path_limits);
 
         let dispatch_task = tokio::spawn(async move {
--- a/crates/bacnet-client/src/client/lifecycle.rs
+++ b/crates/bacnet-client/src/client/lifecycle.rs
@@ -42,6 +42,7 @@ impl<T: TransportPort + 'static> BACnetClient<T> {
         options.validate()?;
 
         let mut network = NetworkLayer::new(transport);
+        let mut network_control_rx = network.enable_network_control_receiver()?;
         let mut apdu_rx = network.start().await?;
         config.max_apdu_length = cap_max_apdu_to_transport(
             config.max_apdu_length,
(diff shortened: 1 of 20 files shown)


jscott3201:
Cycle 1 review found two major root causes to repair as one batch:

1. **RP469-M01 — bounded path state:** `RoutedPathLimits` inserts one entry per caller-selected router+DNET and never reclaims no-evidence paths or caps configured/learned entries. Add a deterministic capacity policy, safely reclaim idle entries without splitting a held/waited gate, and fail closed when durable path state is full. Add hostile unique-path tests.
2. **DF-001 — stale reason-4 correlation:** reason 4 carries only reason+DNET. A delayed control from request A can currently claim request B after B becomes active on the same path. Preserve ingress ordering/generation and add a conservative stale-control quarantine for ambiguous prior termination so only feedback attributable to the current active generation can terminate/teach. Add a regression for delayed feedback across request generations.

Please preserve the reviewed envelope math, owner-qualified TSM completion, direct/unconfirmed behavior, and existing API scope. This is the sole cycle-1 repair batch.

PR #472, opened 2026-08-31
feat(objects): add reliability evaluation inhibit

### LINKED ISSUE #232: Reliability_Evaluation_Inhibit is unimplemented on all nine intrinsic-reporting object types
## The gap

`PropertyIdentifier::RELIABILITY_EVALUATION_INHIBIT = 357` exists in `crates/bacnet-types/src/enums/property_id.rs:371` and is referenced nowhere in the workspace.

It is the standard's runtime override for stopping reliability-evaluation, and Clause 13.2.2 identifies it as the *only* such lever:

> "In objects that support event-state-detection, the reporting of changes in Reliability cannot be disabled while event-state-detection is enabled. Instead, the object may be stopped from detecting faults through the Reliability_Evaluation_Inhibit property."

Since #167 (PR #215), `Reliability` drives `Event_State` through the shared fault-precedence path, so an operator with a known-flapping sensor has no conformant way to stop it generating FAULT transitions short of disabling event detection entirely.

## Definition and conformance

The definition is printed identically on all nine types:

> "This property, of type BOOLEAN, indicates whether (TRUE) or not (FALSE) reliability-evaluation is disabled in the object. This property is a runtime override that allows temporary disabling of reliability-evaluation.
>
> When reliability-evaluation is disabled, the Reliability property shall have the value NO_FAULT_DETECTED unless Out_Of_Service is TRUE and an alternate value has been written to the Reliability property."

| Object type | Clause | Code |
|---|---:|---|
| Analog Input | 12.2 | `O7` |
| Analog Output | 12.3 | `O6` |
| Analog Value | 12.4 | `O8` |
| Binary Input | 12.6 | `O9` |
| Binary Output | 12.7 | `O8` |
| Binary Value | 12.8 | `O10` |
| Multi-state Input | 12.18 | `O7` |
| Multi-state Output | 12.19 | `O5` |
| Multi-state Value | 12.20 | `O8` |

Every footnote reads:

> "If this property is present, then the Reliability property shall be present."

So it is optional, but adding it obliges `Reliability` to be present — which bears on #218, where six of these types expose `Reliability` read-only.

## Note on the semantics, which are not what they look like

TRUE does **not** freeze the current value and does **not** by itself make `Reliability` writable. It *forces* `NO_FAULT_DETECTED`, with exactly one exception: while `Out_Of_Service` is TRUE and an alternate value has been written.

That matters because the obvious reading — "inhibit means stop touching it" — would make this property look like the fix for #219, and it is not. Wiring it that way would pin `Reliability` at `NO_FAULT_DETECTED` permanently rather than let a client own the value.

## Scope

- Add the property to the nine object types, defaulting FALSE (the standard specifies no default; Clause 15.3 makes uninitialized values a local matter)
- While TRUE, force `Reliability` to `NO_FAULT_DETECTED`, honoring the `Out_Of_Service` exception
- Make it a standing invariant rather than an edge-triggered action on the write, for the same reason `Event_Detection_Enable` needed that treatment in #222 — an object constructed or restored with the property already TRUE never sees a write
- Skip such objects in whatever performs reliability-evaluation (today `FaultDetector`, see #231)

## Related

#219 (fault detector clobbers a client-owned `Reliability`), #231 (the evaluation derives the wrong thing from the wrong properties), #218 (`Reliability` not writable on six types).

jscott3201:
## Summary
- elect `Reliability_Evaluation_Inhibit` on all nine analog, binary, and multi-state Input/Output/Value object types
- enforce the standing `NO_FAULT_DETECTED` invariant while inhibited, including the exact out-of-service/client-written Reliability exception
- preserve object-owned range-fault state, OOS restoration ownership, and WPM rollback state
- skip inhibited hooks in periodic fault evaluation without changing event-state detection or Event Enrollment
- expose the product-elected default-FALSE, BACnet-readable/writable property through Property_List and PICS

## Standard traceability
- ASHRAE 135-2020 §§12.2–12.4, 12.6–12.8, 12.18–12.20, and 13.2.2
- the Standard defines behavior when the optional property is present; presence, default, and writability are explicit product choices

## Validation
- focused red-first nine-object matrix (8 expected failures before production wiring)
- conformance-doc, file-size, no-secret, and diff checks

Closes #232.


jscott3201:
Cycle 1 consolidated review found one major issue (`HOL-472-001`): the no-range expansion of `impl_intrinsic_write_rollback!` evaluates `None?` while capturing Reliability/OOS/inhibit state, so Analog Output and the six Binary/Multi-State objects return no rollback token. A failed WPM can therefore leave Reliability normalized after replaying `Reliability_Evaluation_Inhibit=false`. The bounded repair is to emit a token with `range_fault_ownership: None` for no-range objects and add a failed-WPM regression on a non-range object proving exact restoration.

jscott3201:
Final cycle resolution: `HOL-472-001` is disproved/resolved. The trailing `?` in `None$(.or(Some(...)))?` is the macro optional-repetition quantifier, so no-range objects retain the rollback token. Commit `5012715` adds a permanent Binary Input failed-WPM regression proving restoration of inhibit, Reliability, OOS state, same-value client ownership, and subsequent OOS cycles. Both final reviewers passed; all required exact-head checks are green.

PR #473, opened 2026-08-31
fix(objects): add multi-state configuration reliability

### LINKED ISSUE #226: Multi-state objects never set Reliability to MULTI_STATE_OUT_OF_RANGE or CONFIGURATION_ERROR when Number_Of_States shrinks below a property value
## What the standard requires

Clause 12.19 (Multi-state Output), under `Number_Of_States`, prescribes two distinct reliability outcomes when `Number_Of_States` is reduced below the value already held by another property:

> "If the Number_of_States property value becomes less than the value of the Present_Value, the object shall have a Reliability of MULTI_STATE_OUT_OF_RANGE as long as this situation remains, unless the object is out of service. It is a local matter whether Priority_Array, Relinquish_Default, Present_Value, and Feedback_Value properties are modified when the Number_of_States property value becomes less than their current values. If any of those properties other than Present_Value are out of range, the value of the Reliability property shall remain CONFIGURATION_ERROR, unless the object is out of service."

So:

- `Present_Value` out of range → `Reliability` = `MULTI_STATE_OUT_OF_RANGE`
- `Priority_Array`, `Relinquish_Default` or `Feedback_Value` out of range → `Reliability` = `CONFIGURATION_ERROR`

Both persist "as long as this situation remains", and both are suppressed while the object is out of service.

## What the code does

Neither reliability value is ever assigned on multi-state objects. `Reliability` on `MultiStateOutputObject` stays at its initial value regardless of how `Number_Of_States` is written.

Writing a smaller `Number_Of_States` over an object whose `Present_Value`, `Relinquish_Default` or `Feedback_Value` already exceeds it leaves the object silently inconsistent: it reports a `Present_Value` outside its own declared state set with `Reliability` = `NO_FAULT_DETECTED`.

## Why this matters more now

Since #167 (PR #215), `Reliability` drives `Event_State` through the shared fault-precedence path. An object that should be reporting `CONFIGURATION_ERROR` or `MULTI_STATE_OUT_OF_RANGE` therefore also fails to reach `FAULT` and never emits the `CHANGE_OF_RELIABILITY` notification a controller would use to detect the misconfiguration.

## Note on the range-rejection alternative

It is tempting to reject an out-of-range `Feedback_Value` write instead. That is not equivalent, and is worse: the standard does not forbid an out-of-range `Feedback_Value`, it *defines a state for it*. Rejecting the write makes `CONFIGURATION_ERROR` unreachable for that property. `Feedback_Value` is "an indication of the actual value of the entity controlled by Present_Value" and "the manner by which the Feedback_Value is determined shall be a local matter" — a physical feedback reading can legitimately fall outside the configured state set, and representing that as a reliability state is exactly the mechanism the standard provides.

The same reasoning does not automatically extend to `Present_Value`, which is commanded rather than sensed; rejecting a commanded value outside the state set is defensible and is the current behavior.

## Scope

- `crates/bacnet-objects/src/multistate/output.rs`, `input.rs`, `value.rs`
- Recompute reliability on writes to `Number_Of_States`, and clear it when the situation is resolved
- Honor the "unless the object is out of service" suppression in both cases
- Check the equivalent `Number_Of_States` text in Clause 12.18 (Multi-state Input) and 12.20 (Multi-state Value) before implementing; the property sets differ, so the list of properties that trigger `CONFIGURATION_ERROR` differs too

Found while implementing #222.

jscott3201:
## Summary
- add object-owned Multi-state reliability evaluation for MSI, MSO, and MSV
- report `MULTI_STATE_OUT_OF_RANGE` for invalid effective values and `CONFIGURATION_ERROR` for retained invalid MSO/MSV configuration
- recompute synchronously after relevant mutations while retaining the periodic reliability hook as a safety net
- preserve first-stage/external Reliability, OOS/inhibit authority, intrinsic event flow, and WPM rollback
- add fallible local `set_number_of_states` setters with one atomic State_Text resize policy

## Standard traceability
|---|---|
| §12.18 Multi-state Input | Effective Present_Value outside `1..=Number_Of_States` produces `MULTI_STATE_OUT_OF_RANGE` unless Out_Of_Service. |
| §12.19 Multi-state Output | Invalid retained Priority_Array entries, Relinquish_Default, or Feedback_Value produce `CONFIGURATION_ERROR`; otherwise an invalid effective Present_Value produces `MULTI_STATE_OUT_OF_RANGE`. |
| §12.20 Multi-state Value | Invalid retained Priority_Array entries, Relinquish_Default, or Alarm_Values produce `CONFIGURATION_ERROR`; otherwise an invalid effective Present_Value produces `MULTI_STATE_OUT_OF_RANGE`. |

Configuration errors take precedence over an invalid effective Present_Value. Evaluation releases only Reliability values owned by this evaluator.

## Owner decisions
- `Number_Of_States` remains network/PICS read-only; this PR adds only a fallible local Rust setter.
- `Fault_Values` remains unsupported and is deferred.
- State_Text preserves retained entries, truncates on shrink, and appends `State {n}` labels on growth.

## Validation
- focused red-first Multi-state reliability tests

## Deferred
- Multi-state `Fault_Values`
- BACnet/Python writes for `Number_Of_States`
- Event Enrollment or event-kernel redesign
- changes to existing direct Priority_Array command semantics
- broad support/conformance claims

Closes #226

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Object-owned multi-state configuration Reliability and recovery for
+  Multi-state Input, Output, and Value (#226). The Standard requires
+  `MULTI_STATE_OUT_OF_RANGE` for an invalid retained `Present_Value`, and
+  `CONFIGURATION_ERROR` (with precedence) for invalid retained MSO priority,
+  default, or feedback configuration and MSV priority, default, or alarm
+  configuration. Relevant successful mutations now recompute synchronously;
+  the periodic object hook shares the same owned first-stage evaluator, while
+  intrinsic FAULT/NORMAL reporting remains in the existing event kernel. A new
+  fallible Rust-only `set_number_of_states` API supplies the local count-change
+  path: zero is rejected before mutation, shrink truncates `State_Text`, and
+  growth retains its prefix and appends constructor-style `State {n}` labels.
+  The local API and label policy are product choices; `Number_Of_States` stays
+  read-only through BACnet WP/WPM, PICS, and Python. MSV `Fault_Values` remains
+  unsupported, and no broader conformance claim is made.
+
 - Product-elected `Reliability_Evaluation_Inhibit` support on Analog,
   Binary, and Multi-state Input/Output/Value objects (#232). The optional
   Boolean is present on all nine types, defaults to FALSE, is BACnet-readable
--- a/crates/bacnet-objects/src/multistate/input.rs
+++ b/crates/bacnet-objects/src/multistate/input.rs
@@ -20,6 +20,7 @@ pub struct MultiStateInputObject {
     reliability: u32,
     reliability_before_out_of_service: Option<u32>,
     reliability_inhibit: common::ReliabilityInhibitState,
+    reliability_evaluator: MultiStateReliabilityState,
     state_text: Vec<String>,
     /// CHANGE_OF_STATE event detector.
     event_detector: ChangeOfStateDetector,
@@ -48,6 +49,7 @@ impl MultiStateInputObject {
             reliability: 0,
             reliability_before_out_of_service: None,
             reliability_inhibit: common::ReliabilityInhibitState::default(),
+            reliability_evaluator: MultiStateReliabilityState::default(),
             state_text: (1..=number_of_states)
                 .map(|i| format!("State {i}"))
                 .collect(),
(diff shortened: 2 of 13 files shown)


jscott3201:

**RB-PR473-001 (major): WPM cannot restore retained invalid Multi-state command configuration.** A local `Number_Of_States` shrink may intentionally retain an out-of-range MSO/MSV `Relinquish_Default` or inactive `Priority_Array` slot. If a later WritePropertyMultiple repairs that value and then fails on another property, rollback replays the old value through normal client validation, which rejects it. The failed request can therefore leave the repair applied and report rollback failure, violating the repository’s atomic WPM policy.

**Required bounded repair:** capture and restore the relevant MSO/MSV command configuration plus derived Present_Value, Reliability, and private evaluator ownership through an object-owned rollback token that bypasses client range validation only during rollback. Add failing-WPM regressions for invalid retained defaults and priority slots on both MSO and MSV.

All exact-head CI checks passed; no other blocker/major finding was confirmed.

PR #475, opened 2026-08-31
fix(cli): decode and iterate AtomicReadFile reads

### LINKED ISSUE #419: bacnet-cli file read writes the raw AtomicReadFile-ACK bytes to --output instead of the file payload
## Summary

Follow-up from PR #421 (#397). `crates/bacnet-cli/src/commands/file.rs` (`read` subcommand, roughly lines 28–41) takes the value returned by `BACnetClient` (`crates/bacnet-client/src/client/file_list.rs`, which returns the undecoded AtomicReadFile-ACK service bytes) and writes it straight to `--output`, then prints "Wrote N bytes". Before #397 the server returned empty payloads so the mistake was invisible; now the CLI writes tag-prefixed ACK encoding over a real payload.

## Acceptance criteria

- Decode the ACK (`AtomicReadFileAck::decode`) and write only `file_data` (stream) or the concatenated / delimited records (record access — decide and document the on-disk form).
- Honour `end_of_file` to loop until the whole file is read when the request count is smaller than the file, or document that a single window is read.
- A CLI-level test or example round trip against the in-tree server (write then read a small file).

## Relevant code

- `crates/bacnet-cli/src/commands/file.rs`
- `crates/bacnet-client/src/client/file_list.rs`

jscott3201:
## Summary
- strictly decode AtomicReadFile ACKs and add a source-compatible typed single-window client helper while preserving the raw Rust API
- make `bacnet file-read` iterate returned stream/record windows until authoritative EOF and output payload bytes rather than encoded ACKs
- stage stream files and deterministic per-record directories until EOF, refuse existing destinations, and clean up failed transfers
- correct built-in empty File reads to return EOF and narrowly update the existing conformance evidence without changing support status

## CLI contract
- `--access stream|record` defaults to stream; `--count` is a positive per-request window
- stream mode writes/displays only decoded octets
- record mode requires `--output DIR` and writes `record-{absolute-index:010}.bin`, preserving zero-length records
- existing output targets are refused; successful staged output is published only after EOF

## Standard basis
- ANSI/ASHRAE 135-2020 §14.1 / Table 14-1: EOF is authoritative and ACKs return actual stream/record positions and payload/count
- Clause 21 AtomicReadFile-ACK production: stream data is an OCTET STRING; record data is a sequence of OCTET STRINGs
- host filesystem naming, overwrite, and cross-request cleanup are documented local CLI policy, not BACnet requirements

## Scope boundaries
- existing `BACnetClient::atomic_read_file -> Result<Bytes>` and Python behavior remain unchanged
- no implicit record concatenation, unrelated CLI/server redesign, new CI, README/PICS support promotion, or endpoint work
- `BACNET-14-FILE-STORAGE` status and support counts remain unchanged

## Validation
- focused strict ACK, decoded-client, CLI state-machine/round-trip, and empty EOF tests
- `cargo test -p bacnet-cli --locked`
- generated conformance docs, file-size, no-secret, and diff checks

Closes #419


jscott3201:
Cycle 1 consolidated review findings for head `b47e1c8bfd9f197c07786f77ebb165e7d9a11b1d`:

- **H-475-01 (major):** stream reads without `--output` retain every window and then allocate a complete hex rendering. A large or never-EOF peer can exhaust memory. Add a documented bounded display limit with a safe error directing larger reads to `--output`, plus a multi-window limit regression.
- **DF-001 (major):** the retrieval loop accepts returned starts that overlap or skip the requested cursor and appends data anyway, including terminal EOF windows, allowing silent corruption/incomplete output. Require exact cursor continuity for both stream and record ACKs before any sink write, and add overlap/gap regressions proving staged output is not published.

These are being handled as the single permitted batched repair. The retargeted head will receive the final immutable review cycle.

Issue #242, opened 2026-07-28
Decide whether Acked_Transitions should be network-writable on any object type


jscott3201:
### Why this is filed as a question rather than a defect

Nothing observably misbehaves today. `Acked_Transitions` is denied on every object type that exposes it, and that denial became consistent everywhere with #227. This issue exists because the *reasoning* behind the denial has never been settled against the standard, and the question has been carried as a loose item across three sessions instead of being tracked.

### The question

ASHRAE 135-2020 Clause 13.2.3 gives the **alarm-acknowledgment process** ownership of `Acked_Transitions`: the flags are cleared when a transition is distributed and set when an AcknowledgeAlarm is received for that transition. That is an internal lifecycle, not a client-settable value.

Two readings follow, and they lead to different code:

1. **Denial is correct and should be documented as normative.** `Acked_Transitions` is owned by the acknowledgment process, so a network write is meaningless and `WRITE_ACCESS_DENIED` is right. In that case the current behavior stands, but the *rustdoc should say why*, citing the clause — right now the denial reads as an omission rather than a decision, which is exactly how the #227 regression happened (a brief asserted the denial was inherited rather than intended, and the fix made it network-writable on the analog types).

2. **The conformance table permits it at implementor option.** Clause 12.1 allows an `O` or `R` property to be writable at implementor option "unless specifically prohibited in the text describing that particular standard object's property". If Clause 13.2.3 does not amount to such a prohibition, denial is a choice rather than a requirement — and Annex K's BIBBs should be checked, since AE-AVM-A was found during #216 to positively require a workstation to *write* certain event properties.

### What settling it would produce

- Either a rustdoc comment on each denial citing the clause that mandates it, so the next brief cannot assert it is accidental;
- or a defect, if the standard turns out to require writability somewhere this stack denies it.

Both outcomes are cheap. The value is in removing an unexamined assumption from nine object types.

### Relevant context

- #227 — the regression that made `Acked_Transitions` network-writable on the analog types, where it had been an explicit denial. The brief that caused it stated the denial as an inherited premise rather than a decision.
- #123 — maintaining `Acked_Transitions` and `Event_Time_Stamps` when an event fires. Nothing currently clears the field, so the acknowledgment process this issue is about is itself unimplemented; these two should probably be settled together.
- #175 — AcknowledgeAlarm never issues the mandated ACK_NOTIFICATION, which is the other half of the same process.

### Relevant code

[...]

PR #490, opened 2026-09-02
fix(server): preserve committed WritePropertyMultiple prefix

jscott3201:
## Summary
- execute WritePropertyMultiple attempts in wire order and retain every successful write before the first failure
- encode the formal service-16 Error body with the first failed object/property/optional array index
- classify malformed requests as Reject before any commit and `SERVICES / INVALID_TAG` Result(-) after a committed prefix
- emit the confirmed response before event/COV notifications derived only from the committed prefix
- preserve legacy low-level/high-level error compatibility while removing the bundled server's obsolete WPM rollback behavior

## Source-to-behavior traceability
|---|---|
| §15.10 and §15.10.1.3 | Process writes in request order, retain the successful prefix, stop on the first failed write, and report its coordinate. |
| §15.10.1.3 read-only failure row | `PROPERTY / WRITE_ACCESS_DENIED` identifies the exact failed property; dispatch coverage uses Event Enrollment `Description` followed by `Acked_Transitions`. |
| Clause 18.9 | Pre-commit syntax failures use a narrow Reject reason. |
| Clause 21 Error and `BACnetObjectPropertyReference` | Service-16 Result(-) uses constructed `[0] Error` followed by `[1]` first-failed reference, without prepending the generic Error pair. |

For syntax failure after a successful prefix, this implementation returns `SERVICES / INVALID_TAG`; when the failed coordinate cannot be decoded, it uses the documented local sentinel `DEVICE:4194303 / ALL / no index`. Empty requests and empty property lists retain the existing SimpleACK/no-op behavior.

## Compatibility
- This intentionally replaces the repository's stronger atomic-rollback behavior with the ordered prefix semantics required by §15.10.
- Public object snapshot/restore hooks remain source-compatible for object-local/downstream use, but Service 16 no longer invokes them.
- Legacy generic WPM Error PDUs remain decodable.
- Low-level `ErrorPdu.error_data` retains the complete formal body; high-level Rust/Python clients continue exposing class/code only.

## Validation
- focused WPM cursor, formal Error codec/APDU, ordered-prefix, Event Enrollment, client projection, and committed-prefix COV tests
- macOS workspace feature gate with `bacnet-types/serde`, `bacnet-transport/ipv6`, and `bacnet-transport/sc-tls`

Linux-only serial/Ethernet feature coverage remains delegated to exact-head CI.

## Deferred
- PR-0208 `Acked_Transitions` policy work
- public high-level first-failed-reference projection
- changes to WriteProperty, other formal service Error productions, public TSM/Python shapes, or object snapshot APIs

Refs #242


jscott3201:

**WPM-DATAFLOW-001 (major): decoded failed coordinates are lost on value-framing errors.** In `crates/bacnet-services/src/common/property_value_decode.rs`, value-opening failures reset `reference_complete` even after the property identifier and optional array index have decoded. The cursor then omits that reference, so a malformed second value after a committed first write returns `SERVICES / INVALID_TAG` with the `DEVICE:4194303 / ALL` sentinel instead of the actual failed object/property/index.

**Required bounded repair:** retain the completed coordinate for value-stage framing failures once property ID and optional index parsing succeeded, and add a committed-prefix regression that asserts the exact formal WPM Error reference. No protocol/API redesign is needed.

Holistic review otherwise passed, and all required exact-head checks are green.

PR #491, opened 2026-09-02
feat(objects,server): supply Present_Value from the application

alvskar:
Adds `BACnetObject::set_present_value_internal` and `BACnetServer::set_present_value_local`, so the application hosting the database can set an Input's `Present_Value`. Mirrors the existing `set_reliability_internal`.

Without something like this, it's not possible for the device process to update the values in the object database, as `write_property` is only the network route and there's no process side equivalent.

### DIFF
--- a/crates/bacnet-objects/src/analog/input.rs
+++ b/crates/bacnet-objects/src/analog/input.rs
@@ -71,6 +71,21 @@ impl AnalogInputObject {
         self.present_value = value;
     }
 
+    /// Validate and store a `Present_Value` write, without any access check.
+    ///
+    /// Shared by the network and internal routes, which differ only in the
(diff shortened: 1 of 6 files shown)


jscott3201:
I'll add this here shortly and check the release I'm working on for this as well.

jscott3201:
Thank you for this contribution. GitHub did not grant maintainer permission to update this fork branch onto current `dev`, so I imported exact commit `12f556433d111ddecfafbcf96ff7f3cd9bc76342` into #504, preserved co-author attribution, and added the source-backed Input-only validation/COV tests needed for the current object contracts.

Superseded by #504. The separate Python follow-up is tracked in #503.

PR #496, opened 2026-09-02
fix(server): correct GetEnrollmentSummary projection

### LINKED ISSUE #131: Restrict GetEnrollmentSummary and report actual event type and priority

jscott3201:
## Summary
- restrict deprecated `GetEnrollmentSummary` output to objects with explicit event-generation capability
- apply acknowledgment, enrollment, event-state, event-type, priority, and notification-class filters conjunctively
- report the object’s configured/implied event algorithm and the Notification Class priority associated with its latest committed transition
- support the complete device-or-address `BACnetRecipientProcess` enrollment filter
- reject malformed required projection state instead of fabricating type, priority, or class values

## Source-to-behavior traceability
|---|---|
| §13.11 / Table 13-15 | Searches only explicit event-initiating capabilities, ignores advertised `Event_Detection_Enable = FALSE`, combines every filter logically, and returns zero or more exact summaries. |
| §13.11 Acknowledgment Filter | Evaluates all three semantic `Acked_Transitions` bits for ACKED / NOT-ACKED. |
| §13.11 Enrollment Filter | Matches exact configured `BACnetRecipient` (Device or Address) plus Process Identifier membership; delivery day/time/transition/confirmed eligibility does not change enrollment membership. |
| §13.11 Event Type / Priority filters | Projects the object’s one configured or implied event algorithm and selects priority by the most recently committed TO_OFFNORMAL / TO_FAULT / TO_NORMAL coordinate. |
| §13.11.1.2.1.5 and §12.21 | Resolves the referenced Notification Class directly by object instance; the class object’s own `Notification_Class` property is unrelated to its identity. |
| Clause 21 service productions | Encodes and decodes the full `BACnetRecipientProcess` CHOICE and preserves optional Notification Class ACK fields. |

The Standard retains this service only for historical client interoperability and discourages new server implementations. This change corrects the already-exposed server behavior; it does not broaden a support or conformance claim.

## Compatibility
- `RecipientProcess` now stores `recipient: BACnetRecipient` rather than a device-only identifier. This is an owner-approved breaking Rust source migration required by the wire contract.
- Existing stock event families opt in explicitly: AI/AO/AV use `OUT_OF_RANGE`; BI/BV/MSI/MSV use `CHANGE_OF_STATE`; BO/MSO use `COMMAND_FAILURE`; configured/supported Event Enrollment uses its stored `Event_Type`.
- `CHANGE_OF_RELIABILITY` remains a transition notification override, not a second summary algorithm.
- Alert Enrollment and objects without proven event-generation capability remain excluded.
- A fresh object with no committed transition uses TO_NORMAL priority only when `Event_State == NORMAL` and canonical `Acked_Transitions == 0b111`; inconsistent no-history state fails with `DEVICE / OPERATIONAL_PROBLEM`.

## Validation
- focused event-history, capability, Notification Class, service-codec, and handler/filter tests

## Deferred
- new event algorithms or Alert Enrollment evaluation
- persistence or wire exposure of latest-transition history
- GetAlarmSummary / GetEventInformation changes
- Python/client surfaces and README/PICS/conformance expansion
- the pre-existing generic notification-routing class lookup policy

Closes #131


jscott3201:

**DSF-001 (major): the new latest-transition coordinate is omitted from intrinsic write rollback.** `IntrinsicWriteRollback::EventDetection` captures/restores Event_State, Acked_Transitions, pending/fault state, timestamps, original states, and message text, but not `EventHistory.last_transition`. Disabling Event_Detection resets the new marker; restoring a captured snapshot can therefore leave a previously committed non-initial event state with no transition coordinate. `GetEnrollmentSummary` then returns `DEVICE / OPERATIONAL_PROBLEM` instead of the restored transition’s priority.

**Required bounded repair:** include the private latest-transition coordinate in the shared intrinsic EventDetection rollback token and restoration path, and add a regression proving commit → disable/reset → restore preserves the coordinate and summary projection. Event Enrollment/Alert Enrollment already snapshot the full `EventHistory`; keep those paths intact.

Holistic review otherwise passed, and all required exact-head CI checks are green.

Issue #526, opened 2026-09-04
BACnet/SC: validate reconnect backoff to prevent zero-delay retry storms


jscott3201:
## Finding

Reviewed at `92f1ee7b2962d17514dd813bde6ce175961f5192`.

`ScReconnectConfig` exposes public delay fields with no validation:

- [initial_delay_ms and max_delay_ms accept any u64](https://github.com/jscott3201/rusty-bacnet/blob/92f1ee7b2962d17514dd813bde6ce175961f5192/crates/bacnet-transport/src/sc/reconnect.rs#L1-L24)
- reconnect [sleeps the configured duration and doubles it](https://github.com/jscott3201/rusty-bacnet/blob/92f1ee7b2962d17514dd813bde6ce175961f5192/crates/bacnet-transport/src/sc/mod.rs#L570-L600)

`initial_delay_ms = 0` and `max_delay_ms = 0` produce an immediate retry loop. A configuration error combined with a Hub outage can create a self-induced CPU, DNS/connect, and log storm across every configured node. This is a defensive availability concern even though the initiating error is local configuration.

## Requested change

- Validate reconnect configuration at construction/build time.
- Require a positive minimum delay, `max_delay >= initial_delay`, and a bounded retry policy.
- Add jitter and a deployment-safe lower bound to avoid synchronized reconnect herds.
- Rate-limit repeated reconnect diagnostics.
- Return a clear configuration error instead of silently accepting a storm setting.

## Acceptance evidence

- Zero delay, inverted ranges, and otherwise unsafe values are rejected deterministically.
- Valid configurations retain capped exponential backoff with tested jitter bounds.
- A simulated Hub outage cannot produce more connection attempts than the configured budget.
- Multiple clients do not retry in lockstep under the default configuration.

jscott3201:
## Partial delivery: PR #537 merged
Merged into dev at 1f0f62582db5521c6cbb6f88ee7ba1ae3a9f9634 (reviewed head 769af4e5eb174f6dc40ce852ededc02471745e85).

Delivered a shared configuration validator rejecting zero initial/max delays and inverted ranges, including max_retries=0. Raw SC startup validates before owned state, socket I/O, or tasks; Rust SC client/server builders validate before TLS lookup and dialing, with server validation also preceding binding construction. Existing fields, signatures, defaults, positive ordered values, and failover/restoration behavior are preserved.

[...]

Issue #525, opened 2026-09-04
BACnet/SC Hub: cancel removed client handlers so connection slots cannot leak


jscott3201:
## Finding

Reviewed at `92f1ee7b2962d17514dd813bde6ce175961f5192`.

Several Hub removal paths can remove a client from the routing map without reliably terminating its connection task:

- [handle_client waits in select before checking its closed flag](https://github.com/jscott3201/rusty-bacnet/blob/92f1ee7b2962d17514dd813bde6ce175961f5192/crates/bacnet-transport/src/sc_hub.rs#L171-L190)
- replacement [sets the old closed flag, then later calls Notify::notify_waiters](https://github.com/jscott3201/rusty-bacnet/blob/92f1ee7b2962d17514dd813bde6ce175961f5192/crates/bacnet-transport/src/sc_hub.rs#L326-L399)
- heartbeat timeout similarly sets the flag and calls `notify_waiters`, while [heartbeat send failure only removes the map entry](https://github.com/jscott3201/rusty-bacnet/blob/92f1ee7b2962d17514dd813bde6ce175961f5192/crates/bacnet-transport/src/sc_hub/connection.rs#L90-L129)
- the per-connection guard holds one of [512 active connection slots](https://github.com/jscott3201/rusty-bacnet/blob/92f1ee7b2962d17514dd813bde6ce175961f5192/crates/bacnet-transport/src/sc_hub/connection.rs#L32-L67) until that task exits

Tokio `Notify::notify_waiters` does not retain a permit for a future waiter. If notification occurs while the handler is processing another branch, the next loop can block forever on `read.next()`. The heartbeat-send-failure branch does not signal the handler at all.

## Availability impact

Repeated replacement or write-side failure can leave detached handlers and active-slot guards alive after their clients disappear from the Hub map. Enough leaked handlers can exhaust the pre-handshake connection limit and deny new legitimate SC connections. This is a concrete lifecycle defect related to, but not replaced by, the public lifecycle/API work in #476.

## Requested change

- Use retained cancellation semantics, or check the closed flag before every blocking wait with a race-free wakeup.
- Centralize removal as map removal -> cancel -> close -> bounded join.
- Signal the handler on heartbeat send failure as well as timeout/replacement.
- Track handlers so Hub shutdown can prove they exited.

## Acceptance evidence

- Replacement while the old handler is not currently waiting still exits the old task and releases its slot.
[...]

Issue #514, opened 2026-09-04
BACnet/SC Hub: publish pending heartbeat state before sending to avoid a lost ACK race


jscott3201:
## Reliability and availability finding

Reviewed at dev commit 92f1ee7b2962d17514dd813bde6ce175961f5192.

The hub heartbeat sweep sends a Heartbeat-Request before it publishes the pending message ID:

- crates/bacnet-transport/src/sc_hub/connection.rs:102-119 sends the request.
- Only after the send succeeds, lines 129-140 store pending_heartbeat_id and its timestamp.
- crates/bacnet-transport/src/sc_hub/heartbeat.rs:44-64 clears an ACK only when it matches a currently published pending ID.

The receive path runs concurrently. A fast peer can ACK while the pending ID is still zero, causing the valid ACK to be ignored. The sweep then publishes the ID and a later sweep can evict the healthy client as timed out.

## Impact

A responsive BACnet/SC node can be disconnected nondeterministically. Reconnect/replacement behavior can compound this into control-plane churn and temporary reachability loss.

## Recommended bounded change

- Atomically publish the pending heartbeat generation/ID and timestamp before the send becomes observable.
- If encoding, lock acquisition, send, or timeout fails, clear that exact pending generation and remove/close the peer according to one consistent transition.
- Prevent an older send failure or ACK from clearing a newer pending heartbeat.
- Avoid the zero sentinel collision when the u16 message ID wraps.

## Acceptance evidence

- A deterministic test forces an immediate ACK concurrently with send completion and the client remains registered across the next sweep.
- Send failure clears only the matching pending generation.
- Stale/wrong ACK IDs do not clear a newer request.
- Message-ID wrap, including zero, remains trackable.
- A live hub heartbeat timeout/ACK integration test complements the existing decision-helper tests.

## References

Standard 135-2020 Annex AB.6.3.

jscott3201:
Implementation is available in PR #538 at fd823cacbb4d8ec22d576bb0694ad35675c88934, based on dev1f0f62582db5521c6cbb6f88ee7ba1ae3a9f9634. Deterministic/live TLS-WebSocket regressions and hosted Linux CI33937720386 pass; both peer review lanes reported no findings.

The PR remains unmerged and this issue remains OPEN: the final review did not start because its target packet was incomplete. A corrected review plan and completed independent final assessment/reconciliation are still required. No code defect was identified by that procedural stop, and the existing merge authorization does not bypass the gate.

[...]

PR #538, opened 2026-09-05
fix(sc): make hub heartbeat attempts generation-owned

jscott3201:
## Summary
Refs #514.
- Publish map-owned pending ID, timestamp, and local generation before any heartbeat send becomes observable.
- Track every wire ID, including zero, without a sentinel collision; keep generations after ACK so stale local failures cannot retire newer attempts.
- Revalidate registration identity and generation before send/timeout cleanup; use consistent heartbeat retirement and a durable one-reader close signal.

## Contracts and exclusions
Preserve the existing hub-originated liveness probe as a qualified local extension, not a substitute for the initiating node's mandatory Annex AB.6.3 keepalive duty. Preserve the wall-clock domain, 30-second sweep, 60-second idle threshold, five-second ACK/send budgets, and strict deadline comparisons.
Local generations protect local asynchronous work, not delayed wire ACKs after reuse of the same 16-bit ID. No public configuration, dependency, CI, timing-policy, or support-claim expansion. General handler cancellation/hub supervision (#525/#476), already-in-flight old-sink bytes, and #526 follow-ups remain excluded.

## Source grounding
Local Standard 135-2020: AB.6.3 printed p.1405/PDF p.1407; AB.2.14–15 printed pp.1390–1391/PDF pp.1392–1393; AB.5.3.1 provides hub-connection context. Request/ACK IDs correlate on wire; generation is local state only. No licensed text reproduced or full Annex conformance claim.
Tokio 1.53.1 Notify source verifies notify_one retains one permit, while the closed predicate remains authoritative across select cancellation.

## Validation
- Deterministic red-before-green real TLS/WebSocket ACK-during-send-completion regression failed because the next sweep removed the ACKed client.
- Additional reds proved zero-ID non-expiry, stale failure removing newer state, send failure leaving a reader open, and close-before-wait requiring another notification.
- Thirteen deterministic/live regressions cover ACK/timeout ordering, replacement, generation exhaustion, wrap/zero and wrong IDs, sink/send budgets, immediate ACK and stale completion, closure and cross-client lock isolation.
- Transport sc-tls: 425 unit + 2 integration tests passed; no-default-features: 369 unit + 2 integration tests passed.
- Workspace fmt, Clippy, tests including the existing 11-test live SC suite, Python test compilation, conformance generation, file-size/no-secret, and diff checks passed.
- Exact-head review and hosted Linux full-feature CI required before merge.


jscott3201:
## Review gate — cycle 1 stopped; REPLAN_REQUIRED
Frozen base: 1f0f62582db5521c6cbb6f88ee7ba1ae3a9f9634
Frozen head: fd823cacbb4d8ec22d576bb0694ad35675c88934

Exact-head CI33937720386 passed Ubuntu full-feature tests, Clippy, Rustfmt, file-size, and no-secret checks. Independent holistic and dataflow reviews both passed with no findings.

The independent final assessor returned INCOMPLETE_TARGET before inspecting source: its dispatch lacked an explicit assessment artifact ID, prior-finding N/A field, and named required-check evidence set. This is a review-setup failure, not a code finding, and it is not an ASSESSED result. Peer PASS results cannot substitute for the missing independent final assessment/reconciliation.

The PR is intentionally NOT merged. Preserve this clean head and replan the final gate with a complete target before any merge. No repair or transport recovery was used; no source change is requested by this stop. #514 remains open. Existing scope exclusions and the qualified local-probe contract remain unchanged.

